PR #110, opened 2015-03-21
Python 3 compatibility

nils-werner:
This branch makes mir_eval compatible with Python 3. Nothing fancy, mainl `print` and `xargs` were acting up.

The only bigger thing was in `evaluators/eval_utilities.py`. In Python 3 `dict.items()` was always an iterable, while in Python 2 you had to explicitly call `dict.iteritems()`. There is a function `iteritems` in `future.utils` that implements it cross-version.


nils-werner:
Again, wrong branch.

PR #111, opened 2015-03-21
Fixed failing tests

nils-werner:
There were two issues with the tests:

Running `nosetests` from the root directory of the project failed as the scripts could not find the testdata. Prepending `tests/` to their path fixed the issue.

Also, there was an error only showing up intermittently: 

    FAIL: test_chord.test_join(<function join at 0x7f9dcc65ec80>, ('F#', '', {'4', '*b3'}, 'b7'), 'F#:(*b3,4)/b7')
      File "/usr/lib/python3.4/site-packages/nose/case.py", line 198, in runTest
        self.test(*self.arg)
      File "/home/nils/Arbeit/mireval3/mir_eval/tests/test_chord.py", line 22, in __check_valid
        assert function(*parameters) == result
    AssertionError

This had to do with the testdata being wrapped in `dict`s where they should've been in `lists` as `dict`s are not ordered. In some of the testruns the order of the arguments were reversed and the string comparison failed.

### DIFF
--- a/tests/test_chord.py
+++ b/tests/test_chord.py
@@ -89,11 +89,11 @@ def test_join():
     # Arguments are root, quality, extensions, bass
     splits = [('F#', '', None, ''),
               ('F#', 'hdim7', None, ''),
-              ('F#', '', {'*b3', '4'}, ''),
+              ('F#', '', ['*b3', '4'], ''),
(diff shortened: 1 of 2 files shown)


nils-werner:
Dang, wrong branch.

Issue #49, opened 2014-04-21
convention for empty intervals


bmcfee:
I ran across this curious example in the SALAMI dataset this morning, and it got me thinking about how we should handle semantically questionable (but syntactically valid) inputs.  The example in question comes from `988/parsed/textfile1_functions.txt`, and the offending lines (events) are:

```
321.788979591   Instrumental
321.788979591   Outro
321.788979591   Solo
```

mir_eval happy loads these as events, and following the zipping convention, converts to intervals without a problem:

```
(array([ 321.78897959,  321.78897959]), 'Instrumental'),
(array([ 321.78897959,  321.78897959]), 'Outro'),
(array([ 321.78897959,  397.66927438]), 'Solo'),
```

The problem here is that the event file listed three labels for a single boundary time.  The interval converter slams these together, and makes two length-0 intervals.  On our part, this is incorrect because the assumption of sequential unique timing is not satisfied.

This causes problems in segmentation metrics in a few ways: label equivalence is no longer well defined, and there are artificial boundaries in place.

It seems like we should settle on a convention for dealing with this type of problem, as it may be reasonable to expect similar things in other tasks.  Anyone care to chime in?

Issue #44, opened 2014-04-18
Warn when reference or estimation are empty and return 0


craffel:
The warning code should go in each submodule's validator.  Metrics should return 0 (or an otherwise appropriate "you got it wrong" value).
- [x] beat  @craffel
- [x] onset @craffel
- [x] separate @dawenl 
- [x] segment @bmcfee
- [x] chord @ejhumphrey 
- [x] melody @justinsalamon 
- [x] pattern @urinieto

PR #116, opened 2015-04-09
Fix boundary detection F-score in extreme cases

f0k:
Precision and recall were computed incorrectly for the two extreme cases of not predicting any boundary or having no ground truth boundaries to be matched.
(Note that this may not be the only occurrence, it's just the only one I looked at.)

### DIFF
--- a/mir_eval/segment.py
+++ b/mir_eval/segment.py
@@ -231,16 +231,23 @@ def detection(reference_intervals, estimated_intervals,
         reference_boundaries = reference_boundaries[1:-1]
         estimated_boundaries = estimated_boundaries[1:-1]
 
-    # If we have no boundaries, we get no score.
+    # If we have no boundaries, we do not need to match them.
     if len(reference_boundaries) == 0 or len(estimated_boundaries) == 0:
-        return 0.0, 0.0, 0.0
-
-    matching = util.match_events(reference_boundaries,
-                                 estimated_boundaries,
-                                 window)
-
-    precision = float(len(matching)) / len(estimated_boundaries)
-    recall = float(len(matching)) / len(reference_boundaries)
+        matching = []
+    else:
+        matching = util.match_events(reference_boundaries,
+                                     estimated_boundaries,
+                                     window)
+    # If we did not predict any boundary, all of them are correct.
+    if len(estimated_boundaries) == 0:
+        precision = 1
+    else:
+        precision = float(len(matching)) / len(estimated_boundaries)
+    # If there was no boundary to detect, we detected all of them.
+    if len(reference_boundaries) == 0:
+        recall = 1
+    else:
+        recall = float(len(matching)) / len(reference_boundaries)
 
     f_measure = util.f_measure(precision, recall, beta=beta)
 


bmcfee:
> Precision and recall were computed incorrectly

For the record: by "incorrectly" you mean we take `0/0=0` instead of `0/0=1`.

f0k:
> `0/0=0` here is consistent with sklearn, so I'd recommend keeping it as is unless there are strong arguments in favor of changing it.

An easy argument involves computing a precision/recall-curve. Say you have a threshold that you can vary to output more or fewer predicted boundaries. If you move the threshold towards zero, you will output more and more boundaries, driving your recall towards 1.0 and your precision towards 0.0. If you move the threshold towards the maximum, you will output fewer and fewer boundaries, driving your precision towards 1.0 and your recall towards 0.0. Shortly before reaching the maximum, you will predict a single (assumedly correct) boundary, i.e., you have a precision of 1.0 and recall of 1.0/num_gt_boundaries. When you go a little higher, your (and sklearn's) method will suddenly drop the precision back to 0.0, while the limit says it should be 1.0.

bmcfee:
The argument is actually much more subtle than you say.

If you're varying a single detection threshold (ie, for precision) while holding the true positives constant, then yes, the limit goes to 1.  In this case, mir_eval and sklearn do the right thing by default.

However, the case we're talking about here involves taking both the predicted and true positives to 0, which is not covered by your argument.  More precisely, we'd have to compute `lim_{tp, pp -> 0} pp/tp`, which is not a well-behaved sequence (ie, not bounded above), and monotone convergence doesn't apply. (I think; I'd have to consult a measure theory textbook...)

If you want to break it up into two limits, you get either
```
lim_{pp->0} lim_{tp->0} pp/tp
```
or
```
lim_{tp->0} lim_{pp->0} pp/tp
```
These limits are not interchangeable, and give different answers.  (Note that it's either 0 or +inf; 1 is not one of them.)

bmcfee:
> I'd be interested in sklearn's reasoning, though.

I just asked (It's handy having @amueller nextdoor :)), and to paraphrase: yeah, it matters which limit you take.

The only reason to return a value at all is to avoid breaking too much code.

My $2e-2 extra: if this corner case is hitting, those empty reference annotations should be removed from the evaluation anyway.

craffel:
Relevant issues: 
https://github.com/craffel/mir_eval/issues/44
https://github.com/craffel/mir_eval/issues/49

f0k:
> My $2e-2 extra: if this corner case is hitting, those empty reference annotations should be removed from the evaluation anyway.

My argument was only about the first case of the code (setting precision=1 when the set of predicted boundaries is zero). I agree that the second case (setting recall=1 when the set of reference boundaries is zero) should probably raise a warning instead.

> More precisely, we'd have to compute `lim_{tp, pp -> 0} pp/tp`, which is not a well-behaved sequence

For the precision, it's `lim_{tp, fp -> 0} tp/(tp+fp)`. And now I see the flaw in my reasoning. I assumed that `fp` would become zero first (the case of increasing a peak-picking threshold on a good novelty curve), so you'd end up with `lim_{tp -> 0} tp/tp`, which is 1. However, for a bad peak-picking curve, `tp` will become zero first, and we end up with `lim_{fp -> 0} 0/fp`, which is 0. Both is possible, so it doesn't matter which one you choose. Depending on the quality of your boundary detection algorithm, one or the other will give a more reasonable precision/recall curve.

> I just asked

Cool! :)

I'm now convinced that your choice is just as fine, this has been a handy discussion for me, thanks a lot!

bmcfee:
Cool.  I'm glad we hashed it out!

PR #118, opened 2015-04-24
allow unicode filenames in input_output

### LINKED ISSUE #117: io.load_delimited fails for unicode strings
It's easy to work around this problem, but I'll mention it regardless. When a file path that is passed to io.load_delimited is a unicode string, a value error is raised.

```
$ nonexistent_file = u"fake_filepath.csv"
$ mir_eval.io.load_delimited(nonexistent_file, ',')
ValueError: filename must be a string or file handle
```

rabitt:
fixes #117


craffel:
Thanks for this - so this does what you want it to?

rabitt:
yessir it does

craffel:
Merged, thanks!

PR #125, opened 2015-09-16
[WIP] T-measures for hierarchical segmentation

bmcfee:
cc @urinieto 

This PR implements the tree measures (t-measures) from [McFee, Nieto, and Bello 2015](http://bmcfee.github.io/papers/ismir2015_hierarchical_eval.pdf) for evaluating hierarchical boundary detection.

It is a direct port of our own preliminary implementation, but the interface has been stripped down to a more mir_eval-ish style and we have removed dependencies on JAMS and MSAF.

Externally, only a single function has been added `segment.tmeasure`, which accepts the following parameters:

- `reference_intervals_hier` : just like the other interval-based segmentation metrics, except now the data is a list of lists of intervals.  `reference_intervals_hier[i]` contains the segmentation at the `i`th level; levels are ordered by increasing depth or specificity.
- `estimated_intervals_hier` : as above, for the estimation.
- `window` : the length of the analysis window (in seconds).  See section 3.4, eq. 8.  Default is 15s, as per our recommendations (section 6).
- `transitive` : a flag to indicate whether transitive comparisons should be included. See section 3.5.  Default is `False`, as per our recommendations.
- `frame_size` : the length of sampled frames in the evaluation. This serves an identical purpose to other segmentation metrics, eg, `nce` or `pairwise`.
- `beta` : balance parameter for F-measure, as in all precision-recall metrics.

There are a handful of private support functions as well, including: `_round`, `_lca` (builds a sparse least-common ancestor matrix), and `_gauc` (implements eqs. 8 and 9).

To-do:

- [x] documentation
- [x] tests
- [x] validation and exception handling


craffel:
Looking cool.   I am tempted to suggest that this be added to a new submodule because there is a vague idea that every evaluator function in each submodule should take reference and estimated annotations in the same format, and this breaks that.  But, I could be convinced otherwise.  Also, looks like Travis is broken, I will create a separate issue for that.

bmcfee:
> there is a vague idea that every evaluator function in each submodule should take reference and estimated annotations in the same format, and this breaks that. But, I could be convinced otherwise.

`segment` already breaks that since we re-merged the boundary detection and structure annotation metrics.  (The latter require labels, the former do not.)

craffel:
Right, but you can call the boundary metrics with a subset of the structure metric arguments, so we can still call everything in the evaluate function. There's no overlap with hierarchical eval (unless you choose one layer somehow and use that?) so we would have to have two different evaluate functions.

bmcfee:
> There's no overlap with hierarchical eval (unless you choose one layer somehow and use that?) so we would have to have two different evaluate functions.

Sure.  I'm fine with either way.  There is some overlap in the validation routines, but I guess those could be imported.

craffel:
> There is some overlap in the validation routines, but I guess those could be imported.

We should put them in `util` then.

urinieto:
Even though I also would generally prefer being consistent across all functions, I think this is a special case, since music segmentation is an evolving task (that has never really been formally defined), so we should be a bit more flexible allowing cases like the hierarchical evaluation taking custom parameters.

We might see a new task in MIREX in the following years that evaluates hierarchical structure, but until then I wouldn't create a separate module.

craffel:
> Even though I also would generally prefer being consistent across all functions, I think this is a special case, since music segmentation is an evolving task (that has never really been formally defined),

So what should we do, have two separate `evaluate` functions in `segment`?  Or just make the hierarchical eval function an orphan function, not included in `evaluate`?  If this metric is evolving/not formally defined, we probably should wait to include it.  

> we should be a bit more flexible allowing cases like the hierarchical evaluation taking custom parameters.

The issue isn't that it takes custom parameters (most metrics do), it's that its call spec is incompatible with the other metrics in `segment` (as far as I've deduced, correct me if I'm wrong)

bmcfee:
> So what should we do, have two separate evaluate functions in segment? Or just make the hierarchical eval function an orphan function, not included in evaluate? 

That's how it currently is in my fork.  I don't mind it being that way, but it's your call.


> If this metric is evolving/not formally defined, we probably should wait to include it.

The metric is what it is (modulo default parameter settings).  The task of structure analysis itself is what's evolving (if I understand @urinieto correctly).


> The issue isn't that it takes custom parameters (most metrics do), it's that its call spec is incompatible with the other metrics in segment (as far as I've deduced, correct me if I'm wrong)

Sure.  I don't mind putting it in a different submodule, if that's what makes the most sense.  It seems a little silly to have a module that implements only one metric, but there's also enough support code involved with this one that it's not a completely terrible idea.

craffel:
> That's how it currently is in my fork. I don't mind it being that way, but it's your call.

I'd like it that every function in mir_eval can be called by an evaluator.  That's one of the ease-of-use principles, in my mind.

> It seems a little silly to have a module that implements only one metric, 

You mean like `separation` and `onset`? :)

bmcfee:
Hah, fair enough.  I'll move it.

Otherwise, does the interface for tmeasure meet your expectations of how things should look in mir_eval land?

craffel:
It does, but I think it may make most sense to use numpydoc format to conform with the imminent merge of develop.  I also wouldn't complain about a docstring for all of the underscore functions.  It's good that the return convention is (P, R, F) which matches `segment`, but you've just reminded me that `onset` returns things in a different order.  Separate issue there.

bmcfee:
> It does, but I think it may make most sense to use numpydoc format to conform with the imminent merge of develop.

right... i forgot all about that!

If you merge dev-> master, I'll rebase this one and then get everything back in line.

craffel:
Merged develop into master, please rebase!

bmcfee:
Yessir!

I should be able to wrap up the docs/validation/testing stuff by tonight.

bmcfee:
One question about input validation (mainly directed to @urinieto ): should tmeasure require that the inputs are strict hierarchies (olda-style) or should we also allow inconsistent multi-level segmentations (lsd-style)?

Currently, there's a check for consistency by verifying that boundaries at layers 1..k-1 are included in layer k.  This is necessary for the derivation of tmeasure, but I'm not sure it's necessary for the application.

The semantics of tmeasure are, basically, "count all frame triples (q, i, j) such that (q,i) are closer than (q, j)", where "closeness" is measured by comparing the deepest segments containing both elements of each pair.  As far as I can tell, this would work fine with inconsistent segmentations, in that it's still possible to build the pairwise depth matrix.  I have no idea if the resulting scores would be all that meaningful, but it seems plausible enough, given that we have implicit temporal continuity from the nature of the input anyway.

So: maybe we should relax the exception here to a warning? (Warnings can be caught and treated as exceptions later on anyway, if that's what the user is into.)

Please advise: I don't have strong opinions either way on this.

bmcfee:
Modulo response to the validation question above, I think this is feature-complete and ready for review.

craffel:
Two high-level things before I do an actual review (which will probably happen in about 1.5 weeks):

1. Is there a standard format for saving this kind of data?  Ideally, there should be a function for loading it in `io`; the idea being that `mir_eval` implements the entire toolchain from saved annotation files to metrics.

1. Related: There should be an evaluator script, for completeness, but that's only really possible with a standardized format and a way to load it in.

craffel:
Oops, one more:

3. Cite your ISMIR paper somewhere.

urinieto:
> should tmeasure require that the inputs are strict hierarchies (olda-style) or should we also allow inconsistent multi-level segmentations (lsd-style)?

Even though lsd-style segmentations don't make much sense (at least from a GTTM perspective), I like the idea of accepting them while throwing a warning message. Especially given how easy that would be in terms of implementation.

> Is there a standard format for saving this kind of data? 

We are [working](https://github.com/marl/jams/issues/61) on including this type of data in the jams format. Ultimately, we would like to connect jams with mir_eval for the hierarchical segmentation task, as we [did](http://pythonhosted.org/jams/jams.html#evaluation) for the rest of the tasks.

craffel:
Is there, or will there be, a non-jams format?  Or are you only ever planning on releasing this kind of data as jams?

urinieto:
If there is a standard, non-jams format to store this kind of data, I'm not aware of it.

craffel:
Isn't it basically you guys who are creating this data/task?  So isn't it up to you whether there's a non-jams format?

bmcfee:
> Is there, or will there be, a non-jams format? Or are you only ever planning on releasing this kind of data as jams?

I don't have a non-jams format in mind, and I'm not exactly sure how it would look.  It doesn't fit nicely into CSV since the annotations are basically a ragged tensor.  I suppose we could invent a flattened lab-style format with rows of the format ``start, end, level, label``, but since it would be entirely contrived in response to this comment thread, I don't see much point in doing so, and wouldn't encourage people to use it anyway.

This probably complicates the implementation of an evaluator script and io loader, right?

craffel:
>  It doesn't fit nicely into CSV since the annotations are basically a ragged tensor.

Well, csv-like formats are possible with ragged data, of course.  Reading them in would just involve `split` and `strip` string operations.  Not that I am advocating for that - as I said, my understanding is that this is sort of your task and your data so I'm not going to encourage you to use an inferior format because it's lighter weight than jams.

> This probably complicates the implementation of an evaluator script and io loader, right?

It only complicates things to the extent that it either adds a dependency to jams or it re-implements jams functionality.  I would be ok with the evaluator importing jams, not having a function in `io` to load files for this task, and adding a note to the task submodule about loading data in jams, iff the only format this data will ever be distributed in is jams.

bmcfee:
> It only complicates things to the extent that it either adds a dependency to jams or it re-implements jams functionality. I would be ok with the evaluator importing jams, not having a function in io to load files for this task, and adding a note to the task submodule about loading data in jams, iff the only format this data will ever be distributed in is jams.

Since jams depends on mir_eval, that's maybe not such a hot idea. :)

Currently, the only existing annotations in this format are from SALAMI, where each layer is stored as a different csv file.  (I'm not counting those crazy native salami formats with the event boundaries and nested interval labels...ugh.)

In our ismir experiments, everything was packed into JAMS first and then run through the evaluator.  Maybe it's worth paging @grrrr and @f0k for opinions, since they've also worked with multi-layer data, and may have opinions about the easiest way to deal with the annotations.

craffel:
> Currently, the only existing annotations in this format are from SALAMI, where each layer is stored as a different csv file. (I'm not counting those crazy native salami formats with the event boundaries and nested interval labels...ugh.)

Hmm, perhaps the best way to handle it would be to have a convenience function which reads in boundaries from each file in a list of CSV files and returns a ragged list.  This is an ugly solution, but it's sort of the baseline lightest-weight option (analogous to lab files for chord/segment): It's not necessarily recommended, but they're easy to support and probably worth supporting.  Maybe this brings into the broader question of how to balance compatibility with existing brittle formats in `io` while encouraging people to use jams instead?

> Maybe it's worth paging @grrrr and @f0k for opinions, since they've also worked with multi-layer data, and may have opinions about the easiest way to deal with the annotations.

Yes, to bring you up to speed: Brian is adding hierarchical segment evaluation metrics from his ISMIR 2015 paper.  For all existing tasks, we have functionality in `io` to read in annotations for that task.  We are discussing what kind of file parser should be added to `io` to handle data for this task.

f0k:
> Yes, to bring you up to speed: Brian is adding hierarchical segment evaluation metrics from his ISMIR 2015 paper. For all existing tasks, we have functionality in `io` to read in annotations for that task. We are discussing what kind of file parser should be added to `io` to handle data for this task.

We parsed all the SALAMI annotations into a pickled dictionary that we subsequently use for everything we need annotations for. So for us, it doesn't matter what `io` provides, we just need a way to easily use or copy out the evaluation function itself and feed it with the data in the form of some standard Python data structure such as a list of tuples.

> Currently, the only existing annotations in this format are from SALAMI, where each layer is stored as a different csv file.

You could provide some example for parsing these files, of course, since most people will be interested in those (as you said, they're the only published hierarchical segmentations). And maybe some other format that's easy to write out. Both would mainly be targeted at people who don't work in Python and still want to use `mir_eval` (e.g., Matlab users).

bmcfee:
> So for us, it doesn't matter what io provides, we just need a way to easily use or copy out the evaluation function itself and feed it with the data in the form of some standard Python data structure such as a list of tuples.

Cool, thanks for the feedback!

The current api uses a list of interval arrays.  I think this should be simple enough to construct from what you describe: each entry in the list is a flat segmentation, and the segmentations are ordered by level.  This is intended to be as close to the flat segmentation metrics' interface as possible.



> You could provide some example for parsing these files,

The trouble here is that it would make the interface to the eval script pretty nasty.  Instead of the usual
```bash
./task_eval.py ref_file est_file arguments
```
we'd need something like
```bash
./task_eval.py -r ref_file_0 -r ref_file_1 -r ref_file_2 -e est_file_0 -e est_file_1 -e est_file_2  arguments
```
and I don't think anyone wants that.

An alternative would be an intermediate parser/converter script that smashes multiple csv files into a sane format (eg, the 4-tuple of ``start, end, label, level`` described above), and then we can go back to the simpler eval script format.  I prefer this approach, but it's still hackish IMO.

f0k:
> An alternative would be an intermediate parser/converter script that smashes multiple csv files into a sane format (eg, the 4-tuple of start, end, label, level described above), and then we can go back to the simpler eval script format. I prefer this approach, but it's still hackish IMO.

I think that's a perfectly valid approach. Just provide a helper script that's specifically targeted at preprocessing the SALAMI annotations, and tell users to write their predictions in the sane format right from the start. Then `./task_eval.py ref_file est_file arguments` would be enough.

To clarify what I've written before, I think that Python users may not need a runnable `task_eval.py` script (that dictates the I/O format to use), because they can use the evaluation functions directly. So whatever file format you choose for `task_eval.py` should be easy to handle in other languages (such as Matlab or Java).

bmcfee:
Ok, this all sounds good to me.

@craffel if I make a salami multi-seg -> csv converter, where ought it live?  Should we make a separate directory of scripts for things other than evaluators?

craffel:
> @craffel if I make a salami multi-seg -> csv converter, where ought it live? Should we make a separate directory of scripts for things other than evaluators?

Is the reason you want this to be a script rather than a function in `io` which takes, say, *args which are paths to csv files to merge, so that non-Python users have an easier time with it?  If that's the main reason, I'd advocate for just using a function in `io` with good documentation.  If a non-Python user wants to do this, they can do it in their native language, or even just use the function in the one or two lines of Python code which is included in the docstring.

If we do decide to have a separate place for non-evaluator helpful scripts, I'd suggest putting them in an "examples" or "scripts" directory but I'm having trouble thinking of what else to put in there.

bmcfee:
> Is the reason you want this to be a script rather than a function in io which takes, say, *args which are paths to csv files to merge, so that non-Python users have an easier time with it? If that's the main reason, I'd advocate for just using a function in io with good documentation.

We can have an io function for this --- though I don't think it's necessary --- but unless we have an intermediate file format, we'd still have to call it from the eval script, and then we're back into the situation I described above.

An intermediate file format would help with this, or at least, it would standardize the eval scripts and push the complexity off to pre-processing.

(Wouldn't life be eaiser if everyone used jams? :))

craffel:
> We can have an io function for this --- though I don't think it's necessary

Wait why not?  I thought you wanted to create a script for this?

> --- but unless we have an intermediate file format, we'd still have to call it from the eval script, and then we're back into the situation I described above.

Isn't the intermediate file format a flattened CSV, or something?  Why does having a function in `io` preclude having an intermediate file format?

bmcfee:
> Wait why not? I thought you wanted to create a script for this?

1) It's feature-creep, and basically a one-liner if you've already done the work of arranging your annotation files and loading them through existing io functions:
```python
>>> files = ['functions.lab', 'large.lab', 'small.lab']
>>> hier_intervals = [mir_eval.io.load_intervals(filename) for filename in files]
```

2) yes, i was thinking a script to convert multiple salami annotations to a single, flattened csv format.  An io function for *that* format might make sense, but not the multi-file loader.  (I also don't think we should necessarily encourage people to use this concocted flat csv format, but that's a separate debate.)

craffel:
> 1) It's feature-creep, and basically a one-liner if you've already done the work of arranging your annotation files and loading them through existing io functions:
 ```Python
 >>> files = ['functions.lab', 'large.lab', 'small.lab']
 >>> hier_intervals = [mir_eval.io.load_intervals(filename) for filename in files]
```

Ok, maybe just put that exact code snippet in the docstrings of the hierarchical eval metric functions/evaluate function?

> 2) yes, i was thinking a script to convert multiple salami annotations to a single, flattened csv format. An io function for that format might make sense, but not the multi-file loader. (I also don't think we should necessarily encourage people to use this concocted flat csv format, but that's a separate debate.)

It seems like the only reason we'd need this is to have an evaluator script that had a sane command-line signature.  I don't think it's worth it to make up a new annotation format (flattened CSV) that you don't want people to use just to make an evaluator script.  If there isn't a standard format for this task (which is not JAMS, which you mentioned not wanting to use because the JAMS library depends on `mir_eval`), we can skip the evaluator script I guess.

**Comment by bmcfee** [comment]
> Ok, maybe just put that exact code snippet in the docstrings of the hierarchical eval metric functions/evaluate function?

Sure, that seems like a reasonable compromise.  I'll come up with a complete docstring example that handles salami annotations exactly.



> we can skip the evaluator script I guess.

Ok, but only if you really don't mind. :)  I appreciate the motivation for having an evaluator function for each task.  It just seems a little unwieldy at this point.  Maybe we can revisit this in the future if more people start working on multi-level segmentation?

**Comment by craffel** [comment]
> Ok, but only if you really don't mind. :) I appreciate the motivation for having an evaluator function for each task. It just seems a little unwieldy at this point. Maybe we can revisit this in the future if more people start working on multi-level segmentation?

It does bother me to have a task in `mir_eval` with no evaluator script, but what that simplifies to is that it bothers me to have a task in `mir_eval` with no standardized annotation format.  But we will limit ourselves somewhat if we only implement tasks with standardized annotation formats, I guess.

**Comment by bmcfee** [comment]
> but what that simplifies to is that it bothers me to have a task in mir_eval with no standardized annotation format.

:+1: :heavy_multiplication_x: :100: 

.. but that said, I don't have any other plausible suggestions.  (Forking the eval scripts out to a separate package might solve some problems, but I doubt that would be a popular idea here. :))

Anyway, commit af9cf97e8f1b9154614feb2da39db8412f2577c3 implements the improved docstring + salami example.  (Also, why is salami 2.0 still using labeled events instead of intervals? :disappointed: )

**Comment by urinieto** [comment]
My two cents after reading all of this:

I think having something like what @bmcfee initially proposed, i.e.,
```./task_eval.py -r ref_file_0 -r ref_file_1 -r ref_file_2 -e est_file_0 -e est_file_1 -e est_file_2```
is better than having a new CSV format that accepts hierarchical segmentation. 

Having this solution would i) allow for an easily implemented evaluation script, ii) avoid a new CSV format, and iii) let users keep working with CSV files that are (somewhat) standard in music segmentation. The only downside I see is the "ugly" interface to that script, even though I consider it a minor thing.

Regardless, and as it has been mentioned in this thread, the hierarchical version of this task does not have a standardized annotation format (yet), so all of this might be just a temporary solution.

**Comment by f0k** [comment]
> Having this solution would i) allow for an easily implemented evaluation script, ii) avoid a new CSV format, and iii) let users keep working with CSV files that are (somewhat) standard in music segmentation. The only downside I see is the "ugly" interface to that script, even though I consider it a minor thing.

To make it a little more handy, `-r` and `-e` could accept patterns as well as filenames, to be expanded with `glob.glob()` (then sorted by file name). That would turn the example call to `./task_eval.py -r ref_file_? -e est_file_?`.

**Comment by bmcfee** [comment]
> To make it a little more handy, -r and -e could accept patterns as well as filenames, to be expanded with glob.glob() (then sorted by file name). That would turn the example call to ./task_eval.py -r ref_file_? -e est_file_?.

The trouble with this is that the salami files wouldn't work by default, since `'l' < 'u'`.  I'd prefer to leave globbing to the shell in this case.

Otherwise, if folks don't think the interface is too clunky, I'm happy to add this script.  One more potential snag: would we want it to accept salami-style or lab-style inputs for the individual segmentations?  (Trying to support both in one go seems complicated.)

* SALAMI
  - Pro: it would work out of the box on an existing data set
  - Con: it's a bad format that we shouldn't encourage using.  It's also a little circuitous to [handle](https://github.com/craffel/mir_eval/blob/af9cf97e8f1b9154614feb2da39db8412f2577c3/mir_eval/hierarchy.py#L380-L383) in mir_eval right now.
* Lab-style
  - Pro: it's consistent with the other segment eval script
  - Con: No existing datasets in this format for hierarchy, so it wouldn't really work out of the box.

**Comment by f0k** [comment]
> The trouble with this is that the salami files wouldn't work by default, since 'l' < 'u'. I'd prefer to leave globbing to the shell in this case.

If you leave globbing to the shell, you'd get `./task_eval.py -r ref_file_0 ref_file_1 ref_file_2 -e est_file_0 est_file_1 est_file_2`.

>  One more potential snag: would we want it to accept salami-style or lab-style inputs for the individual segmentations?

If supporting both is too hard, I'd suggest to go for lab-style, and provide a helper script for converting the SALAMI annotations to lab-style format, renaming them on the way to indicate the level as a number.

But shouldn't it be easy to support both? What do the lab files look like?

**Comment by urinieto** [comment]
What I suggest is to add that script with the syntax that @f0k proposed. The parsed files in SALAMI can already be considered lab files (e.g. [this](https://github.com/DDMAL/salami-data-public/blob/master/annotations/10/parsed/textfile1_functions.txt)), so I don't think it's necessary to add an additional converter.

**Comment by craffel** [comment]
Yes, assuming that the SALAMI files can be considered Lab files, I am ok with the original verbose evaluator script syntax.

**Comment by bmcfee** [comment]
They aren't exactly lab files, since they lack ending time.  We can hack something together that tries a vanilla lab import, and then bounces down to event->interval conversion on failure.  In general, I hate doing that sort of thing, since parser cleverness leads to eventual madness.  But I suspect nobody will use this thing anyway, so it hardly matters. :)

**Comment by craffel** [comment]
Hm.  Yeah, I have no idea, and it probably doesn't matter as there aren't a lot of people doing this task.  Let's just assume nice lab file format because that's what we support,without hacks,and leave it to the user to convert salami to lab interval format.

**Comment by bmcfee** [comment]
That sounds like a reasonable compromise to me.

Meanwhile, maybe I'll send a PR to the salami repo that has all the data in a sane format.

**Comment by craffel** [comment]
Sounds like a plan.

**Comment by urinieto** [comment]
+1

**Comment by bmcfee** [comment]
Latest commits add the evaluator script, using the sequence-of-labs style input:

```
$ ./segment_hier_eval.py -h
usage: segment_hier_eval.py [-h] [-w WINDOW] [-o OUTPUT_FILE]
                            [-r REFERENCE_FILE [REFERENCE_FILE ...]]
                            [-e ESTIMATED_FILE [ESTIMATED_FILE ...]]

mir_eval hierarchical segmentation evaluation

optional arguments:
  -h, --help            show this help message and exit
  -w WINDOW, --window WINDOW
                        Window length for t-measures
  -o OUTPUT_FILE        Store results in json format
  -r REFERENCE_FILE [REFERENCE_FILE ...], --reference REFERENCE_FILE [REFERENCE_FILE ...]
                        path to the reference annotation(s) in .lab format,
                        ordered from top to bottom of the hierarchy
  -e ESTIMATED_FILE [ESTIMATED_FILE ...], --estimate ESTIMATED_FILE [ESTIMATED_FILE ...]
                        path to the estimated annotation(s) in .lab format,
                        ordered from top to bottom of the hierarchy
```

Example usage/output:
```
$ ./segment_hier_eval.py -r ~/data/beatles_iso/seglab/The\ Beatles/01_-_Please_Please_Me/01_-_I_Saw_Her_Standing_There.lab \
                         -e ~/data/beatles_iso/seglab_tut/The\ Beatles/01_-_Please_Please_Me/01_-_I_Saw_Her_Standing_There.lab 
01_-_I_Saw_Her_Standing_There.lab [...] vs. 01_-_I_Saw_Her_Standing_There.lab [...]
	T-Precision reduced : 0.642909431702
	   T-Recall reduced : 0.639117011778
	  T-Measure reduced : 0.641007612476
	   T-Precision full : 0.642909431702
	      T-Recall full : 0.639117011778
	     T-Measure full : 0.641007612476

```

craffel:
Looks good to me overall.  A few misc things:

1. Can you run pep8 on all the files you're adding?  It looks like your line width limit is set larger than PEP-8.
1. It'd be cool to have coverage of 100% of `hierarchy.py` before merging, but we don't have 100% coverage for the rest of the library so I don't mind.
1. I don't really know anything about this evaluation, so I can't really review the implementation and vouch for whether its correct, but I think you are the person to do it.  If the code is any different from your original code, it'd be cool to get @urinieto to give the go-ahead.
1. For what it's worth, we generally do not support command-line setting of evaluation parameters in the evaluator scripts, so it's a bit inconsistent to allow the window parameter to be set.  But, if you think that's important, that's fine with me.
1. Can you squash some commits?  I don't think we need full history for a new submodule.

bmcfee:
> Can you run pep8 on all the files you're adding? It looks like your line width limit is set larger than PEP-8.

Man, if there's one thing PEP8 should never have instituted... but yeah.

> It'd be cool to have coverage of 100% of hierarchy.py before merging, but we don't have 100% coverage for the rest of the library so I don't mind.

Sure.

> I don't really know anything about this evaluation, so I can't really review the implementation and vouch for whether its correct, but I think you are the person to do it. If the code is any different from your original code, it'd be cool to get @urinieto to give the go-ahead.

It's a little different, but it does line up with the synthetic examples in the paper.  Extra CR from @urinieto is always welcome!

> For what it's worth, we generally do not support command-line setting of evaluation parameters in the evaluator scripts, so it's a bit inconsistent to allow the window parameter to be set. But, if you think that's important, that's fine with me.

I just copied this over from the existing segment evaluator, which allows a flag for trimming.  I don't have strong feelings about this for the window parameter, and am happy to drop it.

> Can you squash some commits? I don't think we need full history for a new submodule.

Will do after the smoke clears.

urinieto:
Titans,

The code also looks good to me, so after PEP8 revision I think you should
merge this :)

I would leave w as a parameter, but if Colin feels strongly against it it's
ok for me to drop it (and leave the default w to 15 seconds).

craffel:
> I would leave w as a parameter, but if Colin feels strongly against it it's
ok for me to drop it (and leave the default w to 15 seconds).

Either is fine.  If it's a parameter which you see as being non-standardized enough that people will want to change it, I'd say it's worth including a command line argument for it.  If it's something you want to be standardized and fixed at 15 seconds forever, don't include a command line argument.

bmcfee:
> If it's a parameter which you see as being non-standardized enough that people will want to change it, I'd say it's worth including a command line argument for it.

Maybe not "standardized", but a range of values might be interesting to report.  We do a lot of this in the paper, so it might be nice to expose that interface to make it easy to replicate.

craffel:
Sounds good to me, feel free to leave it in.

bmcfee:
Coverage is up to 100 and noisy commits are squashed.  I added some regression tests based on salami track #636 (Figure 8 in the paper).

I think this one's ready.

urinieto:
Nice!

craffel:
Thanks!

Issue #129, opened 2015-09-17
travis build on 3.4, 3.5?


bmcfee:
The current config is set for 2.7 and 3.3; we should at least bump to 3.4 since [nobody really uses 3.3](http://www.randalolson.com/2015/01/30/python-usage-survey-2014/).  3.5 would be nice as well.

craffel:
Sounds good, go for it!

bmcfee:
Any objection to adding coveralls support, while i'm at it?

craffel:
No, please do!  It was on some to-do list somewhere.

bmcfee:
Roger.

Also, any thoughts about replacing this wheels download with a miniconda environment?  Conda's scipy packages get updated much more regularly, apparently.

craffel:
> Also, any thoughts about replacing this wheels download with a miniconda environment? Conda's scipy packages get updated much more regularly, apparently.

Yeah, I'd prefer miniconda.  I actually thought that's what we were using, but I didn't add the travis support, so what do I know.

bmcfee:
Alright.  Maybe i'll rename this issue to "generally fix up the CI configs"

craffel:
:+1:

bmcfee:
Got it.  Miniconda doesn't yet have 3.5 support http://continuum.io/blog/python35 , so let's hold off on that one for now.

craffel:
Closed via #130.

PR #130, opened 2015-09-17
General upgrades to CI system #129

bmcfee:
re thread in #129 : 

- Revised environment to use miniconda
- Upgraded 3.3 to 3.4 testing
- Added coveralls

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,34 @@
 language: python
+
+notifications:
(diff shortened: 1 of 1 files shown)


bmcfee:
I think this one's ready to go.

We have some work to do in improving coverage, but we knew that already...

craffel:
LGTM, merging.

PR #134, opened 2015-10-16
Remove call to inspect.getargspec, use 3.5 for testing

### LINKED ISSUE #133: filter_kwargs and python 3.5 compatibility
`util.filter_kwargs` uses the inspect module's `getargspec` function [here](https://github.com/craffel/mir_eval/blob/master/mir_eval/util.py#L715), which is deprecated in python 3.5.

In python 3, this has been replaced with the `signature` function (see [here](https://docs.python.org/3/library/inspect.html#inspect.signature)), which sadly doesn't work on py2.

You might be able to side-step this by using `six.get_function_defaults()`.

Alternatievly, our buddies at sklearn have a fix for this [here](https://github.com/scikit-learn/scikit-learn/commit/6966881328c256bbef37752cce23bbac8e23fb27).

craffel:
Resolves #133.  Replaces the call in `util.filter_kwargs` to `inspect.getargspec` with a two-line equivalent, and ups the testing Python3 version to 3.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ notifications:
 python:
     - "2.7"
     - "3.4"
+    - "3.5"
 
 before_install:
     - if [[ "$TRAVIS_PYTHON_VERSION" == "2.7" ]]; then
--- a/README.rst
+++ b/README.rst
@@ -10,6 +10,8 @@ If you're looking for the mir_eval web service, which you can use to run mir_eva
 Dependencies:
 
 * `Scipy/Numpy <http://www.scipy.org/>`_
+* future
+* six
 
 If you use mir_eval in a research project, please cite the following paper:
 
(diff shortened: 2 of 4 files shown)


craffel:
Travis likes it.  @bmcfee any input?

bmcfee:
Looks okay to me.  I would recommend keeping both the 3.4 and 3.5 test setups though.

craffel:
> I would recommend keeping both the 3.4 and 3.5 test setups though.

Ok, done.

bmcfee:
LGTM!

Side note: you may want to switch over to the container setup on travis so that test environments can be cached, but that can be a separate issue.

craffel:
> Side note: you may want to switch over to the container setup on travis so that test environments can be cached, but that can be a separate issue.

Hm, sounds nice, I don't know how to do that.  Feel free if you're up for it.

PR #142, opened 2015-10-25
Warn when a non-uniform timebase is passed to resample_melody_series

### LINKED ISSUE #103: "incomplete" files passed to melody.evaluate should pass a warning.
f0 annotations are often output by algorithms with the "unvoiced" frames omitted, resulting in an incomplete set of time stamps. I was evaluating one of these types of melody output against a ground truth (with all time stamps, f0=0 for unvoiced frames), and melody.evaluate runs without problems.

After a brief talk with @justinsalamon, I learned that melody.evaluate is actually interpolating the missing f0 values. I can see use cases where you would want this behavior, but I might not have caught this problem. So, maybe melody.evaluate should pass a warning if a file with unevenly spaced time stamps is inputted, or, better yet, have a parameter to "fill missing frames".

Example files can be found [here](https://gist.github.com/rabitt/c45b969adf927cbe14b2)

craffel:
Resolves #103.  Please give the go-ahead, @justinsalamon or @rabitt.

### DIFF
--- a/mir_eval/melody.py
+++ b/mir_eval/melody.py
@@ -213,6 +213,17 @@ def resample_melody_series(times, frequencies, voicing,
         Boolean voicing array resampled to new timebase
 
     """
+    # Warn when the delta between the original times is not constant,
+    # unless times[0] == 0. and frequencies[0] == frequences[1] (see logic at
+    # the begnning of to_cent_voicing)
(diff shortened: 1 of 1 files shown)


craffel:
Ignore the coveralls failure, see #136

justinsalamon:
Just had a look, looks good.

PR #144, opened 2015-10-25
Fixed bug on intervals_to_boundaries function

urinieto:
This function was having an undefined behavior since it was using `np.unique` for floats. This made the whole segment detection yield --sometimes-- wrong numbers (I was getting like ~15% lower scores in [msaf](https://github.com/urinieto/msaf)). It took me a while to figure this one out...

### DIFF
--- a/mir_eval/util.py
+++ b/mir_eval/util.py
@@ -203,7 +203,8 @@ def intervals_to_boundaries(intervals):
 
     """
 
-    return np.unique(np.ravel(intervals))
+    return np.concatenate((intervals.flatten()[::2], [intervals[-1, -1]]),
+                          axis=0)
 
 
 def boundaries_to_intervals(boundaries, labels=None):
@@ -280,7 +281,7 @@ def adjust_intervals(intervals,
     Returns
     -------
     new_intervals : np.ndarray
-        Intervals spanning [t_min, t_max]    
+        Intervals spanning [t_min, t_max]
     new_labels : list
         List of labels for new_labels
 
@@ -441,7 +442,7 @@ def intersect_files(flist1, flist2):
         Parameters
         ----------
         abs_path :
-            
+
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


craffel:
What's wrong with `np.unique` with floats?  Rounding error issues?

It looks like the new code fails some tests, you'll need to fix those.

You should issue the pull request against master.  We aren't using develop anymore.

@bmcfee is best suited to comment on this, I think.

urinieto:
Yeah, rounding errors. I believe this worked with 2.7 but not 3.4 (not sure
about this, but it was working for some versions for sure). Will fix the
tests later, and put it in master.

craffel:
I don't think this necessarily has the desired behavior:

```Python
>>> intervals = np.array([[1., 2.], [3., 4.]])
>>> np.unique(np.ravel(intervals))
array([ 1.,  2.,  3.,  4.])
>>> np.concatenate((intervals.flatten()[::2], [intervals[-1, -1]]), axis=0)
array([ 1.,  3.,  4.])
```

bmcfee:
I suspect what you probably want is a unique after rounding to some number of decimals by `np.around`.

urinieto:
The intervals in your example should not valid. They should be of the form:

```
>>> intervals = np.array([[1., 2.], [2., 3.]])
>>> np.concatenate((intervals.flatten()[::2], [intervals[-1, -1]]), axis=0)
array([ 1.,  2.,  3.])
```

Regardless, I like @bmcfee 's idea to round it up. So I will shortly push changes (to master, ugh..).

urinieto:
mmmhh not sure how to change the origin from develop to my master version in the PR...

craffel:
> The intervals in your example should not valid. They should be of the form:

Wait, what?  Intervals which do not cover the entire song/are disjoint are valid (for example, chord labels), right?

> mmmhh not sure how to change the origin from develop to my master version in the PR...

Close this one and open a new one.

urinieto:
But this function is exclusively for boundaries, right? (hence its name).
At this moment, all segment tasks must have segments that cover the entire
track.

bmcfee:
> But this function is exclusively for boundaries, right? (hence its name).

I think the original logic behind the function was correct: the starts and ends of intervals are valid boundaries, whether or not the intervals span the full duration. (This is why event boundaries are ambiguous.)

PR #147, opened 2015-10-27
Rounding up the intervals before converting them

urinieto:
This comes from #144, where undefined behavior occurs when using `np.unique` with large float numbers. This should fix it.


bmcfee:
LGTM, pending tests

craffel:
Merged in 541718df03f10df0107296c256f7d0be11b101e4, thanks titan.

PR #155, opened 2015-11-10
Bugfix: Typo in variable name

stefan-balke:
Can be reconstructed by setting ref to est.

ref_time = melodia_time
ref_freq = melodia_freq

est_time = ref_time
est_freq = est_time

melody_scores = mir_eval.melody.evaluate(ref_time, ref_freq,
                                                                      est_time, est_freq)

Error was then:
NameError: global name 'frequences' is not defined


craffel:
Oof, this is a pretty good argument for 100% coverage.  Thanks for the catch.

PR #163, opened 2016-02-03
implemented interval timing on time-frequency sonification

bmcfee:
This PR extends sonify.time_frequency to support overlapping frequency lines.

This is done by extending the API to support either implicit timing (start times only, the old behavior) or interval timing (start, end times), and modifying the synthesis accordingly.

This change should be backward compatible.

### DIFF
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -4,6 +4,7 @@ All functions return a raw signal at the specified sampling rate.
 '''
 
 import numpy as np
+from numpy.lib.stride_tricks import as_strided
 from . import util
 from . import chord
 
@@ -98,15 +99,28 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
     times = times[:gram.shape[1]]
 
     def _fast_synthesize(frequency):
-        """A faster (approximate) way to synthesize a signal
-            synthesize a few periods then repeat that signal
+        """A faster way to synthesize a signal.
+            Generate one cycle, and simulate arbitrary repetitions
+            using array indexing tricks.
         """
+        frequency = float(frequency)
+
         # Generate ten periods at this frequency
-        ten_periods = int(10*fs*(1./frequency))
-        short_signal = function(2*np.pi*np.arange(ten_periods)*frequency/fs)
-        # Repeat the signal until it's of the desired length
+        n_samples = int(10.0 * fs / frequency)
+
+        short_signal = function(2.0 * np.pi * np.arange(n_samples) *
+                                frequency / fs)
+
+        # Calculate the number of loops we need to fill the duration
         n_repeats = int(np.ceil(length/float(short_signal.shape[0])))
-        return np.tile(short_signal, n_repeats)[:length]
+
+        # Simulate tiling the short buffer by using stride tricks
+        long_signal = as_strided(short_signal,
+                                 shape=(n_repeats, len(short_signal)),
+                                 strides=(0, short_signal.itemsize))
+
+        # Use a flatiter to simulate a long 1D buffer
+        return long_signal.flat
 
     # Pre-allocate output signal
     output = np.zeros(length)
@@ -115,6 +129,9 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
         wave = _fast_synthesize(frequency)
         # Scale each time interval by the piano roll magnitude
         for m, (start, end) in enumerate((times * fs).astype(int)):
+            # Clip the timings to make sure the indices are valid
+            start, end = max(start, 0), min(end, length)
+
             # Sum into the aggregate output waveform
             output[start:end] += wave[start:end] * gram[n, m]
 
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -81,12 +81,21 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
     Returns
     -------
     output : np.ndarray
-        synthetized version of the piano roll
+        synthesized version of the piano roll
 
     """
     # Default value for length
+    if times.ndim == 1:
+        # Convert to intervals
+        times = util.boundaries_to_intervals(times)
+
     if length is None:
-        length = int(times[-1]*fs)
+        length = int(times[-1, 1] * fs)
+
+    times, _ = util.adjust_intervals(times, t_max=length)
+
+    # Truncate times so that the shape matches gram
+    times = times[:gram.shape[1]]
 
     def _fast_synthesize(frequency):
         """A faster (approximate) way to synthesize a signal
@@ -104,13 +113,11 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
     for n, frequency in enumerate(frequencies):
         # Get a waveform of length samples at this frequency
         wave = _fast_synthesize(frequency)
-        # Zero out up to first time
-        wave[:int(times[0]*fs)] = 0
         # Scale each time interval by the piano roll magnitude
-        for m, (start, end) in enumerate(zip(times[:-1], times[1:])):
-            wave[int(start*fs):int(end*fs)] *= gram[n, m]
-        # Sume into the aggregate output waveform
-        output += wave
+        for m, (start, end) in enumerate((times * fs).astype(int)):
+            # Sum into the aggregate output waveform
+            output[start:end] += wave[start:end] * gram[n, m]
+
     # Normalize
     output /= np.abs(output).max()
     return output


bmcfee:
```mir_eval/sonify.py:119:80: E501 line too long (83 > 79 characters) ```

ಠ_ಠ

**Comment by bmcfee** [comment]
A couple of other things we might want to address in this PR:

- [ ] I think fast_synthesize could be made more efficient using stride tricks to simulate a much larger buffer
- [ ] For "low" sampling rates (ie 22050) there's some audible modulation due to the not-quite-repeating synthesis buffer
- [ ] We could avoid transient pops by aligning start/end times to the closest zero crossing in the buffer.  I really hate these pops :(

**Comment by bmcfee** [comment]
In other exciting news, 2.7 and 3.4 builds are failing because of conda issues. The 3.5 build is humming along nicely though.

Any thoughts, @craffel ?  Maybe a build cache issue?

**Comment by craffel** [comment]
> This is done by extending the API to support either implicit timing (start times only, the old behavior) or interval timing (start, end times), and modifying the synthesis accordingly.
> This change should be backward compatible.

Looks reasonable to me.  Does it do what you want it to do now?

>  I think fast_synthesize could be made more efficient using stride tricks to simulate a much larger buffer

Go for it, if you want!

> For "low" sampling rates (ie 22050) there's some audible modulation due to the not-quite-repeating synthesis buffer
> We could avoid transient pops by aligning start/end times to the closest zero crossing in the buffer. I really hate these pops :(

Neither of these bother me - this is not supposed to be a high-quality resynthesis, nor is it really possible to create one out of a piano roll, arguably.  Also, using `scipy.signal.square` as the periodic function will suppress both these issues, and I like using square waves any ways because they sound a little nicer when you're resynthesizing time-frequency representations, but that's beside the point.  If you want to address either/both, feel free, but finding the nearest zero crossing sounds like it will get a little ugly.  A nice solution to the pops would be to smoothly interpolate between amplitude levels, which would be extremely straightforward with the original start-time-only call structure but less than straightforward with non-overlapping intervals.

> In other exciting news, 2.7 and 3.4 builds are failing because of conda issues. The 3.5 build is humming along nicely though.
> Any thoughts, @craffel ? Maybe a build cache issue?

No idea why. I will look into it.

**Comment by bmcfee** [comment]
> Looks reasonable to me. Does it do what you want it to do now?

Yup!

> Go for it, if you want!

Done.  I had to use a combination of stride_tricks and flatiter, but you can now synthesize arbitrarily long signals with a single-cycle buffer.  I didn't time it, but the overhead should be negligible.



> No idea why. I will look into it.

Thanks.  I'll rebase off master once you merge the travis fix.

**Comment by bmcfee** [comment]
I also changed the synth logic a little to use a linspace rather than re-normalized arange.  This should be both slightly more efficient, and (to my ears), eliminates the modulation I was noticing with lower sampling rates.

**Comment by bmcfee** [comment]
I'll punt on the pop removal.  Otherwise, I think this one's ready to go.

**Comment by craffel** [comment]
Cool, looks good to me.  Can you squash your PEP-8 commits and I'll merge?

> I'll punt on the pop removal.

This would be nice to fix, at some point. The simple fix was interpolating between amplitude levels, which is less obvious in the non-uniform-boundary case.  If you want to work on it, that would be welcome.

**Comment by bmcfee** [comment]
> This would be nice to fix, at some point. The simple fix was interpolating between amplitude levels, which is less obvious in the non-uniform-boundary case.

Eh, it's not too hard if you compute zc on the single-cycle buffer, compute time mod cycle-length, round to the nearest zc, and add remainder back to the original time.

But that's for another day!

**Comment by craffel** [comment]
> Eh, it's not too hard if you compute zc on the single-cycle buffer, compute time mod cycle-length, round to the nearest zc, and add remainder back to the original time.

Happy to let you to embark on that journey.

**Comment by craffel** [comment]
Rebase it

**Comment by craffel** [comment]
Wait a minute, just realized that because you created the branch in my own repo, I can rebase for you!  You've given me too much power here...  Just changed one of your commit titles, too.

File: mir_eval/sonify.py [hunk 1]
```diff
@@ -81,36 +82,56 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
...
+            using array indexing tricks.
         """
-        # Generate ten periods at this frequency
-        ten_periods = int(10*fs*(1./frequency))
-        short_signal = function(2*np.pi*np.arange(ten_periods)*frequency/fs)
-        # Repeat the signal until it's of the desired length
+        # Generate one period at this frequency
+        short_signal = function(np.linspace(0, 2 * np.pi,
+                                            num=fs/float(frequency),
+                                            endpoint=False))
```
craffel: Using `linspace` like this results in the wrong timebase, resulting the frequencies being detuned.  Assuming an integer sampling index `n`, if you want amplitude of a signal of frequency `f` with sampling rate `fs`, you need to evaluate `function` at `2*np.pi*n*frequency/fs`.  This `linspace` will evaluate `function` at `2*np.pi*n/int(fs/frequency)`:
```Python
fs = 11
frequency = 1.9
print np.linspace(0, 2 * np.pi, num=fs/float(frequency), endpoint=False)
print 2*np.pi*np.arange(5)*frequency/fs
print 2*np.pi*np.arange(5)/int(fs/frequency)
```
```
[ 0.          1.25663706  2.51327412  3.76991118  5.02654825]
[ 0.          1.08527746  2.17055492  3.25583239  4.34110985]
[ 0.          1.25663706  2.51327412  3.76991118  5.02654825]
```
I'm not sure how to coerce `linspace` to do the right thing here.

bmcfee: Derp, you're right.  Will revert.

craffel: Ok, don't mean to get super pedantic here, but when you're synthesizing a single period and repeating it, the effective frequency of your signal is `fs/int(fs/frequency)`.  For low frequencies, this isn't too different from `frequency`; for high frequencies it can be very different.  For example, for MIDI note 90 (~1480 hz) with `fs = 22050`, `fs/int(fs/frequency)` is 1575 Hz.  That's why I synthesized ten periods; this was a quick hack to make the effective frequency `10*fs/int(10*fs/frequency)` which for MIDI note 90 results in about 1490 Hz.  It seemed like a good trade-off as an alternative to adding some logic to keep track of the sample error when you're repeating `short_signal`.  I'm open to ideas if you have a better one (one would be to use a very-fast-to-compute pure Python triangle wave implementation or something, and don't do any repeat tricks).  For reference, here are plots of the semitone error of all MIDI notes from 0 to 110 (assuming notes above 110 are a wash) for repeating a single period vs. repeating 10 periods:
```Python
fs = 22050.
def plot(errs):
    plt.figure(figsize=(15, 8))
    plt.xticks(np.arange(0, 128, 4))
    plt.plot(errs)
    plt.grid()
    plt.xlabel('MIDI Note')
    plt.ylabel('Semitone error')
plot([librosa.hz_to_midi(fs/int(fs/librosa.midi_to_hz(f))) - f
      for f in np.arange(0, 110)])
plt.title('1 period')
plot([librosa.hz_to_midi(10*fs/int(10*fs/librosa.midi_to_hz(f))) - f
      for f in np.arange(0, 110)])
plt.title('10 periods')
```
download 4
download 7

craffel:
Oops, it hid my [most recent comment](https://github.com/craffel/mir_eval/pull/163#discussion_r52192914).  With regards to my suggestion

> one would be to use a very-fast-to-compute pure Python triangle wave implementation or something, and don't do any repeat tricks

This actually is not a great idea, because the fastest-possible function I can think of would be
`2*np.mod(frequency*np.arange(int(length*fs))/fs, 1) - 1`, which it turns out is about O(10^3) times slower than `_fast_synthesize`.  So, I'm open to ideas here, if you are not happy with the "synthesize 10 periods" variant.

bmcfee:
> So, I'm open to ideas here, if you are not happy with the "synthesize 10 periods" variant.

Seems like the best of both worlds would be to combine the in-place tiling hack with a 10-period buffer.  This way, we minimize frequency deviations, but don't pay the price of copying a buffer thousands of times.  Sound good?

bmcfee:
I reverted back to the 10-cycle buffer.  The buzzing is back, of course, but otherwise everything's ready to go on this one.

craffel:
> I reverted back to the 10-cycle buffer. The buzzing is back, of course, but otherwise everything's ready to go on this one.

Well, to be fair, the buzzing should be there either way, it's just that in the 1-period case it has the same period as the wave being synthesized as opposed to 10x the period.  To mitigate this, I think an OK solution would be to synthesize, say, 30 periods of the wave and then copy `n` periods where `10 <= n <= 30` is chosen empirically as 
```Python
n_vals = int(np.arange(10, 30)*fs/frequency)
n = np.argmin(np.abs(short_signal[0] - short_signal(2*np.pi*(n_vals + 1)*frequency/fs))
```
Of course, both 10 and 30 should be user-settable parameters because they trade off tuning accuracy/potential suppression of clicks with execution time.  But we can leave that for another time.

PR #164, opened 2016-02-04
Fix Travis

craffel:
Removing pypi channel.


craffel:
This looks like it fixes it.  @bmcfee it looks like this line I removed to stop it failing is a holdover from some of your repositories, and that it isn't needed to keep things running.  Are your other repositories also failing?  I think all of the packages we want are available via conda, so adding a hook to pypi isn't needed.

bmcfee:
:ship:

PR #169, opened 2016-02-04
[WIP] Make Travis faster

craffel:
Avoid timeouts.


craffel:
It is the source separation metrics which run insanely slow on Travis.  Having conda install atlas doesn't help.  Switching back to `np.fft` instead of `scipy.fftpack` doesn't help.  I can think of a few remaining options, none are great:

* Make travis skip the separation tests.  Let's not do this.
* Figure out a way to get nose to print things out from time to time, to prevent travis from quitting.  This is kind of ugly and doesn't avoid the fact that it will still take forever to run the tests.
* Shorten all the source separation cases (they are all currently 10s, which is already pretty damn short).  Maybe I can just crop the middle 1s from each of them?  This is the best option, IMO, right now.
* Some other travis voodoo to make FFT's faster (I assume it's the FFT, not the linalg.solve or toeplitz) that I don't currently know of.

@dawenl what's your take on only testing on 1s clips?  Does that cover bss_eval?  @dpwe would be great to get your input on this too.  @bmcfee please opine.

bmcfee:
opining:

- i'm not sure it's the fft; best to profile it and be certain before sinking time into optimizing.
- scipy.fftpack is generally preferred to np.fft.  If you're really hurting for fft speed, maybe fftw could help here, but i doub tit.
- @ejhumphrey showed me some magic to do timing as part of unit tests, but i can't remember what/how it worked.  splain?
- 10sec seems pretty long to me for a unit test, especially at high sampling rates.  I'm not totally familiar with the source separation metrics, but it looks like that toeplitz op is going to put you in at least quadratic (in signal length) time/space.  The least squares on that is now pushing into cubic land.. so trimming down these input signals should give you super-linear speedup.  Likewise for dropping the sampling rate.

craffel:
> * i'm not sure it's the fft; best to profile it and be certain before sinking time into optimizing.
scipy.fftpack is generally preferred to np.fft. If you're really hurting for fft speed, maybe fftw could help here, but i doub tit.
@ejhumphrey showed me some magic to do timing as part of unit tests, but i can't remember what/how it worked. splain?

Having Travis produce a profile would be very useful, but I don't know how to tell nose to profile a test.

> 10sec seems pretty long to me for a unit test, especially at high sampling rates. I'm not totally familiar with the source separation metrics, but it looks like that toeplitz op is going to put you in at least quadratic (in signal length) time/space. The least squares on that is now pushing into cubic land.. so trimming down these input signals should give you super-linear speedup. Likewise for dropping the sampling rate.

Well, philosophically I'd like these tests to be testing the kind of data they see in real-world usage, and 10-second clips is already on the (very) short range for duration, and 16 kHz is already on the (very) low range for sampling rate.  This is the most obvious and probably most effective solution, but it doesn't sit well.

craffel:
No response from @dpwe and @dawenl means "go ahead", so I've shortened all the separation test clips to 1s.  We'll see how much this speeds things up.

craffel:
:laughing: tests sped up to ~3 minutes on travis, guess we'll go with this even though it makes me a little sad.  Merging.

PR #173, opened 2016-02-11
Bug-fix for sonification of chords

### LINKED ISSUE #172: sonify.chords, unique times strikes again
There appears to be a bug in sonify.chords stemming from running unique on time events.  It's difficult to reproduce, since it boils down to numerical precision issues, but the resulting artifact is that some chords do not get sonified.

I've gisted a [notebook](https://gist.github.com/829701891958b7f9ae41) to illustrate the problem, but it depends on having the jams-sonify branch and beatles isophonics data, so it may take some effort to reproduce outside of my environment.  Still, I'm pretty sure this is caused by our old buddy [unique](https://github.com/craffel/mir_eval/blob/master/mir_eval/sonify.py#L213).

bmcfee:
Fixes #172

### DIFF
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -68,9 +68,9 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
     frequencies : np.ndarray
         array of size ``gram.shape[0]`` denoting the frequency of
         each row of gram
-    times : np.ndarray
-        array of size ``gram.shape[1]`` denoting the start time of each
-        column of gram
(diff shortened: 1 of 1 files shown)


mir_eval/sonify.py
```diff
@@ -69,8 +69,8 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
         array of size ``gram.shape[0]`` denoting the frequency of
         each row of gram
     times : np.ndarray
```
craffel: While you're at it, why not put the same shape spec here too?

bmcfee: Done.

PR #178, opened 2016-02-15
hierarchy.evalute() supports labels (NO-OP)

### LINKED ISSUE #177: hierarchy.evaluate(), forward-compatibility
Dragging back up some discussion from #125...

The current `evaluate` function calls only one metric (tmeasures), which in turn only takes boundaries and not labels.  Consequently, `evaluate` only exposes positional arguments for boundaries.

I can imagine the not-too-distant future`*` in which we may have metrics for label hierarchies, which will require a change to the `evaluate` API.  The resulting api should look similar to `segment.evaluate()`, excepting that each of the inputs is a list of annotations (intervals or labels) instead of a single flat interval/label array.

The question here is: when should we change the API?

- Option 1: ASAP, to prevent too much breakage down the road
- Option 2: wait until a metric using the labels is introduced.

I'd advocate for the former, since it costs us nothing and will make the transition easier later on.  What do others think?

`*` I have an implementation of one such metric already, but it's nowhere near ready for PR/inclusion.

bmcfee:
Adds API for labels in `hierarchy.evaluate`.  Fixes #177

### DIFF
--- a/evaluators/segment_hier_eval.py
+++ b/evaluators/segment_hier_eval.py
@@ -74,11 +74,16 @@ if __name__ == '__main__':
     ref_files = parameters['reference_file']
     est_files = parameters['estimated_file']
 
-    ref_intervals = [load_labeled_intervals(_)[0] for _ in ref_files]
-    est_intervals = [load_labeled_intervals(_)[0] for _ in est_files]
+    ref = [load_labeled_intervals(_) for _ in ref_files]
+    est = [load_labeled_intervals(_) for _ in est_files]
+    ref_intervals = [seg[0] for seg in ref]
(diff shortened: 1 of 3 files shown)


bmcfee:
I think this is ready, unless @craffel wants some docstring that indicates no current metrics use the labels.

craffel:
> I think this is ready, unless @craffel wants some docstring that indicates no current metrics use the labels.

No need at this point, I don't think.

The salami example is getting a bit hairy.  I can't think of a more direct way to do that though.

bmcfee:
> The salami example is getting a bit hairy. I can't think of a more direct way to do that though.

Agreed on both counts.  Life would be easier if salami had a sane format.

PR #181, opened 2016-02-25
WIP: Key detection

craffel:
Adding functionality for key detection evaluation.  AFAIK the only metric is the one described [here](http://www.music-ir.org/mirex/wiki/2015:Audio_Key_Detection#Evaluation_Procedures). This is overall pretty simple to implement, with the following caveats:

* It states that a "Perfect fifth" should receive a score of 0.5.  Looking at the [IMIRSEL code](https://github.com/jpauwels/nemadiy/blob/master/analytics/trunk/src/main/java/org/imirsel/nema/analytics/evaluation/key/KeyEvaluator.java), I believe this is only triggered when the estimated key is a perfect fifth above reference key, but _not_ when the reference key is a perfect fifth above estimated key.  This doesn't make a lot of sense to me because it seems like this relationship should be symmetric - if it's ".5 score" to estimate the key as a perfect fifth above, why is it "score 0" bad to estimate the key as a perfect fifth below?  I've followed this convention and have made it a little clearer in the docs, but I'd like to get someone who actually has a music theory background to weigh in.  @ejhumphrey @justinsalamon @urinieto maybe?
* This metric could actually be implemented very efficiently as a 24x24 lookup table, indexed by reference and estimated key, which could be constructed very quickly at import-time.  However, I think the current conditional version is more readable, and is also probably plenty fast.
* I do not have algorithm output + ground truth + MIREX score for this task, but I did compare the behavior for every ref/est key pair against a Python port of the IMIRSEL code, see [here](https://gist.github.com/craffel/004dc82b7557a8ad484b).  Of course, my port could be wrong, but as I said it's a pretty straightforward metric.  I made up synthetic data for regression testing, because I would argue in this case synthetic data is about as useful as real data.  It would be great to have real data though.

Any eyes on this would be great!  Hoping to add this to the impending 0.3 release.


bmcfee:
I just took a quick skim through, and it looks good.

The rest of your comments make sense.

It might be good to add documentation precisely explaining what strings are valid for "key" and "mode".  For instance, would "C, maj" work, or does it have to be "C, major"?

I bring this up because "mode" gets used differently in different data sets.  In JAMS we went [holistic](https://github.com/marl/jams/blob/master/jams/schemata/namespaces/key/key_mode.json) on this, and lumped major/minor in with greek scales.  This was maybe not the best decision, but we needed something like it to support isophonics data.  I imagine this could cause trouble in eval.

craffel:
> It might be good to add documentation precisely explaining what strings are valid for "key" and "mode". For instance, would "C, maj" work, or does it have to be "C, major"?

Yes, I thought about this.  To me, the sanest format (for major/minor only) is 0-23, but I decided just to follow MIREX exactly (so only `'* major'` or `'* minor'`).  I will make this a bit clearer in the module-level docstring.

justinsalamon:
@craffel I haven't looked at the code (or the mirex code) so I'm not sure if this is relevant, but here goes:

I imagine that if the estimated key is a perfect fifth above the true key then you get 0.5 because they are highly related (tonally). The [circle of fifths](https://en.m.wikipedia.org/wiki/Circle_of_fifths) is not symmetric: the fifth of C is G, but the fifth of G is D. So if the true key is a C, estimating a G (a fifth) is closely related, but estimating a perfect fifth down (an F) is not as tonally related (F is the 4th for C, not the 5th) and hence not rewarded at all.

Estimating a G *below* the C (ie a perfect 4th down) should still get rewarded, but since this is key estimation I imagine this is octave agnostic right? Hence you only have to look for a fifth "up" (in reality its a chroma circle, so there is no up and down)

Writing from phone excuse typos

craffel:
> The circle of fifths is not symmetric: the fifth of C is G, but the fifth of G is D. So if the true key is a C, estimating a G (a fifth) is closely related, but estimating a perfect fifth down (an F) is not as tonally related (F is the 4th for C, not the 5th) and hence not rewarded at all.

So it's intuitive to you that if the ground truth is C and you estimate G you get 0.5; if the ground truth is G and you estimate C you get 0.0?  Because keys are only highly related when you traverse the circle of fifths clockwise (or something)?

> Estimating a G below the C (ie a perfect 4th down) should still get rewarded, but since this is key estimation I imagine this is octave agnostic right? Hence you only have to look for a fifth "up" (in reality its a chroma circle, so there is no up and down)

Yes, there's no notion of octave since it's key detection.

justinsalamon:
> So it's intuitive to you that if the ground truth is C and you estimate G you get 0.5; if the ground truth is G and you estimate C you get 0.0?

Yes. C is not the same to G as G is to C.

craffel:
> Yes. C is not the same to G as G is to C.

Ok, I'll leave it as is.

craffel:
Any other comments/eyes on this before I merge?

PR #182, opened 2016-02-26
Add tempo to docs (and minor fix(es))

craffel:
The tempo submodule never got added to index.rst.  There was also some minor formatting inconsistencies in the submodule-level docstring.  Finally, there was an unused `warnings` import; may as well take care of that here.  @bmcfee any objections?


bmcfee:
No objections here.

Issue #185, opened 2016-03-04
Implement multi-f0 frame level eval metrics


rabitt:
As far as I can tell, MIREX computes 14 metrics. Each of the following 7 metrics are computed first for the original values, and second for octave-wrapped values:
- Accuracy
- Precision
- Recall
- Total Error (E_tot)
- Substitution Error (E_sub)
- Miss Error (E_miss)
- False Alarm Error (E_fa)

Metrics described in part [here, page 5](http://academiccommons.columbia.edu/download/fedora_content/download/ac:144450/CONTENT/1687-6180-2007-048317.pdf).

I need this for my own work and am happy to/planning to implement it.
@craffel - what specific things would you want me to cover for a squeakly clean PR? I imagine code with tests + comparison with MIREX code on MIREX datasets? :)

craffel:
Great!

> @craffel - what specific things would you want me to cover for a squeakly clean PR? I imagine code with tests + comparison with MIREX code on MIREX datasets? :)

We should document this somewhere when there are more requests like this, but the idea PR has
- metrics implemented in mir_eval/metric_name.py, plus `validate` method which makes sure annotations are in a sane format and `evaluate` method which calls all metrics and returns an ordered dict of the results (see other metrics for examples)
- loader in `io` which loads in MIREX-format data
- evaluator in the evaluators folder which loads in data, calls `evaluate`, and writes out a json results file
- Docs everywhere in numpydoc format, with an auto-module hook added to docs/index.rst
- Unit tests and regression tests for all metrics, with a regression test data ideally coming from some real algorithm's output and human-annotated examples, as well as a generating hook for regression output in tests/generate_data.py
- For the PR (but doesn't have to live permanently in the repository), ideally a comparison between the metric's output and some other implementation (e.g. MIREX); for examples, see https://github.com/craffel/mir_eval-ismir)

Hopefully that doesn't sound like an overwhelming amount - apart from implementing the metrics themselves and comparing the output to an existing implementation, most things should be pretty quick and there are examples here!

rabitt:
@craffel great! Seems straightforward enough ;)

rabitt:
@craffel: I ran into a small hitch and want to run it by you.
As far as I've seen, multi-f0 algorithms report scores as a time stamp followed by a variable number of f0 values:
```
29.52	246.94	622.25	311.13	246.94
29.53	246.94	622.25	311.13	246.94
29.54	622.25	311.13	246.94
[...]

PR #186, opened 2016-03-16
Implement multi-f0 frame level eval metrics

rabitt:
Implements #185.
The following metrics are implemented for raw and chroma-wrapped multipitch estimates:
- [x] Accuracy
- [x] Precision
- [x] Recall
- [x] Total Error (E_tot)
- [x] Substitution Error (E_sub)
- [x] Miss Error (E_miss)
- [x] False Alarm Error (E_fa)

@craffel's to-do list:
- [x] metrics implemented in mir_eval/metric_name.py, plus validate method which makes sure annotations are in a sane format and evaluate method which calls all metrics and returns an ordered dict of the results (```mir_eval/multipitch.py```)
- [x] loader in io which loads in MIREX-format data (```mir_eval.io.load_ragged_time_series```)
- [x] evaluator in the evaluators folder which loads in data, calls evaluate, and writes out a json results file (```evaluators.multipitch_eval.py```)
- [x] Docs everywhere in numpydoc format
- [x] auto-module hook added to docs/index.rst
- [x] Unit tests and regression tests for all metrics (```tests/test_multipitch.py```)
- [x] a regression test data ideally coming from some real algorithm's output and human-annotated examples (```tests/data/multipitch/```, ```test_multipitch.regression_test_evaluate```)
- [x] a generating hook for regression output in tests/generate_data.py


mir_eval/io.py
```diff
@@ -477,3 +477,79 @@ def load_key(filename, delimiter=r'\s+'):
...
+def load_ragged_time_series(filename, dtype=float, delimiter=r'\s+'):
+    r"""Utility function for loading in data from a delimited time series
+    annotation file with a variable number of columns.
+
+    Parameters
+    ----------
+    filename : str
+        Path to the annotation file
+    dtype : function
+        Data type to apply to columns 1 through n.
```
craffel: Is this 0- or 1-indexed?

mir_eval/io.py
```diff
@@ -477,3 +477,79 @@ def load_key(filename, delimiter=r'\s+'):
...
+    delimiter : str
+        Separator regular expression.
+        By default, lines will be split by any amount of whitespace.
+
+    Returns
+    -------
+    times : np.ndarray
+        array of timestamps (float)
+    values : list of np.ndarrays
```
craffel: "list of np.ndarray"

mir_eval/io.py
```diff
@@ -477,3 +477,79 @@ def load_key(filename, delimiter=r'\s+'):
...
+        own_fh = True
+        # Open the file for reading
+        input_file = open(filename, 'r')
+    # If the provided has a read attribute, we can use it as a file handle
+    elif hasattr(filename, 'read'):
+        input_file = filename
+    # Raise error otherwise
+    else:
+        raise ValueError('filename must be a string or file handle')
+    for row, line in enumerate(input_file, 1):
```
craffel: You are ignoring the first line of the file here, correct?  How come?  This should be settable by a function argument, I think.  Or documented.

mir_eval/io.py
```diff
@@ -477,3 +477,79 @@ def load_key(filename, delimiter=r'\s+'):
...
+        try:
+            converted_time = float(data[0])
+        except:
+            raise ValueError("Couldn't convert value {} using {} "
+                             "found at {}:{:d}:\n\t{}".format(
+                                 data[0], float.__name__, filename, row,
+                                 line))
+        times.append(converted_time)
+
+        if len(data) > 1:
```
craffel: You don't need if/else here, because e.g. `['a'][1:]` will evaluate to `[]`.  I would also add a comment here, because comments are great.

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
+'''
+The goal of multiple f0 (multipitch) estimation and tracking is to identify
+all of the active fundamental frequencies in each time frame in a complex music
+signal.
+
+Conventions
+-----------
+Multipitch estimates are represented by a time series, and a corresponding list
+of arrays of frequency estimates. Frequency estimates may have any number of
+frequency values, including (represented by an empty array). Time values are in
```
craffel: "including 0"

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
+'''
+The goal of multiple f0 (multipitch) estimation and tracking is to identify
+all of the active fundamental frequencies in each time frame in a complex music
+signal.
+
+Conventions
+-----------
+Multipitch estimates are represented by a time series, and a corresponding list
```
craffel: So to me, a "time series" means any sequence of temporal data, not the times at which that data occurs.  I would call the times at which some sequential temporal data occurs the "timebase".  Am I alone here?

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+series that are outside of the range of the reference time series are given
+null (empty array) frequencies.
+
+By default, a frequency is "correct" if it is within 0.5 semitones of a
+reference frequency. The metrics are based on those described in
+[#poliner2007]_ and [#bay2009]_.
+
+Metrics
+-------
+* :func:`mir_eval.multipitch.accuracy`: Precision, Recall, and Accuracy scores
```
craffel: Unfortunately we don't have a good convention as to what PRF metrics should be called.  You are just reminding me that we should make one.

docs/index.rst
```diff
@@ -129,6 +129,14 @@ The following subsections document each submodule.
    :show-inheritance:
    :member-order: bysource
 
+:mod:`mir_eval.multipitch`
```
craffel: Hopefully multipitch is an [intentional reference](https://en.wikipedia.org/wiki/Multi-pitch_climbing)...

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+
+import numpy as np
+import collections
+from scipy.interpolate import interp1d
+from . import util
+import warnings
+
+
+MAX_TIME = 30000.  # The maximum allowable time stamp (seconds)
+MAX_FREQ = 5000.  # The maximum allowable frequency (Hz)
```
craffel: Is this really the highest frequency anyone would ever annotate for multipitch?  Is this written down anywhere?

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+    ref_freqs : list of np.ndarrays
+        reference frequencies in Hz
+    est_time : np.ndarray
+        estimate time stamps in seconds
+    est_freqs : list of np.ndarrays
+        estimated frequencies in Hz
+
+    """
+    if ref_time.size == 0:
+        raise ValueError("Reference frequencies are empty.")
```
craffel: "times", not "frequencies".  Also, we usually issue a warning here and have the metric return some sane limiting value (e.g. for PRF it's 0, 0, 0).

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+        reference time stamps in seconds
+    ref_freqs : list of np.ndarrays
+        reference frequencies in Hz
+    est_time : np.ndarray
+        estimate time stamps in seconds
+    est_freqs : list of np.ndarrays
+        estimated frequencies in Hz
+
+    """
+    if ref_time.size == 0:
```
craffel: Since you're using `size`, it makes sense to put another check that `ref_time.ndim == 1` (right?).

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+    if len(est_freqs) == 0:
+        raise ValueError("Estimated frequencies are empty.")
+    if ref_time.size != len(ref_freqs):
+        raise ValueError('Reference times and frequencies have unequal '
+                         'lengths.')
+    if est_time.size != len(est_freqs):
+        raise ValueError('Estimate times and frequencies have unequal '
+                         'lengths.')
+
+    util.validate_events(ref_time, max_time=MAX_TIME)
```
craffel: Put `validate_events` above the checks above, since you are using `size`.  I appreciate you using this here, although we should consider the fact that the wording is a little weird (calling the timebase events).

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+    util.validate_events(est_time, max_time=MAX_TIME)
+
+    for freq in ref_freqs:
+        util.validate_frequencies(freq, max_freq=MAX_FREQ, min_freq=MIN_FREQ)
+
+    for freq in est_freqs:
+        util.validate_frequencies(freq)
```
craffel: You don't pass values `max_freq` and `min_freq` here, is that intentional?

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+def resample_multipitch(times, frequencies, target_times):
+    """Resamples multipitch time series to a new timescale. Values in
+    'target_times' outside the range of 'times' return no pitch estimate.
+
+    Parameters
+    ----------
+    times : np.ndarrayre
```
craffel: Remove "re" at the end of the line.

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+    for freq in ref_freqs:
+        util.validate_frequencies(freq, max_freq=MAX_FREQ, min_freq=MIN_FREQ)
+
+    for freq in est_freqs:
+        util.validate_frequencies(freq)
+
+
+def resample_multipitch(times, frequencies, target_times):
+    """Resamples multipitch time series to a new timescale. Values in
+    'target_times' outside the range of 'times' return no pitch estimate.
```
craffel: These should be ` or ``, not ', I think

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+    .. [#poliner2007] G. E. Poliner, and D. P. W. Ellis, "A Discriminative
+    Model for Polyphonic Piano Transription", EURASIP Journal on Advances in
+    Signal Processing, 2007(1):154-163, Jan. 2007.
+    .. [#bay2009] Bay, M., Ehmann, A. F., & Downie, J. S. (2009). Evaluation of
+    Multiple-F0 Estimation and Tracking Systems. In ISMIR (pp. 315-320).
+'''
+
+import numpy as np
+import collections
+from scipy.interpolate import interp1d
```
craffel: Do you mind doing a `import scipy.interpolate` and then using it as `scipy.interpolate.interp1d`?

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+        Array of target time stamps
+
+    Returns
+    -------
+    frequencies_resampled : list of lists
+        Frequency list of lists resampled to new timebase
+    """
+    n_times = len(frequencies)
+
+    frequency_index = np.arange(0, n_times)
```
craffel: I would love some comments here!

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+
+    """
+    n_frames = len(ref_freqs)
+    true_positives = np.zeros((n_frames, ))
+
+    for i, (ref_frame, est_frame) in enumerate(zip(ref_freqs, est_freqs)):
+        if len(ref_frame) == 0 or len(est_frame) == 0:
+            true_positives[i] = 0.0
+        else:
+            matching = util.match_events(ref_frame, est_frame, window)
```
craffel: Nice use of `match_events` here, but the terminology is confusing!  Maybe we should consider changing the name of `match_events`.

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+    true_positives : np.ndarray
+        Array the same length as ref_freqs containing the number of true
+        positives.
+
+    """
+    n_frames = len(ref_freqs)
+    true_positives = np.zeros((n_frames, ))
+
+    for i, (ref_frame, est_frame) in enumerate(zip(ref_freqs, est_freqs)):
+        if len(ref_frame) == 0 or len(est_frame) == 0:
```
craffel: You don't need a special case here I don't think.
```
>>> mir_eval.util.match_events(np.array([]), np.array([]), 0.5)
[]
```

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+            true_positives[i] = 0.0
+        else:
+            matching = util.match_events(ref_frame, est_frame, window)
+            true_positives[i] = len(matching)
+
+    return true_positives
+
+
+def accuracy(true_positives, n_ref, n_est):
```
craffel: Hmm, usually metrics are called with annotations directly, not with an intermediate score.  Is the disadvantage just that `compute_num_true_positives` would have to be computed twice?  How slow is `compute_num_true_positives`? Can we just combine the two metrics?

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+
+Estimate time series should ideally be equal to reference time series, but if
+this is not the case, the estimate time series is resampled using a nearest
+neighbor interpolation to match the estimate. Time values in the estimate time
+series that are outside of the range of the reference time series are given
+null (empty array) frequencies.
+
+By default, a frequency is "correct" if it is within 0.5 semitones of a
+reference frequency. The metrics are based on those described in
+[#poliner2007]_ and [#bay2009]_.
```
craffel: You should mention chroma eval here somewhere.

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+        precision = 0.0
+
+    n_ref_sum = n_ref.sum()
+    if n_ref_sum > 0:
+        recall = true_positive_sum/n_ref.sum()
+    else:
+        warnings.warn("Reference frequencies are all empty.")
+        recall = 0.0
+
+    acc_denom = (n_est + n_ref - true_positives).sum()
```
craffel: We have `util.f_measure` for this part.

mir_eval/multipitch.py
```diff
@@ -0,0 +1,383 @@
...
+        Substitution error
+    e_miss : float
+        Miss error
+    e_fa : float
+        False alarm error
+    e_tot : float
+        Total error
+
+    """
+    n_ref_sum = float(n_ref.sum())
```
craffel: Comments here would be cool.

mir_eval/util.py
```diff
@@ -686,6 +686,40 @@ def validate_events(events, max_time=30000.):
         raise ValueError('Events should be in increasing order.')
 
 
+def validate_frequencies(frequencies, max_freq=5000., min_freq=20.):
+    """Checks that a 1-d frequency ndarray is well-formed, and raises
+    errors if not. Allows negative frequency values.
```
craffel: How come it allows negative frequency values?  Wouldn't they throw off your eval?

tests/test_multipitch.py
```diff
@@ -0,0 +1,285 @@
...
+     actual_etot) = mir_eval.multipitch.error_score(
+         true_positives, n_ref, n_est)
+
+    assert np.allclose(actual_esub, expected_esub, atol=A_TOL)
+    assert np.allclose(actual_emiss, expected_emiss, atol=A_TOL)
+    assert np.allclose(actual_efa, expected_efa, atol=A_TOL)
+    assert np.allclose(actual_etot, expected_etot, atol=A_TOL)
+
+
+def unit_test_evaluate():
```
craffel: Usually what we do is unit test the individual metrics instead of `evaluate`, just because then we know each of the individual metrics, when called, will raise the same errors (or different ones if individual metrics raise their own errors).

tests/test_util.py
```diff
@@ -160,6 +160,54 @@ def test_bipartite_match():
         assert v in G[k] or k in G[v]
 
 
+def test_validate_intervals():
+    # Test for ValueError when interval shape is invalid
```
craffel: Thanks for adding this, but why now?

tests/test_util.py
```diff
@@ -160,6 +160,54 @@ def test_bipartite_match():
...
+    nose.tools.assert_raises(
+        ValueError, mir_eval.util.validate_intervals,
+        np.array([[1.], [2.5], [5.]]))
+    # Test for ValueError when times are negative
+    nose.tools.assert_raises(
+        ValueError, mir_eval.util.validate_events,
```
craffel: A little strange to test `validate_events` in `test_validate_intervals`.

mir_eval/util.py
```diff
@@ -686,6 +686,40 @@ def validate_events(events, max_time=30000.):
         raise ValueError('Events should be in increasing order.')
 
 
+def validate_frequencies(frequencies, max_freq=5000., min_freq=20.):
```
craffel: So, I'd actually prefer not to have default values for these kwargs here, because they are already defined as top-level CONSTANTS in multipitch.

tests/test_util.py
```diff
@@ -160,6 +160,54 @@ def test_bipartite_match():
...
+    # Test for ValueError when event times are not increasing
+    nose.tools.assert_raises(
+        ValueError, mir_eval.util.validate_events,
+        np.array([1., 2., 5., 3.]))
+
+
+def test_validate_frequencies():
+    # Test for ValueError when max_freq is violated
+    nose.tools.assert_raises(
+        ValueError, mir_eval.util.validate_frequencies,
```
craffel: Assuming that you change `validate_frequencies` to not have kwargs with default values, you'd also need to pass a max/min freq here.

craffel:
Hey, this looks great, thanks a lot! I just made a lot of minor comments.  The only larger questions I have:

1. As I mentioned in one comment, it would be more intuitive to me if the metrics were called with timebase/list-of-multipitch-array pairs, which were on the same exact timebase (resampling being considered the only, potentially optional, preprocessing step).  This matches the convention to the most of `mir_eval`, e.g. that the eval is not done in a few stages.  It may result in some duplicated computation, which I am fine with ignoring if the duplication is very low overhead, but I would also potentially be fine with both of these metrics being wrapped into one (although that would be a lot of scores reported by a single function).  Thoughts?
1. Any work yet on comparing this to MIREX or some other existing code?

mir_eval/io.py
rabitt: 0 indexed. I've updated the doc to make this more clear

mir_eval/io.py
rabitt: Good catch - I didn't mean to ignore the first line.

By the way, that part was copy-pasted from ```io.load_delimited```. Is it intentional to ignore the header in general on that function?

mir_eval/io.py
rabitt: :+1:

mir_eval/multipitch.py
rabitt: Not alone at all. Updated!

mir_eval/multipitch.py
rabitt: Good question - I wondered the same thing :) As far as I can tell, MIREX uses precision/recall/accuracy, and I implemented everything to match what they do. We can throw in f-measure though?

mir_eval/multipitch.py
rabitt: 5000 is the nearest "clean" frequency value above C8, which is the highest piano note. I think it's a reasonable "highest value", but it came from my brain, and not from any convention. I know in melody land a much lower value is often used.

We could use 20,000 Hz, but it makes it harder to catch "bad" frequency units like cents.

mir_eval/multipitch.py
rabitt: oops!

mir_eval/multipitch.py
rabitt: Indeed. I added a comment to help avoid confusion for this use case.

mir_eval/multipitch.py
rabitt: > Is the disadvantage just that compute_num_true_positives would have to be computed twice?

yes, it could in theory be quite slow, especially if used on a full track. It's essentially performing an onset detection eval for every time stamp (albeit over a smaller number of "onsets" each time).

> Can we just combine the two metrics?

Sure. I'm a bit confused though. In ```mir_eval/melody.py``` the metrics are all written as separate functions, while in ```mir_eval/onsets.py``` they're bundled together in ```f_measure```. What's your preference? I could put all 7 metrics in one big ```metrics``` function, or I could break them into 7 functions. If the latter, I'd definitely want to pass ```true_positives``` as an input for efficiency's sake.

mir_eval/util.py
rabitt: Hmm...true it would. I was thinking about the melody use case, where they have the convention that negative pitches are reported during "unvoiced" frames. I'll add a ```allow_negatives``` argument to this function.

tests/test_multipitch.py
rabitt: Ok - I'll update this once we settle on how the metrics are organized.

tests/test_util.py
rabitt: Because I saw it was missing ;). I suppose it should have been part of a separate pr.

tests/test_util.py
rabitt: That would be a copy-pasta mistake. Fixed now.

rabitt:
@craffel I went through and made most of the changes you suggested. I had a couple of comments/questions on others. Can you take a look?

> As I mentioned in one comment, it would be more intuitive to me if the metrics were called with timebase/list-of-multipitch-array pairs, which were on the same exact timebase (resampling being considered the only, potentially optional, preprocessing step). This matches the convention to the most of mir_eval, e.g. that the eval is not done in a few stages. It may result in some duplicated computation, which I am fine with ignoring if the duplication is very low overhead, but I would also potentially be fine with both of these metrics being wrapped into one (although that would be a lot of scores reported by a single function). Thoughts?

I definitely don't want to duplicate the one big computation (```compute_true_positives```) because it could potentially be a bottleneck as is. I personally like how I've currently implemented it, but I see how it's inconsistent with the existing ```mir_eval``` conventions. An exception that already exists is that the metrics called in ```melody.evaluate``` have a fair bit of preprocessing (```melody.to_cent_voicing```) happening beforehand as well, though it's closer to the raw input than the ```num_true_positives``` array...
Anyway, you decide what you want, and I'll update it accordingly :)

> Any work yet on comparing this to MIREX or some other existing code?

I have the code & the outputs and I just need to run it! Will update you on this asap

mir_eval/multipitch.py
craffel: Huh, interesting.  I am going to assume that some smart person who knows about multipitch eval made that decision a while ago and am ok with sticking with it.

mir_eval/multipitch.py
craffel: That's ok then.

mir_eval/multipitch.py
craffel: Responding to this on the issue-wide comments.

craffel:
Sorry for the delay, I am in China.  If anyone else wants to review, please do!

> I definitely don't want to duplicate the one big computation (compute_true_positives) because it could potentially be a bottleneck as is. I personally like how I've currently implemented it, but I see how it's inconsistent with the existing mir_eval conventions. An exception that already exists is that the metrics called in melody.evaluate have a fair bit of preprocessing (melody.to_cent_voicing) happening beforehand as well, though it's closer to the raw input than the num_true_positives array...
Anyway, you decide what you want, and I'll update it accordingly :)

So, the distinction I would make is that all of the other metrics do non-evaluation preprocessing before being passed to metrics (if that makes sense).  There are a few places in `mir_eval` where the preprocessing wants both reference and estimated annotations, but it doesn't actually ever compare them in the sense of computing a metric.  For example, beat trims beat annotations completely independently of one other; segmentation does preprocessing independently except they pad estimated annotations to match reference; melody annotations are preprocessed independently except resample estimated annotations to match reference, etc.  So there is some kind of line there.  To me, the option which makes most sense is to combine all of these metrics into a single function - the actual core comparison being done is the same for all of the metrics, so I think it's OK that they are grouped.  There's also basically 0 overhead to computing all the metrics at once.  The disadvantage, of course, is that the function would return a ton of values, but I am OK with that.  I would sort of look at it like PRF metrics - we could potentially have separate functions for all three, but since they all are simple ways of reporting the same comparisons, they make sense to be grouped into a single function.  What do you think?

> I have the code & the outputs and I just need to run it! Will update you on this asap

Cool, looking forward!

mir_eval/io.py
craffel: That is worrisome.  Creating a separate issue #187

rabitt:
> To me, the option which makes most sense is to combine all of these metrics into a single function - the actual core comparison being done is the same for all of the metrics, so I think it's OK that they are grouped. There's also basically 0 overhead to computing all the metrics at once. The disadvantage, of course, is that the function would return a ton of values, but I am OK with that. I would sort of look at it like PRF metrics - we could potentially have separate functions for all three, but since they all are simple ways of reporting the same comparisons, they make sense to be grouped into a single function. What do you think?

I've made the change and grouped it into one function. All that changed in the end is there's a new function ```multipitch.metrics``` that does what was being done within ```multipitch.evaluate```. Hope that fits with what you had in mind. Lmk if it doesn't.

I compared with the mirex results and found that my chroma metrics were pretty different from theirs. Did some digging and found a mistake in my implementation [here](https://github.com/craffel/mir_eval/pull/186/files#diff-80caf5b5c6d33042e5c92838ab23bd72R417). I'm comparing using euclidean distance, but it should be computing distance modulo 12... making the fix now. Once it's done I'll send along the mirex comparison as well.

rabitt:
@bmcfee can you take a look at [this diff](https://github.com/craffel/mir_eval/pull/186/commits/792d00e20941f269bbd24d72d93b474ea5740b80)?

mir_eval/util.py
```diff
@@ -589,13 +589,58 @@ def recurse(v):
...
+    Returns
+    -------
+    outer_distance : np.ndarray, shape=(n, m)
+        The outer 1d-euclidean distance.
+
+    """
+    return np.abs(np.subtract.outer(ref, est))
+
+
+def match_events(ref, est, window, outer_distance=_outer_distance):
```
bmcfee: It's generally bad form to have objects (eg functions) as default arguments, as things can get ugly in case of mutation.

The (only slightly unsavory) fix for this is to have a default argument of `None` and then override it with custom logic in the function.

This also makes the docstrings prettier, since you get `outer_distance=None` instead of `outer_distance=<function at 0x324235rfdfsad45>`.

Also, I'd recommend renaming `outer_distance` to `distance`.

mir_eval/util.py
```diff
@@ -616,9 +664,8 @@ def match_events(ref, est, window):
         ``matching[i] == (i, j)`` where ``ref[i]`` matches ``est[j]``.
 
     """
-
     # Compute the indices of feasible pairings
-    hits = np.where(np.abs(np.subtract.outer(ref, est)) <= window)
+    hits = np.where(outer_distance(ref, est) <= window)
```
bmcfee: Does it make more sense to parameterize this by function, rather than letting the user simply supply their own hits array?  Alternately, is the "outer" distance the level of abstraction you want, or would it be easier to allow a user to supply a point-wise dissimilarity function?

As it was originally written, this function did two things:

1. Find feasible matchings (hits) between ref and est under a specific definition of hit
2. Convert hits to a graph and find an optimal matching.

Maybe a better factorization of this functionality might be to just implement 2, and have a separate function for 1.  Then we can provide a few different hit-matrix constructors that all feed into the same matching alg.

I don't have strong opinions on this though, so maybe it's overkill to factor it that far.

rabitt: I thought about that - this abstracts the distance computation, but only to the minimal extent needed for my application ;). I figured it was best to keep it as simple as possible for now, and if someone has a use case in the future that needs more abstraction we can change it. Until then, I think it's overkill to abstract more. Just my 2 cents.

mir_eval/util.py
```diff
@@ -589,13 +589,58 @@ def recurse(v):
...
 
 
-def match_events(ref, est, window):
+def _outer_distance_mod_n(ref, est, modulus=12):
+    """Compute the absolute outer distance modulo n.
+    Using this distance, d(11, 0) = 1 (modulo 12)
+
+    Parameters
+    ----------
+    ref : np.ndarray, shape=(n,)
```
bmcfee: these shouldn't be denoted as times in the docs

tests/test_util.py
```diff
@@ -160,6 +160,105 @@ def test_bipartite_match():
         assert v in G[k] or k in G[v]
 
 
+def test_outer_distance_mod_n():
```
bmcfee: none of your test inputs are outside the range [0, 11].  Better fix that!

rabitt:
Alright @craffel as promised, a [comparison](https://gist.github.com/rabitt/cfc0a42966d8935308a5) with previous mirex results.

Including the important plot here for reference...

screenshot 2016-03-19 18 22 49

Notice that all of the 'raw' metrics are very close, while the chroma metrics are pretty different. I stared at the mirex implementation for a while and compared it with my version. I found one mistake, referenced above that I fixed, but beyond that I'm sort of stuped as to why they're so different. One interesting thing is that both "False alarm error" and "Miss error" are computed in terms of ```n_ref``` and ```n_est``` only, and therefore should be identical for chroma and raw metrics. In the mirex scores they are not identical...

craffel:
> Notice that all of the 'raw' metrics are very close, while the chroma metrics are pretty different. I stared at the mirex implementation for a while and compared it with my version. I don't see any obvious reasons based on the two implementations why they should be so different. One interesting thing to note is that both "False alarm error" and "Miss error" are computed in terms of n_ref and n_est only, and therefore should be identical for chroma and raw metrics. In the mirex scores they are not identical... My guess is that somehow, they compute an n_ref and n_est for chroma, but I don't see that anywhere in the implementation.

It's [this](https://github.com/jpauwels/nemadiy/blob/ce3787191871d52196e97e6e9eb8d37caf12c911/analytics/trunk/src/main/java/org/imirsel/nema/analytics/evaluation/multif0/MultiF0EstEvaluator.java) right?  I can take a look when I am done traveling.  One thing that has helped me is to port the java code line-for-line to Python and then compare intermediate values (of course if you are comfortable enough in Java you can just compare intermediate values without porting).  I am assuming the values on your graph are not percentages?  Are they just the difference between mir_eval and mirex scores?  If so, I would say for sure that .1 is much too big, we need to investigate!

rabitt:
> It's this right?

That's not the reference implementation I was using - I got the code used for mirex from Li Su who ran the multi-f0 task last year. The code is based on Zhiyao Duan's code [here](http://music.cs.northwestern.edu/includes/projects/multipitch/mpes_eval.zip). The code you pointed to doesn't have all of the measures reported in mirex - in particular the only chroma score it reports is chroma accuracy, so I don't think it's the same.

>  I am assuming the values on your graph are not percentages? Are they just the difference between mir_eval and mirex scores? If so, I would say for sure that .1 is much too big, we need to investigate!

Yep, the plot is just mir_eval scores minus mirex scores. I agree that 0.1 is too big, but....

After more digging, I think there is an inconsistency in the mirex results, and that my implementation might be ok after all. All of the metrics are functions of some combination of ```TP``` (number of true positives), ```n_ref``` (number of reference pitches) and ```n_est``` (number of estimated pitches). Breaking down each metric, they are functions of:
- Accuracy: ```TP```, ```n_ref```, ```n_est```
- Precision: ```TP```, ```n_est```
- Recall: ```TP```, ```n_ref```
- Total Error: ```TP```, ```n_ref```, ```n_est```
- Substitution Error: ```TP```, ```n_ref```, ```n_est```
- False Alarm Error: ```n_ref```, ```n_est```
- Miss Error: ```n_ref```, ```n_est```

The only difference (in my understanding) between chroma metrics and raw metrics is that ```TP``` changes. If that's true, then Substitution Error Chroma should be identical to Substitution Error, and similarly the False alarm error for chroma should be the same as the raw false alarm error. The [results for the "MIREX" dataset](http://www.music-ir.org/mirex/wiki/2015:Multiple_Fundamental_Frequency_Estimation_%26_Tracking_Results_-_MIREX_Dataset#Detailed_Results) do indeed have E_sub and E_fa as being identical for raw and chroma metrics. For the Su dataset however, ([results](http://www.music-ir.org/mirex/wiki/2015:Multiple_Fundamental_Frequency_Estimation_%26_Tracking_Results_-_Su_Dataset#Detailed_Results)), the chroma and raw metrics are different for E_sub and E_fa. (The notebook I sent was comparing against the results of the Su dataset.) I looked a few years back and the [2014](http://www.music-ir.org/mirex/wiki/2014:Multiple_Fundamental_Frequency_Estimation_%26_Tracking_Results#Detailed_Results), [2013](http://www.music-ir.org/mirex/wiki/2013:Multiple_Fundamental_Frequency_Estimation_%26_Tracking_Results#Detailed_Results), and [2012](http://www.music-ir.org/mirex/wiki/2012:Multiple_Fundamental_Frequency_Estimation_%26_Tracking_Results#Detailed_Results) results also have E_sub chroma = E_sub, and E_fa chroma = E_fa.

All of this to say, I think the Su 2015 chroma results have some mistake that explains the large difference in metrics.

I should also mention that I'm trying to tie a bow on this and use the "official" implementation for my ismir paper. I know you're traveling and can use the implementation in the branch if I need to, but if this can be merged sooner that would be ideal.

craffel:
> After more digging, I think there is an inconsistency in the mirex results, and that my implementation might be ok after all. 

Ok, we should confirm what the inconsistency in their code is then (potentially with them).

> The only difference (in my understanding) between chroma metrics and raw metrics is that TP changes. If that's true, then Substitution Error Chroma should be identical to Substitution Error, and similarly the False alarm error for chroma should be the same as the raw false alarm error. 

Do you mean False Alarm Error and Miss Error?  It looks like those are the ones which do not depend on TP in your list.

> The results for the "MIREX" dataset do indeed have E_sub and E_fa as being identical for raw and chroma metrics. 

Can you compare your scores on the MIREX dataset then rather than the Su dataset and see if they are the same? Or are you already doing that?

> For the Su dataset however, (results), the chroma and raw metrics are different for E_sub and E_fa.

Yes, that looks like a bug then.  You should tell them about it.

> All of this to say, I think the Su 2015 chroma results have some mistake that explains the large difference in metrics.

That explains those two, but not "chroma total error", which also differs quite a bit.  Ideally we'd know why there too.

> I should also mention that I'm trying to tie a bow on this and use the "official" implementation for my ismir paper. I know you're traveling and can use the implementation in the branch if I need to, but if this can be merged sooner that would be ideal.

For sure, I don't want to drag heels but hopefully you understand my being pedantic about making sure there are no unknown differences in implementation.  Ideally I'd like to have another code review before merging, preferably by someone else although I bet everyone is in ISMIR deadline mode.  I will be able to look it over again before the end of the week, but as I said I'd like to get implementation differences squared away first.

rabitt:
> > After more digging, I think there is an inconsistency in the mirex results, and that my implementation might be ok after all.

> Ok, we should confirm what the inconsistency in their code is then (potentially with them).

Ok, I'll drop them a line.

>> The only difference (in my understanding) between chroma metrics and raw metrics is that TP changes. If that's true, then Substitution Error Chroma should be identical to Substitution Error, and similarly the False alarm error for chroma should be the same as the raw false alarm error.

> Do you mean False Alarm Error and Miss Error? It looks like those are the ones which do not depend on TP in your list.

Ah yes, sorry, I did mean False Alarm error and Miss Error.

>> The results for the "MIREX" dataset do indeed have E_sub and E_fa as being identical for raw and chroma metrics.

> Can you compare your scores on the MIREX dataset then rather than the Su dataset and see if they are the same? Or are you already doing that?

I don't have the algorithm's output for MIREX- only for the Su dataset. I'll also ask for the MIREX results.

>> For the Su dataset however, (results), the chroma and raw metrics are different for E_sub and E_fa.

> Yes, that looks like a bug then. You should tell them about it.

Will do.

>> All of this to say, I think the Su 2015 chroma results have some mistake that explains the large difference in metrics.

> That explains those two, but not "chroma total error", which also differs quite a bit. Ideally we'd know why there too.

I think it actually explains all of the differing chroma metrics. If ```n_ref``` and ```n_est``` are different in their implementation, then all of the other chroma metrics will be different from mine since they all are functions of at least ```n_ref``` and ```n_est```.

>> I should also mention that I'm trying to tie a bow on this and use the "official" implementation for my ismir paper. I know you're traveling and can use the implementation in the branch if I need to, but if this can be merged sooner that would be ideal.

> For sure, I don't want to drag heels but hopefully you understand my being pedantic about making sure there are no unknown differences in implementation.

Yep, I do, and appreciate it ;).

>Ideally I'd like to have another code review before merging, preferably by someone else although I bet everyone is in ISMIR deadline mode. I will be able to look it over again before the end of the week, but as I said I'd like to get implementation differences squared away first.

Ok, given the current timeline, I'll probably bug the mirex folks after the ismir deadline and we can sort this then.

craffel:
> I think it actually explains all of the differing chroma metrics. If n_ref and n_est are different in their implementation, then all of the other chroma metrics will be different from mine since they all are functions of at least n_ref and n_est.

Makes sense.  Let me know what you find out!

rabitt:
@craffel - an update on this:
There was indeed a bug in the 2015 Su results, but the scores are still quite different from the ones I'm getting. After discussing with Li Su, it looks like it's a difference in how the metrics are aggregated, which brings up a bigger issue:

The mirex scores, at least for this task, are a "meta" metric, created by combining the intermediate scores for each track. For example, in onset detection, the "meta" score would take the number of TP/FP/FN for each track, sum them, and use the new value to compute "meta" P/R/F-measure.
On the other hand, I computed an average of the metrics across tracks.

We're in the process of digging into the individual track level scores and will see if those are equivalent. If they are, I guess I'm wondering what your stance is on how metrics are aggregated for multiple tracks. Computing metrics the mirex way would require somewhat substantial changes to the code structure overall, wouldn't it?

craffel:
> Computing metrics the mirex way would require somewhat substantial changes to the code structure overall, wouldn't it?

Good catch.  Yes, we assume all evaluations are on a track level, and then aggregated across tracks, as opposed to computing on a corpus level.  Unless there is major disagreement from the community, I'd like to keep it that way.

justinsalamon:
> Unless there is major disagreement from the community, I'd like to keep it that way.

+1

rabitt:
Li Su was kind enough to send me the mean-aggregated scores - the comparison is [here](https://gist.github.com/rabitt/41eb897fe74f41171891f9cf0b297e6a).

For 5 out of the 7 algorithms, the results are all within 1/100th of a percent of the mirex scores. For the other two, the difference is between 1/10th and 1 percentage point. I expect that this is due to the optimal vs. greedy graph matching. 

Are these scores close enough to pass the Raffel quality control test? :) Or should I dig into the differences a bit more?

craffel:
Nice, looking good.  If you're willing, usually I like to compute the absolute difference and normalize by the score mean, e.g. `|mirex_score - mir_eval_score|/(mirex_score + mir_eval_score)/2`.  I think most of your scores, apart from CB1 and CB2, are at that point within rounding error.  It would be great to look at the actual matchings produced for CB1 and CB2 to see if there's a difference between those matchings and greedy matchings (or compute them with a greedy matcher instead), just to isolate that difference, but yeah otherwise it looks good.  I would like to do another code review before merging, but am in thesis lockdown mode.  It would be ideal to have someone else to do it either way.  Can you track down someone to do a second one before merging?

rabitt:
> Nice, looking good. If you're willing, usually I like to compute the absolute difference and normalize by the score mean, e.g. |mirex_score - mir_eval_score|/(mirex_score + mir_eval_score)/2

Here you go:
results


> Can you track down someone to do a second one before merging?

Sure thing. Pinging @bmcfee ?

craffel:
> Here you go:

Thanks!  Given the 3%+ difference there, it would be great to get a cursory confirmation that it's because of optimal matching, even if that just means checking that the algorithm output for those algorithms produces many more notes (which would be likely to cause a greedy vs. non-greedy difference), for example.

rabitt:
> Thanks! Given the 3%+ difference there, it would be great to get a cursory confirmation that it's because of optimal matching, even if that just means checking that the algorithm output for those algorithms produces many more notes (which would be likely to cause a greedy vs. non-greedy difference), for example.

Will do!

mir_eval/io.py
```diff
@@ -477,3 +477,86 @@ def load_key(filename, delimiter=r'\s+'):
...
 
     return key_string
+
+
+def load_ragged_time_series(filename, dtype=float, delimiter=r'\s+',
+                            header=False):
+    r"""Utility function for loading in data from a delimited time series
+    annotation file with a variable number of columns.
+    Assumes that column 0 contains time stamps and columns 1 through n contain
+    values. n may be variable from time stamp to time stamp.
```
bmcfee: A small (3-line) example input here might be helpful.

mir_eval/io.py
```diff
@@ -477,3 +477,86 @@ def load_key(filename, delimiter=r'\s+'):
...
+    splitter = re.compile(delimiter)
+
+    # Keep track of whether we create our own file handle
+    own_fh = False
+    # If the filename input is a string, need to open it
+    if isinstance(filename, six.string_types):
+        # Remember that we need to close it later
+        own_fh = True
+        # Open the file for reading
+        input_file = open(filename, 'r')
```
bmcfee: this could leak file descriptors if there's an error anywhere below (before the call to close).

For this sort of thing, it's better to make a wrapper function that returns the file descriptor, and then use a context manager: `with my_open(buf_or_str, 'r') as fdisc: ....`

See [here](https://github.com/marl/jams/blob/master/jams/core.py#L58-L59) for an example of this in jams.

mir_eval/io.py
```diff
@@ -477,3 +477,86 @@ def load_key(filename, delimiter=r'\s+'):
...
+    if header:
+        start_row = 1
+    else:
+        start_row = 0
+    for row, line in enumerate(input_file, start_row):
+        # Split each line using the supplied delimiter
+        data = splitter.split(line.strip())
+        try:
+            converted_time = float(data[0])
+        except:
```
bmcfee: avoid catch-all exceptions; only catch the exception you expect here, let everything else float up-stream.

mir_eval/multipitch.py
```diff
@@ -0,0 +1,509 @@
...
+    times : np.ndarray
+        Array of time stamps
+    frequencies : list of np.ndarray
+        List of np.ndarrays of frequency values
+    target_times : np.ndarray
+        Array of target time stamps
+
+    Returns
+    -------
+    frequencies_resampled : list of lists
```
bmcfee: list of numpy arrays

mir_eval/multipitch.py
```diff
@@ -0,0 +1,509 @@
...
+        return [np.array([])]*len(target_times)
+
+    n_times = len(frequencies)
+
+    # scipy's interpolate doesn't handle ragged arrays. Instead, we interpolate
+    # the frequency index and then map back to the frequency values.
+    # This only works because we're using a nearest neighbor interpolator!
+    frequency_index = np.arange(0, n_times)
+
+    # assume_sorted=True for efficiency
```
bmcfee: This comment is unclear to me.  Are you allowed to assume_sorted=True because the array you're interpolating is constructed above (frequency_index) in sorted order already?

mir_eval/multipitch.py
```diff
@@ -0,0 +1,509 @@
...
+
+
+def frequencies_to_midi(frequencies, ref_frequency=440.0):
+    """Converts frequencies to continuous midi values.
+
+    Parameters
+    ----------
+    frequencies : list of np.ndarray
+        Original frequency values
+
```
bmcfee: missing parameter docstring

bmcfee:
> Sure thing. Pinging @bmcfee ?

Just took another pass over it.  Looks mostly good -- the things I think need "fixing" are the handling of existing file descriptors, and more specific exception handling.  There are also a few dangling comments from earlier reviews that don't seem to have been addressed yet.

mir_eval/io.py
craffel: I think @rabitt copied that from the other loaders, so if we change that we should change all loaders.  The logic is copied from numpy: https://github.com/numpy/numpy/blob/master/numpy/lib/npyio.py#L477
The difference being that we potentially raise an exception and they don't, I suppose.

bmcfee: > The difference being that we potentially raise an exception and they don't, I suppose.

Yup, that's exactly the point.  I guess the descriptor could close when it goes out of scope, but it still seems ugly to me.





> copied that from the other loaders

all the more reason to abstract this logic out and do it right. :)

bmcfee: ... note, if this problem exists elsewhere, I have no problem with leaving the present code as is and fixing it in a separate PR.

rabitt: Ok, I'll leave this for a future pr.

rabitt: Opened Issue #191

rabitt:
@craffel Sorry for the delay. Updates:

1) I thing I've incorporated all the comments from your and brian's code review

2) I found the reason for the 3% difference. The output from the "CB1" and "CB2" submissions had time stamps starting at 0.0, while the mirex ground truth (and all other submissions) have time stamps starting at 0.01. The mirex eval code assumes the stamps are correct while the mir_eval implementation resamples, resulting in the "correct" alignment with mir_eval and an off-by-one alignment between the reference and estimate in the mirex eval.

The "raw" scores are here:
results
(Note that the 3% differences are for CB1 and CB2)

Here are the scores again if I shift the estimates for CB1 and CB2 by one sample to match the mirex behavior:
results-shift

craffel:
Nice sleuthing!  Ok, I am happy to merge this unless there are any pending TODOs we need to take care of that I forgot about.

rabitt:
@craffel I _think_ I covered everything. Just made a second pass through the comments and didn't see anything outstanding.

craffel:
Ok, merged, thanks again!

Issue #180, opened 2016-02-22
Implement more transcription metrics


justinsalamon:
In addition to precision, recall, and f-measure, MIREX also computes the Average Overlap Ratio (defined in [this paper](http://ismir2009.ismir.net/proceedings/PS2-21.pdf)). Furthermore, it also computes all four metrics (p, r, f1, o) but with mapping the pitch values to chroma values (i.e. so that octave errors aren't penalized).

This means that MIREX computes 16 metrics in total:
(Precision, Recall, F-measure, AverageOR) x (with offsets, without offsets ) x (pitch, chroma) = 16

Currently we have:
(Precision, Recall, F-measure) x (with offsets, without offsets ) = 6

craffel:
:+1: Seems pretty straightforward to implement both, right?

justinsalamon:
>:+1: Seems pretty straightforward to implement both, right?

Yup, I just have no cycles left for it right now. Maybe @rabitt would fancy taking a stab at this? ;) Otherwise I'm happy to revisit this some time in the future.

rabitt:
> Maybe @rabitt would fancy taking a stab at this? ;) Otherwise I'm happy to revisit this some time in the future.
@justinsalamon Sorry, no cycles for this right now :(

justinsalamon:
Other than the missing MIREX metrics, there are 3 more sets of metrics I'm interested in - these are still p,r,f, but using *only* onset information, or only offset information, or only pitch information. This basically allows you to break down the results and gain further insight (e.g. some algorithms might be great at onset detection but terrible at pitch detection, etc.).

These metrics are not part of MIREX - would it still be in the scope of ``mir_eval``? From an implementation perspective onset and offset only are very straight forward to implement - they are already computed for calculating the overall note hit matrix, so it would be relatively straight forward derivate work. For pitch only I need to give it some thought since matching pitch values in a completely time-agnostic fashion doesn't make much sense.

But assuming I figure out something sensible for each of the three, would this make sense for PR?

craffel:
> These metrics are not part of MIREX - would it still be in the scope of mir_eval?

Potentially, yes, but from a holistic view I'd say they are much lower priority than the rest of the MIREX metrics.  If you want them though and are up for implementing them, feel free to make a PR.

>  For pitch only I need to give it some thought since matching pitch values in a completely time-agnostic fashion doesn't make much sense.

[...]

PR #189, opened 2016-04-12
Implement additional transcription metrics (#180)

justinsalamon:
This PR will include implementations of: 
- [x] P, R & F measures for onsets only 
- [x] P, R & F measures for offsets only
- [x] Overlap ratio metric (normal and `_no_offset` version)


justinsalamon:
@craffel the onset/offset-only metrics are pretty much done, but I have not added them to the evaluator since they are not officially part of MIREX. Any preference?

I'm still missing the overlap ratio (and maybe chroma alternatives) to complete the PR, but this metric is computed in MIREX so no doubts there.

craffel:
> the onset/offset-only metrics are pretty much done, but I have not added them to the evaluator since they are not officially part of MIREX. Any preference?

They should be put in the evaluator even if they're not in MIREX.

> I'm still missing the overlap ratio (and maybe chroma alternatives) to complete the PR, but this metric is computed in MIREX so no doubts there.

I won't have a chance to review this until mid-May.  Feel free to entice someone else to do a code review in the meanwhile.

justinsalamon:
> They should be put in the evaluator even if they're not in MIREX.

ok

> I won't have a chance to review this until mid-May. Feel free to entice someone else to do a code review in the meanwhile.

no worries, I'll ping the rest of team once this is ready for review

justinsalamon:
@craffel just one quick question: to compute the overlap ratio you need to compute note matching, and then the metric is derived from the matched notes similar to precision, recall and f1. Computationally the most efficient thing to do would be to add this metric to the existing `precision_recall_f1` function since it already computes the note matching. The alternative would be to implement the metric in a new function which also computes the note matching, but this means exactly the same note matching would be computed twice when the evaluator is called (and this where most of the computational load is).

So... add the metric to `precision_recall_f1` (and if so, rename function?), or implement in separate function?

craffel:
We shouldn't duplicate computation.  I would need to look at your code to decide if it makes sense to me to lump this all into one function because I'm not clear on what exactly the computation being duplicated is.

justinsalamon:
The computation being duplicated would be [this call to `match_notes`](https://github.com/craffel/mir_eval/blob/master/mir_eval/transcription.py#L346) which involves computing the 3 hit matrices (for onset/offset/pitch) and then computing the bi-partite graph matching.

craffel:
Ok.  I am fine with adding overlap ratio computation to the PRF function.  Why don't you code it up that way, and if there are any undesirable side-effects we can discuss them later.

justinsalamon:
ok sounds good. shall we rename it to `precision_recall_f1_overlap` then?

craffel:
Sounds good.

justinsalamon:
@craffel I've finished implementing the AOR metric, and so I've compared it to some MIREX results: https://gist.github.com/justinsalamon/b2a837a889a61a69b76a42a4ff38d889

In the first plot you can see the `mir_eval` results are pretty different from the MIREX results. The reason is that this metric is highly dependent on the note matching, and since we use a different matching algorithm, the matching is different and consequently the AOR too. In the second plot I've plotted the scores we'd get if `mir_eval` used the same greedy note matching as in MIREX, and you can see the scores now match just fine.

The reason AOR is so sensitive to note matching is because it's the average (over all matched notes) of the ratio between the overlapping segment of two matched notes and the maximum time duration spanned by the two notes. So for example, lets say greedy matching matches ref note A to est note B with start/end times [1, 3] [0.8, 2.8] respectively. Their overlap ratio will be (2.8 - 1)/(3 - 0.8) = 1.8/2.2 = 0.82. Now, let's say graph-based note matching instead matches ref note A to est note C with start/end times [1.2, 2]. Now the overlap ratio will be (2 - 1.2)/(3 - 1) = 0.8/2 = 0.4, a dramatic drop from 0.82.

All this to say, I'm very confident my implementation of the metric is correct, it's just going to be different compared to MIREX due to the different note matching algorithm used. To get a rough notion of how the metric changes I computed the differences for all algorithms evaluated on the Su dataset in 2015, displayed in the third plot of the notebook. You can see the change in score actually depends more on the specific track being evaluated and less on the algorithm (which is not surprising). For this dataset, changes range from an increase of 0.02 to a drop of almost -0.2. The last (box)plot in the notebook gives a good feel for the spread of differences. The median difference is about -0.02. Results are likely to vary for other datasets.

craffel:
Thanks for this analysis.  I'll take a closer look when I have more cycles.  It's interesting that the greedy vs. optimal matching here makes such a big difference.  Since this keeps popping up and effects a wide variety of metrics, I think it would be valuable to publish something about this (beyond what we discuss in the `mir_eval` paper -- i.e. discussing all of these other metrics and how it effects them), maybe as an LBD this year or something.

justinsalamon:
Cool, let me know when you've had a chance to give it a look. As far as implementation goes, I just need to generate new regression data and add back the regression test and then I think this PR is ready for a CR (I'll ping when it is). We're still be missing the chorma metrics to reach full coverage of the MIREX metrics, but I'm out of cycles. Might add them in a later PR.

If you think it's worth writing something up as an LBD I'm happy to help in where I can.

justinsalamon:
Alright, this is ready for a review. @bmcfee any chance you could give it look?

justinsalamon:
@ejhumphrey @rabitt @urinieto could someone please CR this?

mir_eval/transcription.py
```diff
@@ -138,6 +138,173 @@ def validate(ref_intervals, ref_pitches, est_intervals, est_pitches):
                          "value")
 
 
+def validate_intervals(ref_intervals, est_intervals):
+    """Checks that the input annotations to a metric look like time intervals,
+    and throws helpful errors if not.
+
```
bmcfee: "errors" -> "warnings"

Is this function redundant with [util.validate_intervals](https://github.com/craffel/mir_eval/blob/master/mir_eval/util.py#L636)?

mir_eval/transcription.py
```diff
@@ -138,6 +138,173 @@ def validate(ref_intervals, ref_pitches, est_intervals, est_pitches):
...
+        warnings.warn("Reference notes are empty.")
+    if est_intervals.size == 0:
+        warnings.warn("Estimate notes are empty.")
+
+
+def match_offsets(ref_intervals, est_intervals, offset_ratio=0.2,
+                  offset_min_tolerance=0.05, strict=False):
+    """Compute a maximum matching between reference and estimated notes,
+    only taking note offsets into account.
+
```
bmcfee: "notes" -> "offsets" ?  It's actually not clear what "match_offsets" means here.  Is there a better name we could give this function?

mir_eval/transcription.py
```diff
@@ -138,6 +138,173 @@ def validate(ref_intervals, ref_pitches, est_intervals, est_pitches):
...
+            G[est_i] = []
+        G[est_i].append(ref_i)
+
+    # Compute the maximum matching
+    matching = sorted(util._bipartite_match(G).items())
+
+    return matching
+
+
+def match_onsets(ref_intervals, est_intervals, onset_tolerance=0.05,
```
bmcfee: the documentation on this looks nearly identical to that of `match_offsets`.  Maybe provide some context (in each one) that describes how they relate to each-other, and why they both exist?

mir_eval/transcription.py
```diff
@@ -138,6 +138,173 @@ def validate(ref_intervals, ref_pitches, est_intervals, est_pitches):
...
+                 strict=False):
+    """Compute a maximum matching between reference and estimated notes,
+    only taking note onsets into account.
+
+    Given two note seqeunces represented by ``ref_intervals`` and
```
bmcfee: spellcheck "seqeunces"

mir_eval/transcription.py
```diff
@@ -264,21 +431,22 @@ def match_notes(ref_intervals, ref_pitches, est_intervals, est_pitches,
...
-    have an offset value within 20% (by default, adjustable via the
-    offset_ratio parameter) of the ref note's duration around the ref note's
-    offset, or within offset_min_tolerance (50 ms by default), whichever is
-    larger.
+    transcribed notes, and the Average Overlap Ratio for correctly transcribed
+    notes (see :func:`average_overlap_ratio`). "Correctness" is determined
+    based on note onset, pitch and (optionally) offset: an estimated note is
+    assumed correct if its onset is within +-50ms of a ref note and its pitch
+    (F0) is within +- quarter tone (50 cents) of the corresponding reference
+    note. If with_offset is False, note offsets are ignored in the comparison.
```
bmcfee: this function doesn't seem to have a "with_offset" parameter?

mir_eval/transcription.py
```diff
@@ -353,7 +523,182 @@ def precision_recall_f1(ref_intervals, ref_pitches, est_intervals, est_pitches,
...
+    """Compute the Average Overlap Ratio between a reference and estimate
+    note transcription. Given a reference and corresponding estimate note,
+    their overlap ratio (OR) is defined as the ratio between the duration of
+    the time segment in which the two notes overlap and the time segment
+    spanned by the two notes combined (earliest onset to latest offset): OR =
+    (min(ref_offset, est_offset) - max(ref_onset, est_onset)) /
+    (max(ref_offset, est_offset) - min(ref_onset, est_onset)). The Average
+    Overlap Ratio (AOR) is given the mean OR computed over all matching
+    reference and estimate notes. The metric goes from 0 (worst) to 1 (best).
+
```
bmcfee: This text would be easier to read if the equation was broken out somehow.

mir_eval/transcription.py
```diff
@@ -353,7 +523,182 @@ def precision_recall_f1(ref_intervals, ref_pitches, est_intervals, est_pitches,
...
+    avg_overlap_ratio: float
+        The computed Average Overlap Ratio score
+    """
+    ratios = []
+    for match in matching:
+        ref_int = ref_intervals[match[0]]
+        est_int = est_intervals[match[1]]
+        overlap_ratio = (
+            (min(ref_int[1], est_int[1]) - max(ref_int[0], est_int[0])) /
+            (max(ref_int[1], est_int[1]) - min(ref_int[0], est_int[0])))
```
bmcfee: is there anything here to prevent a divide-by-zero?

mir_eval/transcription.py
```diff
@@ -353,7 +523,182 @@ def precision_recall_f1(ref_intervals, ref_pitches, est_intervals, est_pitches,
...
+    if len(ref_intervals) == 0 or len(est_intervals) == 0:
+        return 0., 0., 0.
+
+    matching = match_onsets(ref_intervals, est_intervals,
+                            onset_tolerance=onset_tolerance,
+                            strict=strict)
+
+    onset_precision = float(len(matching))/len(est_intervals)
+    onset_recall = float(len(matching))/len(ref_intervals)
+    onset_f_measure = util.f_measure(onset_precision, onset_recall)
```
bmcfee: The other parts of mir_eval that do this sort of prf calculation also accept a `beta` parameter to balance the F-score calculation.  I think for consistency, this should be supported in the transcription metrics as well.

mir_eval/transcription.py
```diff
@@ -353,7 +523,182 @@ def precision_recall_f1(ref_intervals, ref_pitches, est_intervals, est_pitches,
...
+        return 0., 0., 0.
+
+    matching = match_offsets(ref_intervals, est_intervals,
+                             offset_ratio=offset_ratio,
+                             offset_min_tolerance=offset_min_tolerance,
+                             strict=strict)
+
+    offset_precision = float(len(matching))/len(est_intervals)
+    offset_recall = float(len(matching))/len(ref_intervals)
+    offset_f_measure = util.f_measure(offset_precision, offset_recall)
```
bmcfee: also here with `f_measure(..., beta=beta)`

bmcfee:
@justinsalamon I did a quick-skim CR -- generally good, but a few things need cleanup / consistency with the rest of mir_eval.

mir_eval/transcription.py
justinsalamon: I wasn't aware this function was already implemented :) That said, it doesn't check if `intervals.size > 0`, so an empty annotation with shape `(0, 2)` could still slip through it without issuing a warning. Perhaps the best option would be to add the warnings from my local implementation to `util.validate_intervals` and then use that one?

mir_eval/transcription.py
justinsalamon: The function matches note offsets, hence the name. Would renaming it to `match_note_offsets` be clearer? Note offset matching (as per MIREX) has some quirky logic, which is why this is implemented as a new function as opposed to using the existing `util.match_events`

mir_eval/transcription.py
justinsalamon: They both exist because note-onset matching and note-offset matching have different rules (according to MIREX). Worth documenting?

Also, both are different from `util.match_events` in that they support both `<=` and `<` comparisons, and round down the event times to avoid artificial mismatches (in particular when `<=` is used) due to floating point precision issues.

Perhaps worth considering incorporating these features into `util.match_events`?

mir_eval/transcription.py
justinsalamon: whoops

mir_eval/transcription.py
bmcfee: > That said, it doesn't check if intervals.size > 0, so an empty annotation with shape (0, 2) could still slip through it without issuing a warning.

I'm not sure that's a validation problem, since it corresponds to a well-formed empty set.  It depends on the metric's internal logic for how to handle empty annotations.

mir_eval/transcription.py
bmcfee: > Would renaming it to match_note_offsets be clearer?

yes, I think so.

mir_eval/transcription.py
justinsalamon: thanks for catching this, must have slipped through in the previous PR. will fix.

mir_eval/transcription.py
justinsalamon: Not sure how to do this, could you illustrate?

bmcfee:
```python
"""...
    the time segment in which the two notes overlap and the time segment
    spanned by the two notes combined (earliest onset to latest offset):

    OR = ((min(ref_offset, est_offset) - max(ref_onset, est_onset)) /
               (max(ref_offset, est_offset) - min(ref_onset, est_onset))).

    The Average Overlap Ratio (AOR) is given the mean OR computed over
    all matching reference and estimate notes.
...
```

mir_eval/transcription.py
justinsalamon: For the denominator to be 0 you need both intervals to have exactly the same start time *and* a duration of zero, which should get caught during validation (assuming your calling this function via `precision_recall_f1_overlap`). However:

1. The existing `validate` function in `transcription.py` doesn't catch this even though it should. I'll update `validate` to call `util.validate_intervals` internally, which will fix this.
2. If for some reason someone calls this function manually without validation then yes, it's not impossible for the denominator to evaluate to 0 (because the input data is invalid). Is it worth adding a check for this?

bmcfee: I think implementing 1. should be sufficient.

You might want to add a bit to the docstring that indicates that intervals should be validated before calling this function.

mir_eval/transcription.py
justinsalamon: It's not part of mirex, but I'm happy to add this.

mir_eval/transcription.py
justinsalamon: will do

mir_eval/transcription.py
justinsalamon: The data may be well-formed, but I think it's worth issuing a warning when the set is empty, because it's a corner case that affects many evaluation metrics. Do you think it's better to leave it out of the validation function and move the warnings into each individual metric function? We're already issuing warnings about empty annotations in a few other validation function across `mir_eval`.

mir_eval/transcription.py
justinsalamon: ok, changed.

mir_eval/transcription.py
justinsalamon: 👍

mir_eval/transcription.py
bmcfee: > They both exist because note-onset matching and note-offset matching have different rules (according to MIREX). Worth documenting?
> 

Yup.

> Also, both are different from util.match_events in that they support both <= and < comparisons, and round down the event times to avoid artificial mismatches (in particular when <= is used) due to floating point precision issues.
> 
> Perhaps worth considering incorporating these features into util.match_events?

I don't have too strong an opinion on this.  In general, I'd prefer to have fewer specialized functions, and instead just sanitize the input (eg by rounding to precision first).  If that's not possible/sufficient here, for whatever reason, then we should keep it as is and leave match_events untouched until there's cause to change it.

mir_eval/transcription.py
justinsalamon: ok cool.

However: validation will fail if the interval array doesn't have `shape[1] == 2`. This makes perfect sense for non-empty intervals lists, but do we want to be equally strict when the interval set is empty? Like you noted somewhere else, an empty interval set is still valid, but I can easily see someone (or an algorithm) returning an empty interval set that has a different shape (e.g. `()`): shouldn't that be acceptable?

mir_eval/transcription.py
justinsalamon: > Yup. 

ok

>  I don't have too strong an opinion on this. In general, I'd prefer to have fewer specialized functions, and instead just sanitize the input (eg by rounding to precision first). If that's not possible/sufficient here, for whatever reason, then we should keep it as is and leave match_events untouched until there's cause to change it.

The reason I'm rounding down is because I had a corner case in my unit tests where the distance between 2 events was exactly equal to the window, so when `<=` is used (as in `util.match_events`) the correct behavior would be to consider the events a match. However, without rounding down first the two events in my test were not being matched, and it turned out the reason was precision. So you could say this is an issue with `util.match_events`, since it uses `<=` but doesn't guarantee correct behavior for corner cases.

justinsalamon:
@bmcfee cheers, I've addressed everything I didn't think merits further discussion. I have 3 pending issues, mainly related to validation:

1. Ideally, I'd rather use `util.validate_intervals` over having a function with the same name inside `transcription.py`, however I do think it's good to issue a warning about empty interval sets, so my preference would be to add warnings about empty est/ref intervals to `util.validate_intervals` and then remove `validate_intervals` from `transcription.py` altogether.

2. Currently, `util.validate_intervals` will raise an error `if intervals.ndim != 2 or intervals.shape[1] != 2`. This makes sense for non-empty interval sets, but is perhaps over-strict in the case of empty interval sets that could (for instance) have `intervals.shape == (0,)`? For example an algorithm might initialize a list `intervals = []` to add stuff (e.g. notes) to, but then doesn't detect any, so just returns `np.asarray(intervals)` which will have `intervals.shape == (0,)`, and I'm not sure it makes sense to raise an error because the shape is not `(0,2)`?

Finally

3. Adding an optional `strict=False` parameter to `util.match_events` such that if `strict==True` then the function uses `<` instead of `<=`, and adding another optional parameter `decimals=None` such that if it's not `None` it's an integer indicating how many decimals to round event time differences down to for matching, would make the function more useful, for example it could probably replace [match_note_onsets](https://github.com/craffel/mir_eval/blob/transcriptionext/mir_eval/transcription.py#L241) which basically implements the aforementioned functionality.

craffel:
Thanks for the review @bmcfee.  I'm not supposed to be looking at this now but here's my 2c on these remaining issues:

> Ideally, I'd rather use util.validate_intervals over having a function with the same name inside transcription.py, however I do think it's good to issue a warning about empty interval sets, so my preference would be to add warnings about empty est/ref intervals to util.validate_intervals and then remove validate_intervals from transcription.py altogether.

`mir_eval`'s convention in this case, which seems sane to me, is that the submodule itself's `validate` function handles warning for empty annotations and then passes the rest off to the `util` validate function.  See https://github.com/craffel/mir_eval/blob/master/mir_eval/beat.py#L77 for example.  I don't think it's totally insane to think that some metric which takes in interval annotations would be ok with empty annotations (i.e. not warranting a warning).

> Currently, util.validate_intervals will raise an error if intervals.ndim != 2 or intervals.shape[1] != 2. This makes sense for non-empty interval sets, but is perhaps over-strict in the case of empty interval sets that could (for instance) have intervals.shape == (0,)? For example an algorithm might initialize a list intervals = [] to add stuff (e.g. notes) to, but then doesn't detect any, so just returns np.asarray(intervals) which will have intervals.shape == (0,), and I'm not sure it makes sense to raise an error because the shape is not (0,2)?

If the annotation specification is that the shape is `(n_intervals, 2)` and an algorithm outputs `np.array([])` or something, then that algorithm would have a bug.  I would argue it's not `mir_eval`'s job to handle that.

> Adding an optional strict=False parameter to util.match_events such that if strict==True then the function uses < instead of <=, and adding another optional parameter decimals=None such that if it's not None it's an integer indicating how many decimals to round event time differences down to for matching, would make the function more useful, for example it could probably replace match_note_onsets which basically implements the aforementioned functionality.

I'll leave that up to @bmcfee.

justinsalamon:
> mir_eval's convention in this case, which seems sane to me, is that the submodule itself's validate function handles warning for empty annotations and then passes the rest off to the util validate function. See https://github.com/craffel/mir_eval/blob/master/mir_eval/beat.py#L77 for example. I don't think it's totally insane to think that some metric which takes in interval annotations would be ok with empty annotations (i.e. not warranting a warning).

ok, in which case I'll have to keep my `validate_intervals`, but will add a call to `util.validate_intervals` inside this function.

> If the annotation specification is that the shape is (n_intervals, 2) and an algorithm outputs np.array([]) or something, then that algorithm would have a bug. I would argue it's not mir_eval's job to handle that.

ok

> I'm not supposed to be looking at this now

Too tempting? 🍰

craffel:
> ok, in which case I'll have to keep my validate_intervals, but will add a call to util.validate_intervals inside this function.

:+1:

> Too tempting? 🍰

💀

mir_eval/transcription.py
justinsalamon: Resolved via [Colin's comment](https://github.com/craffel/mir_eval/pull/189#issuecomment-214465870).

mir_eval/transcription.py
justinsalamon: Resolved via [Colin's comment](https://github.com/craffel/mir_eval/pull/189#issuecomment-214465870).

justinsalamon:
> 💀

👻  @bmcfee I think this means the last remaining discussion point is:

> Adding an optional strict=False parameter to util.match_events such that if strict==True then the function uses < instead of <=, and adding another optional parameter decimals=None such that if it's not None it's an integer indicating how many decimals to round event time differences down to for matching, would make the function more useful, for example it could probably replace match_note_onsets which basically implements the aforementioned functionality.

Appending relevant earlier comment:

> The reason I'm rounding down is because I had a corner case in my unit tests where the distance between 2 events was exactly equal to the window, so when <= is used (as in util.match_events) the correct behavior would be to consider the events a match. However, without rounding down first the two events in my test were not being matched, and it turned out the reason was precision. So you could say this is an issue with util.match_events, since it uses <= but doesn't guarantee correct behavior for corner cases.

bmcfee:
> I think this means the last remaining discussion point is:



> Adding an optional strict=False parameter to util.match_events such that if strict==True then the function uses < instead of <=, and adding another optional parameter decimals=None such that if it's not None it's an integer indicating how many decimals to round event time differences down to for matching, would make the function more useful, for example it could probably replace match_note_onsets which basically implements the aforementioned functionality.

I think it makes sense to adopt the `strict` parameter functionality in general, since there's no other way to do this.

The `decimals` case is less obvious to me. Can't this be accomplished by pulling the rounding logic up to the calling function, rather than building it into match_events?

I'd rather keep match_events (or similar functions) as simple as possible, and not bloat them unnecessarily.

justinsalamon:
>  think it makes sense to adopt the strict parameter functionality in general, since there's no other way to do this.

ok, depending on what our final decision for `match_events` is I can add it as part of this PR or it can be done separately.

> The decimals case is less obvious to me. Can't this be accomplished by pulling the rounding logic up to the calling function, rather than building it into match_events?
I'd rather keep match_events (or similar functions) as simple as possible, and not bloat them unnecessarily.

I agree that keeping it simple is preferable, but if you have a look at [`match_onsets`](https://github.com/craffel/mir_eval/blob/transcriptionext/mir_eval/transcription.py#L241) it's not particularly bloated. More importantly though, right now if you have 2 events at times 1.0 and 1.5, and you call `match_events` with a window of 0.5, then you'd expect the two events to match, whereas in practice it's possible for them not to match with the current implementation. So without rounding, `match_events` can actually behave incorrectly, which I think is decent motivation to include rounding. In `transcription.py` by default we round to 4 decimals, which is probably still more temporal resolution than any of the eval metrics requires, but it guarantees that when you use `<=` and the time difference between events is identical (or very close) to the tolerance window, the behavior will stay correct.

justinsalamon:
@bmcfee if you still prefer to keep `match_events` as is then I think this PR should be ready to merge.

bmcfee:
> @bmcfee if you still prefer to keep match_events as is then I think this PR should be ready to merge.

I'm fine with it as is.  If we decide to merge that logic into match_events, we can do so in another PR.

justinsalamon:
> I'm fine with it as is. If we decide to merge that logic into match_events, we can do so in another PR.

ok cool.

@craffel I think we're good to go on this one

craffel:
Can we compare to MIREX's implementation first, like last time?

justinsalamon:
> Can we compare to MIREX's implementation first, like last time?

You mean beyond the comparison I already shared here: https://gist.github.com/justinsalamon/b2a837a889a61a69b76a42a4ff38d889

The changes I made based on @bmcfee CR were mainly about naming conventions and validation, I haven't changed anything in the implementation of the metrics themselves since I performed the comparison in the notebook, so it still holds.

craffel:
Oh, I totally forgot about that, sorry, too much other stuff going on.  Thanks for reminding me.  I echo my original sentiment

> Since this keeps popping up and effects a wide variety of metrics, I think it would be valuable to publish something about this (beyond what we discuss in the mir_eval paper -- i.e. discussing all of these other metrics and how it effects them), maybe as an LBD this year or something.

Anyways, I am OK with merging in terms of the implementation/API being correct, although it's always helpful to get another pair of eyes on the docstrings etc. and I could do a code review this weekend/next week sometime.  Any big rush to merge?

justinsalamon:
> Anyways, I am OK with merging in terms of the implementation/API being correct, although it's always helpful to get another pair of eyes on the docstrings etc. and I could do a code review this weekend/next week sometime. Any big rush to merge?

No, we can definitely wait a week for a final CR if that works for you. No rush, I just have a few PR's pending in different places and want to make sure they all get merged (eventually). Thanks!

craffel:
Bug me if I have not done it by May 6th.

mir_eval/transcription.py
```diff
@@ -138,6 +135,187 @@ def validate(ref_intervals, ref_pitches, est_intervals, est_pitches):
...
+    ref_intervals : np.ndarray, shape=(n,2)
+        Array of reference notes time intervals (onset and offset times)
+    est_intervals : np.ndarray, shape=(m,2)
+        Array of estimated notes time intervals (onset and offset times)
+    """
+    # If reference or estimated notes are empty, warn
+    if ref_intervals.size == 0:
+        warnings.warn("Reference notes are empty.")
+    if est_intervals.size == 0:
+        warnings.warn("Estimate notes are empty.")
```
craffel: *Estimated

mir_eval/transcription.py
```diff
@@ -352,8 +540,195 @@ def precision_recall_f1(ref_intervals, ref_pitches, est_intervals, est_pitches,
...
+
+
+def average_overlap_ratio(ref_intervals, est_intervals, matching):
+    """Compute the Average Overlap Ratio between a reference and estimate
+    note transcription. Given a reference and corresponding estimate note,
+    their overlap ratio (OR) is defined as the ratio between the duration of
+    the time segment in which the two notes overlap and the time segment
+    spanned by the two notes combined (earliest onset to latest offset):
+
+    OR = ((min(ref_offset, est_offset) - max(ref_onset, est_onset)) /
```
craffel: I think this would probably look nicer wrapped into a code block.

craffel:
Looks good to me.  I made two minor docs nitpicks.  I'd also like it if you standardize how you are referring to reference and estimated notes and offsets in docs - some places it's "ref note", some places it's "reference", and you say "estimate note" instead of "estimated" in some places.  So, all nitpicks.

justinsalamon:
@craffel happy to make these changes, but before I do I just want to be sure - I think in most places I use "estimate", treating it as a noun, and a counterpart to "reference". It's true that in some places I use "estimated notes", but that's just for the sake of flow. For example in one of the two comments you made you note I use "Estimate note", but it's not a typo, it means "a note belonging to the estimate".

To make a long story short, I'm happy to make terminology consistent (as it should be), but I actually have a preference for "Reference notes" and "Estimate notes", so I'd change occurrences of "estimated notes" to "estimate notes". Would that work? How is this handled in other modules in `mir_eval`? If everyone uses "estimated" everywhere and I'm the odd one out I can follow suit.

craffel:
>  If everyone uses "estimated" everywhere and I'm the odd one out I can follow suit.

This is the case.  The convention is that you refer to the annotation as, e.g. "estimated beats", or generically as "the estimate", not as "estimate beats" (try a grep to see).

justinsalamon:
@craffel nitpicks have been nitpicked.

craffel:
Merged!  Thank you very much!

justinsalamon:
ehm, this still seems to be around :) shall I?

craffel:
Please do!

craffel:
Oh no!  You merged instead of closing...  it looks like history is not too messed up, but https://github.com/craffel/mir_eval/commit/d3b8df838ccf0adf1927aa734eef6c4cb75a9905 got committed twice.  I guess leaving that in there is better than rebasing and corrupting everyone's local copy.  In the future don't hit that merge button (or merge otherwise) after the PR has already been merged manually, you need to *close* it.

justinsalamon:
My bad - didn't realize I'd still have the option to merge if the PR had already been merged and thought it hadn't been. You live, you learn.

PR #195, opened 2016-05-26
implemented pitch contour sonification

### LINKED ISSUE #194: continuous pitch sonification
As per [this discussion](https://github.com/marl/jams/pull/91), we could pull in the code from @justinsalamon 's [melosynth](https://github.com/justinsalamon/melosynth) package.

bmcfee:
resolves #194

### DIFF
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -5,6 +5,8 @@ All functions return a raw signal at the specified sampling rate.
 
 import numpy as np
 from numpy.lib.stride_tricks import as_strided
+from scipy.interpolate import interp1d
+
 from . import util
 from . import chord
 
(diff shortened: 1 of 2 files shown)


bmcfee:
@craffel @justinsalamon ready for CR I think.

mir_eval/sonify.py
```diff
@@ -140,6 +142,64 @@ def _fast_synthesize(frequency):
...
+
+    # Estimate frequency at sample points
+    f_est = f_interp(np.arange(length))
+
+    # Sonify the waveform
+    x = function(2 * np.pi * np.cumsum(f_est) / fs)
+
+    # Clamp the extrapolated values
+    x[:int(times.min() * fs)] = 0
+    x[int(times.max() * fs):] = 0
```
craffel: I see why you did this from the test below, but this must create an actual click, right?  Is it really that bad to have a small ramp-down/ramp-up at the beginning and/or end (because that's all your avoiding here, right)?

craffel:
One minor comment otherwise LGTM.

mir_eval/sonify.py
bmcfee: Actually it's worse than that -- the interpolator seems to be constant-padding instead of using `fill_value` for extrapolation.  Not sure why that is, but it seems incorrect based on my reading of the scipy docs, and it happens in both 0.16.1 and 0.17.  This workaround just forces the output to be zero at extrapolation points.

At any rate, the discontinuity here doesn't bother me too much.  I can add a ramp if you think it's necessary, but I'd rather keep it simple.

craffel: > Actually it's worse than that -- the interpolator seems to be constant-padding instead of using fill_value for extrapolation.

Hmm, seems ok to me:

```Python
In [1]: import numpy as np

In [2]: import scipy.interpolate

In [3]: x = np.linspace(.2, 1.1, 10)

In [4]: y = np.random.random_integers(2, 5, 10)

In [5]: f = scipy.interpolate.interp1d(x, y, kind='slinear', fill_value=10., bounds_error=False)

In [6]: f(.15)
Out[6]: array(10.0)
```

Is that not what you'd expect?

> At any rate, the discontinuity here doesn't bother me too much.

Well, not explicitly zeroing out doesn't sound *worse*, right?  Unless there is some nasty behavior I am not seeing, it seems like interpolate will already be zeroing out things.

bmcfee: > Is that not what you'd expect?

... that is what i'd expect, and i get the same results when i run your code locally.  NFC what's going on inside the synthesizer then; i'll investigate.



> Well, not explicitly zeroing out doesn't sound worse, right? Unless there is some nasty behavior I am not seeing, it seems like interpolate will already be zeroing out things.

It should be, and it isn't, unless my tests are really broken.  I'll report back shortly.

bmcfee: Aha!  The problem is that testing for zeros at the end of the signal is not correct, since the interpolated frequencies are accumulated through cumsum.  This is easily fixable by testing that they're constant though.

bmcfee: Fixed.  Thanks for catching that.

tests/test_sonify.py
```diff
@@ -53,3 +54,37 @@ def test_chords():
...
+    for start, end in zip(idx[::2], idx[1::2]):
+        freqs[start:end] *= -1
+
+    # Test with inferring duration
+    x = mir_eval.sonify.pitch_contour(times, freqs, fs)
+    assert len(x) == fs * 5
+
+    # Test with an explicit duration
+    # This forces the interpolator to go off the end of the sampling grid,
+    # which should put zeros at the end of the signal
```
craffel: Are these comments still correct?

bmcfee: If by "signal" we mean "instatnaneous frequencies" then yes; I can clarify though.

craffel:
Merging, thanks for being so quick on this!

bmcfee:
I think something got screwed up in the git history here?  This merge is no longer showing up in the master timeline.  The network graph shows the PR still dangling.  O_o

craffel:
Weird, I have no idea why that happened.  Just re-merged in #198.  Looks ok?  Github was having some hiccups earlier this week, maybe it has to do with that?  I dunno.

bmcfee:
Looks good now, thanks!  It's possible that when I ff-pushed to my own fork on the display branch, something got out of sync and/or fat-fingered to upstream (ie craffel/mir_eval).  At any rate, all is well now.

Issue #28, opened 2014-03-26
display submodule


craffel:
For making useful qualitative plots for eval-by-eye

bmcfee:
FWIW, this issue is starting to pop up in jams https://github.com/marl/jams/issues/93 wrt pretty-printing annotations.

It might be good to sync on this, just to figure out what reasonable requirements would be.

craffel:
I think this issue was mainly for plots, not pretty-printing annotations, but the two are roughly related.

bmcfee:
"plots" meaning ... what?

craffel:
You know, [plots](https://en.wikipedia.org/wiki/Plot_(graphics)).

bmcfee:
Er, plots of what? Scores?  Annotations?

I'd argue that score-plotting is best handled by collecting results in a dataframe and using seaborn.  (Or other custom code; point is we'd likely be reinventing the wheel to do that in mir_eval).

Annotation-plotting is a different story though.

craffel:
Annotatoins.

bmcfee:
> Annotations.

Ok; that's why I ask. :)  Note that by "pretty-printing" what I actually mean is notebook-embed and svg rendering.  (There are several good reasons to want direct svg output for this, rather than going through matplotlib, but i digress.)

It'd be good if we could minimize overlap in functionality/implementation on this.

carlthome:
Dead feature?

bmcfee:
Not dead, but (IMO) it's more natural to fit this into [JAMS](https://github.com/marl/jams/pull/110) than mir_eval.

carlthome:
Yeah, thanks! :smiley:  https://github.com/marl/jams/issues/113

bmcfee:
Reversing my prior comments on this after discussing in the jams threads: mir_eval should have some basic viz/plotting, and JAMS can decorate these with metadata/subplots/etc.

Let's make a wish-list of how this should work?

- events (onsets, beats, etc): vertical lines or dotted points?  i can see a good case for either.
- disjoint intervals: i like how [msaf](https://github.com/urinieto/msaf) does this, but it might get dense for short annotations (ie chord labels)
- segment hierarchy: msaf does a nice job of this as well.
- overlapping intervals (eg instrument activations): i think the only sane thing here is to group intervals by label, and give each unique label a different vertical position.  I suspect an alpha channel setting here might be necessary for when similarly labeled intervals also overlap.
- source separation: ???  should we bother with this?
[...]

PR #196, opened 2016-05-28
display submodule

bmcfee:
This implements #28 

TODO:

- [x] segments
- [x] intervals
- [x] segment hierarchy
- [x] notes
- [x] contours
- [x] source separation
- [x] tests


craffel:
Re: `segments` and `labeled_intervals`, I'm not crazy about this name because you can also use labels in `segments`.  I am fine with `segments` and `intervals`.  Segments meaning "I have a segmentation of time", intervals meaning "I have a bunch of intervals which might overlap.  But also, it's a name, I'm happy with whatever.

> I think it actually makes sense to drop property cycling entirely from the labeled-interval display.

Ok.

> for the segment display, I think it makes more sense to rely on the axes internal fill style cycler, rather than roll our own.

Sounds good.  Maybe we can include a small example of how to set an axis' color cycler?  I bet many people have not done this.  Or, whatever.

> ehh.. y is ambiguous here, in that it can mean target or y-axis (in this case, kinda both).

What's `target`?  `label_set` is fine.

bmcfee:
> But also, it's a name, I'm happy with whatever.

Yeah, I think if we provide examples showing the output of each, it will be obvious enough.


> Maybe we can include a small example of how to set an axis' color cycler? I bet many people have not done this. Or, whatever.

I figure most people will just do this via `plt.style.use('seaborn-muted')` or somesuch.  There's an example of more direct manipulation [here](http://matplotlib.org/examples/color/color_cycle_demo.html), though that works on line styles and not fill styles.  It shouldn't be hard to adapt, but I don't think it's a high priority.

bmcfee:
Here's an example of hierarchical segment plots, comparing two annotations from salami:
```python
plt.figure(figsize=(12, 8))
ax = plt.subplot(2,1,1)
mir_eval.display.labeled_intervals(int2, lab2, ax=ax, alpha=0.8, label='small')
mir_eval.display.labeled_intervals(int1, lab1, ax=ax, alpha=0.8, label='functions')
plt.legend(loc='best')
plt.title('Annotator 1')


ax = plt.subplot(2,1,2)
mir_eval.display.labeled_intervals(int4, lab4, ax=ax, alpha=0.8, label='small')
mir_eval.display.labeled_intervals(int3, lab3, ax=ax, alpha=0.8, label='functions')
plt.title('Annotator 2')
plt.legend(loc='best')
```

image

I think I generally prefer this to the analogous MSAF-style plot:

```python
plt.figure(figsize=(12, 8))
ax = plt.subplot(2,1,1)
mir_eval.display.segments(int1, lab1, base=0.5, ax=ax, text=True)
mir_eval.display.segments(int2, lab2, height=0.5, ax=ax, text=True)
plt.title('Annotator 1')

ax = plt.subplot(2,1,2)
mir_eval.display.segments(int3, lab3, base=0.5, ax=ax, text=True)
mir_eval.display.segments(int4, lab4, height=0.5, ax=ax, text=True)
plt.title('Annotator 2')
```
image

The first one is a little easier to read (IMO) for a few reasons:

1. no reuse of colors between segment levels
2. color can be used to differentiate levels
3. ability to set legend entries (`small` and `functions`)for different levels
4. spatial patterns between levels are easier to discern than the color patterns in the bottom plot.

Anyone have thoughts for how the segment plot can be improved?  @urinieto ?

craffel:
The only thing I like a bit more in the "analogous MSAF-style" is that it's visually easier for me to see which subsegments in the lower plot correspond to larger segments in the upper plot.  The vertical separation in the first one makes this a little less obvious to me.  This could be solved maybe by drawing dashed lines in the lower plot corresponding to boundaries in the upper plot, or something.

Playing devil's advocate, now:

> no reuse of colors between segment levels

This is the worse part, IMO, of the  "analogous MSAF-style", but could pretty easily be solved if you dive back into automatic color cycling.

> color can be used to differentiate levels

Well, so does vertical position - I don't think it's hard to differentiate levels in the  "analogous MSAF-style".

> ability to set legend entries (small and functions)for different levels

These could be y labels.

> spatial patterns between levels are easier to discern than the color patterns in the bottom plot.

I think if there was no color re-use it would be about as easy either way, assuming there were not too many colors.

bmcfee:
> The only thing I like a bit more in the "analogous MSAF-style" is that it's visually easier for me to see which subsegments in the lower plot correspond to larger segments in the upper plot. The vertical separation in the first one makes this a little less obvious to me. This could be solved maybe by drawing dashed lines in the lower plot corresponding to boundaries in the upper plot, or something.

Really?  I find it pretty easy, but maybe that's because I know what to expect.  We could add ticks and grid lines, but it would probably get cluttered quickly.



> but could pretty easily be solved if you dive back into automatic color cycling.

Not really; you'd eventually have to rely on texture and alpha channels.  Even then, the biggest complaint that I've gotten from people is that colors don't mean anything across hierarchy levels, but they really want them to.  I think it's best to avoid that whole hornet's nest if we can.





> Well, so does vertical position - I don't think it's hard to differentiate levels in the "analogous MSAF-style".

For two levels, no, but for deeper structures it can get messy.



> These could be y labels.

Do you mean tick marks on the y axis?  It's possible, sure.



> I think if there was no color re-use it would be about as easy either way, assuming there were not too many colors.

I disagree, but that's anecdotal.  Opinions from others would be useful here.

craffel:
> Really? I find it pretty easy, but maybe that's because I know what to expect. We could add ticks and grid lines, but it would probably get cluttered quickly.

Yeah, I have to keep changing focus from top to bottom plot.  It's not that bad.

> I think it's best to avoid that whole hornet's nest if we can.

Fair enough.

> For two levels, no, but for deeper structures it can get messy.

Wouldn't lower just mean finer structure regardless of the number of levels?  Or is that not how hierarchical structure is defined?

> Do you mean tick marks on the y axis?

Yep.

bmcfee:
Pinging @justinsalamon re: pitch and notes.

I see in your [prototype](https://github.com/marl/jams/pull/110/files) that you used rectangles for notes, and the y axis was set to a midi (numeric) scale.  I think this can be plugged directly into `labeled_intervals` as is.

I think we can do a little better if we define a [FuncFormatter](http://matplotlib.org/api/ticker_api.html#matplotlib.ticker.FuncFormatter) that can reinterpret midi numbers as pitches (with octave and cent deviation) or frequencies (Hz).  I'd prefer this to raw numbers or explicit ticker strings (like librosa currently does) because it will behave nicely with interactive plotting/zooming/etc.

For contours, we've talked about plotting on a cent axis.  I think it makes the most sense to just map to continuous/fractional midi numbers via `12 * (log2(f) - log2(440)) + 69`.  That way you can overlay pitch contours on quantized midi notes.

It would also be nice to plot contours directly in frequency space for easy overlaying on spectrograms.  Ultimately, this should be easy to do with a flag.  The only nontrivial part of pitch contour plotting then will be voicing detection and grouping of samples into fragments.

Am I missing anything here?

justinsalamon:
> I see in your prototype that you used rectangles for notes, and the y axis was set to a midi (numeric) scale. I think this can be plugged directly into labeled_intervals as is.

hmmm, maybe? MIDI notes can but don't have to be quantized. If I understand correctly labeled intervals builds the y-axis labels from the data labels? This might not a good option for non-quantized pitch data?

> For contours, we've talked about plotting on a cent axis. I think it makes the most sense to just map to continuous/fractional midi numbers via 12 * (log2(f) - log2(440)) + 69. That way you can overlay pitch contours on quantized midi notes.

👍 

> The only nontrivial part of pitch contour plotting then will be voicing detection and grouping of samples into fragments.

For melody this shouldn't be a problem since there's only (at most) one active pitch per timestamp. Voicing can be indicated by a change of style I suppose (color? line type? alpha?). Grouping samples into contours becomes a tough cooky for multi-f0, especially since the mir_eval format (correct me @rabitt if I'm wrong) doesn't specify which samples belong to which contour. Maybe apply basic TF-clustering as a best guess?

bmcfee:
> hmmm, maybe? MIDI notes can but don't have to be quantized. If I understand correctly labeled intervals builds the y-axis labels from the data labels? This might not a good option for non-quantized pitch data?

Yeah, fair point.  We might have to refactor a bit to separate the patch logic from the vertical positioning.  The bigger issue though is that we don't have a great way to specify the height of rectangles: should they span an entire semitone?  That doesn't seem quite right.  Maybe it would be better to just dump these into a line plot instead?



> Voicing can be indicated by a change of style I suppose (color? line type? alpha?).

I was going to just not plot the unvoiced segments, but i suppose you might want those anyway.  As a dirty hack you could simply to `display.pitch(t, f) ; display.pitch(t, -f, alpha=0.5)` for the same effect.



> Grouping samples into contours becomes a tough cooky for multi-f0,

ugh... yeah, i forgot about multi-f0 plotting here.  This seems like a problem that needs to be resolved at the multipitch API first, and display later.  I really dislike the idea of clustering tf values in the display module to sort that out.

bmcfee:
Quick proof of concept for pitch contour plotting:

```python
ref_times, ref_freqs = mir_eval.io.load_labeled_events('/home/bmcfee/git/mir_eval/tests/data/melody/ref00.txt')
est_times, est_freqs = mir_eval.io.load_labeled_events('/home/bmcfee/git/mir_eval/tests/data/melody/est00.txt')

plt.figure(figsize=(12, 4))
mir_eval.display.pitch(ref_times, ref_freqs, unvoiced=True, midi=True,
                       linewidth=4, alpha=0.5, label='Reference')
mir_eval.display.pitch(est_times, est_freqs, unvoiced=True, midi=True, 
                       linewidth=1, alpha=1, label='Estimate')
plt.legend(loc='best')
plt.tight_layout()
```

image


It allows you to toggle plotting of unvoiced regions (which are depicted with the same style, but half the alpha as the voiced regions).  A region is considered "voiced" if all of its frequencies are nonzero (and freq_to_voicing says so); this avoids underflow in the log scaling (midi units) conversion.

Styles are inherited from the axis line cycler, and the same style is applied to all fragments within a call to `display.pitch`.

justinsalamon:
> Yeah, fair point. We might have to refactor a bit to separate the patch logic from the vertical positioning. The bigger issue though is that we don't have a great way to specify the height of rectangles: should they span an entire semitone? That doesn't seem quite right. Maybe it would be better to just dump these into a line plot instead?

This is a matter of taste. My personal preference would be for each note to span a semitone (in height), since it matches the convention used in MIDI editors in DAWs so I find it intuitive. While it's true that a note doesn't inherently "span" a pitch range, I think the most common use case for this visualization will be for notes quantized on a semitone grid, for which this styling would be most helpful I think.

> I was going to just not plot the unvoiced segments, but i suppose you might want those anyway. As a dirty hack you could simply to display.pitch(t, f) ; display.pitch(t, -f, alpha=0.5) for the same effect.

Yeah you definitely want to be able to display unvoiced pitch estimates. Not sure what your hack exapmle achieves?

> ugh... yeah, i forgot about multi-f0 plotting here. This seems like a problem that needs to be resolved at the multipitch API first, and display later. I really dislike the idea of clustering tf values in the display module to sort that out.

Even if you sort it for the JAMS API, once you pass it to mir_eval it will be converted to the MIREX format which doesn't specify grouping. So as long as mir_eval supports the MIREX format, it's not clear to me how this can be solved on an API level?

justinsalamon:
> Quick proof of concept for pitch contour plotting:

This looks almost right to me:
 * I can't tell the difference between voiced and unvoiced segments. Maybe using a dashed line for unvoiced would be easier to see?
 * The colors used for ref and est make it very hard to distinguish them, especially when they overlap. I trust this can be changed though.

> A region is considered "voiced" if all of its frequencies are nonzero (and freq_to_voicing says so)

What do you mean by this? Why do you need to infer if a region is voiced beyond what's already indicated by the raw data?

bmcfee:
> This looks almost right to me:

style choices are left as an exercise to the reader.  also note: this is using mpl 2.0.0b1, which may have some kinks to work out.



> What do you mean by this? Why do you need to infer if a region is voiced beyond what's already indicated by the raw data?

A region can be "unvoiced" with either negative or 0 frequency.  If it's negative, it gets mapped to positive, and voicing is set to False.  If it's 0, well, it stays 0.  When you go to take a log for midi conversion, explosions happen.  In general, I think you just don't want to show 0-frequency values at all, since they distort the axes and convey no information.  Am I wrong here?

justinsalamon:
> In general, I think you just don't want to show 0-frequency values at all, since they distort the axes and convey no information. Am I wrong here?

Yes, that's correct. A 0 Hz estimate means "I detected no pitch here" and there's no reason to display it. A negative value means "I detected pitch here but I don't think it's part of the melody (but it might be if I'm wrong)", which is what we want to display as "unvoiced".

bmcfee:
And with dynamic-friendly midi ticker conversion :notes: 
mpl_midi

(check the lower-right for hover coordinates and the ticks on the left as we zoom in)

justinsalamon:
Nice!

bmcfee:
And with the piano roll:

```python

ref_times, ref_pitches = mir_eval.io.load_valued_intervals('/home/bmcfee/git/mir_eval/tests/data/transcription/ref03.txt')
est_times, est_pitches = mir_eval.io.load_valued_intervals('/home/bmcfee/git/mir_eval/tests/data/transcription/est03.txt')


plt.figure(figsize=(8, 4))

mir_eval.display.piano_roll(ref_times, ref_pitches, label='Reference', alpha=0.5)
mir_eval.display.piano_roll(est_times, est_pitches, label='Estimate', alpha=0.5, facecolor='r')

plt.ylim(ref_midis.min(), ref_midis.max())
plt.grid('on')
plt.legend(loc='best')
mir_eval.display.ticker_notes()
plt.tight_layout()

```

after some zooming in interactive mode:

image

justinsalamon:
Nice! I'm confused about the y-axis gridlines: shouldn't notes be centered on the pitch values? It seems like right now they sit "on top" of the actual pitch value?

Another side issue is the density of the y-grid. For a piano roll I'd like to always have a y-tick (and label) for every semitone, independently of the zoom level of the plot. Otherwise it gets a little tricky to determine the pitch of some notes without counting with your fingers on the screen :)

bmcfee:
> Nice! I'm confused about the y-axis gridlines: shouldn't notes be centered on the pitch values? It seems like right now they sit "on top" of the actual pitch value?

They're baseline-aligned to the bars; the ticks/grid lines go below the relevant note, but the text labels are where they should be.  I don't want to have them run through the rects because that would put the patch positions at non-integer values, and generally make life a pain.  (If you don't believe me, try manipulating a bar plot some time.)


> Another side issue is the density of the y-grid. For a piano roll I'd like to always have a y-tick (and label) for every semitone, independently of the zoom level of the plot. Otherwise it gets a little tricky to determine the pitch of some notes without counting with your fingers on the screen :)


Yup, just added minor ticks at all semitones for the piano roll plotter.  Labels are still dynamically generated; if you want them in specific locations, that's your business as a discerning user with impeccable taste.

justinsalamon:
> They're baseline-aligned to the bars; the ticks/grid lines go below the relevant note, but the text labels are where they should be. I don't want to have them run through the rects because that would put the patch positions at non-integer values, and generally make life a pain. (If you don't believe me, try manipulating a bar plot some time.)

It might be easier to implement, but it's hella-confusing. The note *above* the D#6 label is a D#6? eek!

FWIW, this is how the big guys do it:
Image

bmcfee:
> It might be easier to implement, but it's hella-confusing.

hella you talkin about?  That's the same thing: the label "C1" is parallel to the notes at pitch C1.

justinsalamon:
Not sure I follow. In the logic screenshot above, the label C1 is vertically aligned with notes representing a C1. In the piano roll figure, the labels are vertically aligned with the gridlines, which aren't aligned with any of the notes.

bmcfee:
Okay folks.  I feel like the bulk of this, not counting source separation is basically done from a prototyping perspective.  [Here's a notebook that you can play with](https://gist.github.com/d02bad6e74e7a1f1300c526f14273854), feedback is welcome.

At this point, it's worth thinking about how we should test things.  Matplotlib-style image comparison regression tests?  They're kind of a pain to set up, but I'm not sure what else we can do.

craffel:
Thanks @bmcfee.  I am currently moving everything out of my house into a big box so won't get a chance to look at this for a day or two.  One thing that popped up and I don't think was resolved:

> The only nontrivial part of pitch contour plotting then will be voicing detection and grouping of samples into fragments.

(and the ensuing discussion).  I think the only sane thing to do is to not group at all, and just plot multi-f0 as non-connected dots.  That's how they're evaluated, anyways (right?).  I definitely do not think TF clustering is something we want to reckon with, unless there is a very simple approach which does the right thing the vast majority of the time.

justinsalamon:
> (and the ensuing discussion). I think the only sane thing to do is to not group at all, and just plot multi-f0 as non-connected dots. That's how they're evaluated, anyways (right?). I definitely do not think TF clustering is something we want to reckon with, unless there is a very simple approach which does the right thing the vast majority of the time.

Just adding my 2c here: plotting as dots is definitely an option, it's just a little more clumsy because it means plotting thousands of dots which (on some systems) can slow things down, and can make the images quite "heavy" in that if included e.g. in a PDF (as vector graphics) the reader might choke.

I think you can probably get pretty decent results with the most simple TF clustering: only group points if they're adjacent on the time grid and within X cents from each other (e.g. 50). We could try this on a few examples to see what it produces...?

bmcfee:
> Thanks @bmcfee. I am currently moving everything out of my house into a big box so won't get a chance to look at this for a day or two.

Sure.  I'm going to try to finish up the source separation plot today, hack out some tests, and then be done with it.  It might be helpful if you set up a milestone for the next release though; I think we've got enough stuff here (viz + sonification improvements, new metrics) to warrant one ahead of ismir.



> I think the only sane thing to do is to not group at all, and just plot multi-f0 as non-connected dots.

Sounds like just a straight scatter plot with some styling on unvoiced bits.  This should be easy enough to add.



> unless there is a very simple approach which does the right thing the vast majority of the time.

The pitch contour plot currently does something like this.  The algorithm simply finds changepoints between voiced/unvoiced, and treats everything in between as a single contour.  I think this is correct enough for monophonic pitch, but i'm convinced that multi-f0 needs to be a separate (but similar) function.

bmcfee:
> I think you can probably get pretty decent results with the most simple TF clustering: only group points if they're adjacent on the time grid and within X cents from each other (e.g. 50).

I'd really rather not.  Dealing with scatter plot density is a much more general problem that is better solved in the plotting engine.  The more complicated we make display within mir_eval, the more time we'll have to spend maintaining it down the road.  I'd rather just keep things as simple as they can possibly be.

craffel:
> It might be helpful if you set up a milestone for the next release though; I think we've got enough stuff here (viz + sonification improvements, new metrics) to warrant one ahead of ismir.

Ok, made a milestone and added this to it.  There are some misc issues to be fixed up before the next release, too.

> plotting (tens/hundreds of) thousands of dots which (on some systems) can slow things down, and can make the images quite "heavy" in that if included e.g. in a PDF (as vector graphics) the reader might choke.

That's true, although we could always support downsampling in this case (the same sort of thing/justification behind librosa's waveplot).

There's the other issue that as you zoom in things could start to look really silly, like only showing a few dots.  But that's more of an edge case.

> I think you can probably get pretty decent results with the most simple TF clustering: only group points if they're adjacent on the time grid and within X cents from each other (e.g. 50). We could try this on a few examples to see what it produces...?

and 

> I'd really rather not. Dealing with scatter plot density is a much more general problem that is better solved in the plotting engine. The more complicated we make display within mir_eval, the more time we'll have to spend maintaining it down the road. I'd rather just keep things as simple as they can possibly be.

If @justinsalamon works very reliably in most cases, for me that's simple enough to include in eval.  If you're feeling inspired @justinsalamon maybe you can implement it as a fork of @bmcfee's progress and we can decide from there?

bmcfee:
On to source separation...

```python
y0, sr = librosa.load('/home/bmcfee/git/mir_eval/tests/data/separation/ref00/0.wav', sr=None)
y1, sr = librosa.load('/home/bmcfee/git/mir_eval/tests/data/separation/ref00/1.wav', sr=sr)
y2, sr = librosa.load('/home/bmcfee/git/mir_eval/tests/data/separation/ref00/2.wav', sr=sr)

plt.figure(figsize=(8, 4))

plt.subplot(2,1,1)
separation([y0, y1, y2], fs=sr)
plt.xticks([])

plt.subplot(2,1,2)
separation([y0 + y1 + y2], fs=sr, labels=['Mixture'])

plt.tight_layout()
```
produces
image

Some things I don't like about it right now:

* scipy's spectrogram function is a little weird
* log-power scaling with pcolormesh is kind of a pain.  probably I should use mpl normalize objects
* legend drawing is also a pain.  I have to create proxy artists and attach them directly to a legend, which means you can't re-draw the legend at a later point.  Still working out how to fix this...

bmcfee:
One more question for @craffel on this: how do you want to handle dependencies?  The display module now adds two dependencies to mir_eval: matplotlib >= 1.5 and scipy >= 0.16 (for signal.spectrogram).

For now, I've set it up as an extension, so you can `pip install mir_eval[display]` to get all the dependencies needed for the display module to work, and the submodule is not imported by default at the top level.  Do you want to keep it this way, or unify all the dependencies and have everything load as a single block?

craffel:
I am ok with bumping the required scipy version to 0.16.  I hesitate a little bit to add matplotlib as a strict dependency because that would mean people would need to install it in order to run the evaluators, for example, but I also hesitate to complicate installation/usage for the library.  Generally, I think of anyone who is using Python for numeric computing to install numpy, scipy, and matplotlib by default, but I don't really know.  And I also don't really know how many of our users are actually just using the evaluators alone vs. writing python code which calls them.  What do you think?

craffel:
> On to source separation...

Probably this is an artifact of the example, but I can't spot Source 0 in the separate-sources plot.  Is it all quiet high-frequency stuff?  I think I see some blue there.

bmcfee:
> Probably this is an artifact of the example, but I can't spot Source 0 in the separate-sources plot. Is it all quiet high-frequency stuff? I think I see some blue there.

Not high-f, but masked by other sources.  ref05 is a better example:

image




> enerally, I think of anyone who is using Python for numeric computing to install numpy, scipy, and matplotlib by default, but I don't really know.

I'd think so too, but i also get tons of flack on the mpl dependency in librosa, since it's a rather hefty import.  (Maybe 2.0 will fix this, duno.)

craffel:
> Not high-f, but masked by other sources. ref05 is a better example:

That's pretty sick!

> I'd think so too, but i also get tons of flack on the mpl dependency in librosa, since it's a rather hefty import. (Maybe 2.0 will fix this, duno.)

Ok, let's make it optional for now then.

bmcfee:
Killing the `units` todo here; I thought mpl had nice builtins for this kind of thing (Hz -> KHz, secs -> minutes, etc), but it turns out it was just an overwrought example in their gallery :disappointed:.

craffel:
In separation, are you using transparency?  Or can sources actually cover up one another according to the order in which they are presented to the function/plotted? I can see an argument for not using transparency (so that overlapping colors don't create new colors), but without it it seems like the plot could chance according to the order in which the sources are plotted.

bmcfee:
> In separation, are you using transparency?

Yes.  For each source, we generate a colormap that interpolates from (1,1,1,0) (white, transparent) to `(r,g,b,1)` where `(r,g,b)` is a color sourced from the axis property cycler.  The source is then normalized by the cumulative spectrogram peak (ie aggregating all sources and taking the max over time/frequency), and they are overlayed on top of each other.

We could alternately make it cycle up to `(r,g,b,0.75)` (or something) so that overlapping energy between sources adds up to a new color, but it might be confusing on a white background.  Maybe a better option is to do it starting from a black background?

bmcfee:
follow-up: i tried it and it looks like clown tears.  Not sure if there's a better way to do this that supports arbitrarily many sources.  Maybe averaging the spectrogram images in RGB space after applying the per-source color maps?  That seems a little bonkers to me.

craffel:
> follow-up: i tried it and it looks like clown tears. 

Ok, no worries, thanks for trying it out.

bmcfee:
Taking a crack at multipitch right now ... gaaaaaaaah where did this format come from?

justinsalamon:
> Taking a crack at multipitch right now ... gaaaaaaaah where did this format come from?

Image

bmcfee:
multipitch:

```python
ref_times, ref_pitches = mir_eval.io.load_ragged_time_series('/home/bmcfee/git/mir_eval/tests/data/multipitch/ref00.txt')
est_times, est_pitches = mir_eval.io.load_ragged_time_series('/home/bmcfee/git/mir_eval/tests/data/multipitch/est00.txt')

plt.figure(figsize=(8, 4))
mir_eval.display.multipitch(ref_times, ref_pitches, midi=True, alpha=1, label='Reference')
mir_eval.display.multipitch(est_times, est_pitches, midi=True, s=8, alpha=0.75, color='k',
                            label='Estimate')
mir_eval.display.ticker_notes()

plt.legend(loc='best')
plt.grid('on', which='major')
plt.grid('on', which='minor', alpha=0.1)
plt.tight_layout()
```

image

bmcfee:
TODO:
- [ ] rewrite `segments` to use `axvspan` instead of raw rectangle patches, if that makes sense
- [ ] be more sane about automatically setting axis limits

bmcfee:
@craffel this pep8 fascism is really cramping my style :sob: 

Is there any way we can disable it for certain files?  Specifically, I copied over the image comparison test code from librosa (itself copied and ever-so-slightly modified from matplotlib's internal module), and pep8 is throwing a fit.  Since this bit of code is borrowed from an existing library, I'd prefer to not modify it any more than is strictly necessary to make it work, since it will make diffing against future versions a total pain.

craffel:
> Is there any way we can disable it for certain files? 

Sorry, I'd rather not. If there are style choices you disagree with we can address them separately but I definitely do not want different style rules being followed in the same project.  If you're averse you can send the file to me and I will change it for you.

bmcfee:
> If there are style choices you disagree with we can address them separately but I definitely do not want different style rules being followed in the same project.

I'm fine with that for test code that we write for mir_eval, but I think you should seriously reconsider this for the specific case I'm describing here with the image comparison module.

For all intents and purposes, this should be treated as an import from an external module (matplotlib) which would not normally be covered by your style conventions anyway.  The only reason it is included in the repository and locally imported is because the code in question is not public API exposed by matplotlib.  A couple of minor modifications were necessary to make the module work outside of mpl source tree, but otherwise, I think it should be left as close as possible to the original.  This will make it easier to diff and update (if necessary) when new versions of mpl come out.

bmcfee:
One more problem that actually matters:

We can't have display tests conform to pep8 because you need to set `matplotlib.use('Agg')` before importing `matplotlib.pyplot`.  pep8 requires all package imports to be grouped at the top, so you can't do the following in `test_display.py`:

```python
import matplotlib
matplotlib.use('Agg')
...
```
The following does conform to pep8, but doesn't actually work:
```python
import matplotlib
matplotlib.use('Agg')
...
```
TLDR: we can have pep8, or display tests, but not both.

bmcfee:
Aside from pep8 pedantry, tests are all done and pass locally.

craffel:
> I think it should be left as close as possible to the original.

Well, for sure I think it should be *functionally equivalent*, and a style guide never requires anything beyond that (or at least, it shouldn't).  I personally find it much more dangerous/error-prone/untidy to have a file in your source tree which is completely ignored, but of course a lot of this is a matter of taste.  At any rate, here is a pep8 compliant version, with diff: https://gist.github.com/craffel/9065b6472c9776c5ab37c0af9fae40f6  If you want to be sure they are functionally equivalent, feel free to just run it through pep8ify instead.

> TLDR: we can have pep8, or display tests, but not both.

Or, we can ignore this error in our pep8 call in travis.yml, which I am totally open to discussing.  Of course, your `# nopep8` directive is another option which works fine too.

bmcfee:
> and a style guide never requires anything beyond that (or at least, it shouldn't).

True enough, but [quoth Guido](https://mail.python.org/pipermail/python-dev/2010-November/105681.html):
```
You are never going to have all code compliant. Use
the style guide when it helps, ignore it when it's in the way.
[...]
```
and I think, in this case, it is definitely in the way.  All that we gain by pep8'ing mpl_ic is a warm fuzzy feeling that tests pass, but we lose the ability to easily track changes from upstream.  This, to me, isn't worth it, but it's your baby so I'll shut up about this.

> Of course, your # nopep8 directive is another option which works fine too.

Yeah, I think that's the best option for the import problem.

bmcfee:
Okay, I put in the pep8ed mpl_ic and cleaned up some logic around limit-setting of axes.  I think this one's ready for CR.

craffel:
>  think, in this case, it is definitely in the way. All that we gain by pep8'ing mpl_ic is a warm fuzzy feeling that tests pass, but we lose the ability to easily track changes from upstream.

Yeah, I think the issue here is needing to copy-paste stuff (slightly modified) from another project (rather than being able to import it directly), not PEP-8 compliance.  Ideally, this would be shared functionality in some package we could just utilize.

> I think this one's ready for CR.

I am driving cross-country, will try to take some car time to look over it in the next few days.  Thanks again.

bmcfee:
Roger that. Happy driving!

Maybe if I get bored, I'll fork the image comparator out to a separate package

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+        getter, setter = ax.get_xlim, ax.set_xlim
+    elif which == 'y':
+        getter, setter = ax.get_ylim, ax.set_ylim
+    else:
+        raise ValueError('invalid axis: {}'.format(which))
+
+    old_lims = getter()
+    new_lims = list(limits)
+
+    if np.isfinite(old_lims[0]):
```
craffel: When are values returned by `get_*lim` non-finite?  Maybe add a comment.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    labels : list, shape=(n,)
+        reference segment labels, in the format returned by
+        :func:`mir_eval.io.load_labeled_intervals`.
+
+    base : number
+        The vertical position of the base of the rectangles.
+        By default, this will be the bottom of the plot.
+
+    height : number
+        The height of the base of the rectangles.
```
craffel: Isn't this just the height of the rectangles?

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+        By default, this will be the top of the plot (minus `base`).
+
+    text : bool
+        If true, each segment's label is displayed in its
+        upper-left corner
+
+    text_kw : dict
+        If `text==True`, the properties of the text
+        object can be specified here.
+        See `matplotlib.pyplot.text` for valid parameters
```
craffel: Probably more helpful to point people to `matplotlib.pyplot.Text`, because `Text`'s docstring includes all of the possible kwargs and as far as I can tell `text`'s doesn't.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    seg_map = dict()
+
+    for lab in labels:
+        if lab in seg_map:
+            continue
+
+        style = next(cycler)
+        seg_map[lab] = seg_def_style.copy()
+        seg_map[lab].update(style)
+        seg_map[lab]['facecolor'] = seg_map[lab].pop('color')
```
craffel: What's going on here?  The cycler returns a dict with a key `'color'`, but it should be `'facecolor'`?  Maybe a comment to this effect would be helpful.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+        seg_map[lab].update(kwargs)
+        seg_map[lab]['label'] = lab
```
craffel: So I don't actually know what the `'label'` keyword to a `Rectangle` patch does.  Is it for the agg backend or something?

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+
+    for ival, lab in zip(intervals, labels):
+        rect_kwargs = seg_map[lab]
+        rect = Rectangle((ival[0], base), ival[1] - ival[0], height,
+                         **rect_kwargs)
+        ax.add_patch(rect)
+        rect_kwargs.pop('label', None)
```
craffel: Are you modifying this label's keyword arguments so the label only shows up for the first patch, or something?  You don't use `rect_kwargs` after this point.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+            ann.set_clip_path(rect)
+
+    ax.set_yticks([])
+
+    __expand_limits(ax, [intervals.min(), intervals.max()], which='x')
+
+    return ax
+
+
+def labeled_intervals(intervals, labels, label_set=None,
```
craffel: I think we talked about this, but this naming is confusing to me since you can plot labeled intervals with the `segments` function.  I mean, the docstring clears it up pretty quickly, but yeah.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+                         **rect_kwargs)
+        ax.add_patch(rect)
+        rect_kwargs.pop('label', None)
+
+        if text:
+            ann = ax.annotate(lab,
+                              xy=(ival[0], height), xycoords='data',
+                              xytext=(8, -10), textcoords='offset points',
+                              **text_kw)
+            ann.set_clip_path(rect)
```
craffel: Nice matplotlib wizardry here.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+        :func:`mir_eval.io.load_labeled_intervals`.
+
+    labels : list, shape=(n,)
+        reference segment labels, in the format returned by
+        :func:`mir_eval.io.load_labeled_intervals`.
+
+    label_set : list
+        An (ordered) list of labels to determine the plotting order.
+        If not provided, the labels will be inferred from existing
+        `yticklabels`.
```
craffel: Might be useful to be more explicit, e.g. "the labels will be inferred from `ax.get_yticklabels`".

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+        If no `yticklabels` exist, then the sorted set of unique values
+        in `labels` is taken as the label set.
+
+    base : np.ndarray, shape=(n,), optional
+        Vertical positions of each label
```
craffel: Specify behavior when `base` (and below, `height`) is `None`.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+
+    height : np.ndarray, shape=(n,), optional
```
craffel: May be useful to allow the user to provide a single value for `height` to be used for all of the rows, rather than providing an array of heights.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    '''
+
+    if ax is None:
+        # Create a new axis
+        ax = plt.gca()
+
+    if label_set is None:
+        # If we have non-empty pre-existing tick labels, use them
+        label_set = [_.get_text() for _ in ax.get_yticklabels()]
+        if not any(label_set):
```
craffel: Add a comment here, e.g. "When all of the labels are empty strings, treat as an empty label set"

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+        seg_map[lab]['facecolor'] = seg_map[lab].pop('color')
+        seg_map[lab].update(kwargs)
+
+        seg_y[lab] = (y0, yi)
+
+    seen = set()
+    for ival, lab in zip(intervals, labels):
+        if lab not in seg_map:
+            continue
+        over_lab = seg_map[lab].get('label', lab)
```
craffel: When will `'label' not in seg_map[lab]` be `True` (in other words, why are you using `get` here instead of just indexing)?  It seems like it will always be set on line 226, `seg_map[lab] = dict(label=lab)`.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    kwargs:
+        Additional keyword arguments to `labeled_intervals`.
+
+    Returns
+    -------
+    ax
+        A handle to the matplotlib axes
+    '''
+
+    # This will break if a segment label exists in multiple levels
```
craffel: Wait, isn't this bad?  But also, why will it break?

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    if levels is None:
+        levels = list(range(len(intervals_hier)))
+
+    # Get the axis handle up front
+    kwargs.setdefault('ax', plt.gca())
```
craffel: Shouldn't you allow the user to pass an `ax` like in the other display functions?

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    ax
+        Handle to the plotting axes
+    '''
+
+    if ax is None:
+        ax = plt.gca()
+
+    # First, segment into contiguously voiced contours
+    frequencies, voicings = freq_to_voicing(np.asarray(frequencies,
+                                                       dtype=np.float))
```
craffel: Shouldn't `frequencies` already be an `np.ndarray` as per the docstring?  If you're just casting, it is (in my subjective opinion) clearer to use `astype`.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+
+    # Here are all the change-points
+    v_changes = 1 + np.flatnonzero(voicings[1:] != voicings[:-1])
+    v_changes = np.unique(np.concatenate([[0], v_changes, [len(voicings)]]))
+
+    v_slices, u_slices = [], []
```
craffel: Comments here would be good.  What are `v_slices` and `u_slices`?

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    ----------
+    times : np.ndarray, shape=(n,)
+        Sample times of frequencies
+
+    frequencies : list of np.ndarray
+        frequencies (in Hz) of the pitch measurements.
+        Voicing is indicated by sign (positive for voiced,
+        non-positive for non-voiced).
+
+        `times` and `pitches` should be in the format produced by
```
craffel: `times` and `frequencies`

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+            freqs = hz_to_midi(freqs)
+
+        n_voiced = sum(voicings)
+        voiced_times.extend([t] * n_voiced)
+        voiced_freqs.extend(freqs[voicings])
+        unvoiced_times.extend([t] * (len(freqs) - n_voiced))
+        unvoiced_freqs.extend(freqs[~voicings])
+
+    # Plot the voiced frequencies
+    ax.scatter(voiced_times, voiced_freqs, **style_voiced)
```
craffel: This reminded me that I challenged @justinsalamon to do clustering here: https://github.com/craffel/mir_eval/pull/196#issuecomment-223031002 "If you're feeling inspired @justinsalamon maybe you can implement [TF lustering] as a fork of @bmcfee's progress and we can decide from there?"  Just a final ping on that

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    voiced_freqs = []
+
+    unvoiced_times = []
+    unvoiced_freqs = []
+
+    for t, freqs in zip(times, frequencies):
+        if not len(freqs):
+            continue
+
+        freqs, voicings = freq_to_voicing(np.asarray(freqs, dtype=np.float))
```
craffel: Same `asarray` comment as above.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    levels : list of string
+        Each element `levels[i]` is a label for the `i`th segmentation.
+        This is typically used to denote the levels in a segment hierarchy.
+
+    kwargs:
+        Additional keyword arguments to `labeled_intervals`.
+
+    Returns
+    -------
+    ax
```
craffel: Above, you were more informative/descriptive about `ax`.  I liked that doc more.  Same for the functions below here.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    labels_hier : list of list-like
+        A list of segmentation labels.  Each element should
+        be a list of labels for the corresponding element in
+        `intervals_hier`.
+
+    levels : list of string
+        Each element `levels[i]` is a label for the `i`th segmentation.
+        This is typically used to denote the levels in a segment hierarchy.
+
+    kwargs:
```
craffel: Above, you did not use a colon after kwargs in the docstrings.  Below, looks like you are.  I think Sphinx doesn't like to see a colon here, but I dunno.  Maybe just be consistent and do whichever is right.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+        A list of waveform buffers corresponding to each source
+
+    fs : number > 0
+        The sampling rate
+
+    labels : list of strings
+        An optional list of descriptors corresponding to each source
+
+    ax : matplotlib.pyplot.axes
+        An axis handle on which to draw the intervals.
```
craffel: draw the spectrograms, not draw the intervals.

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+
+    for i, src in enumerate(sources):
+        freqs, times, spec = spectrogram(src, fs=fs, **kwargs)
+        specs.append(spec)
+        if cumspec is None:
+            cumspec = spec.copy()
+        else:
+            cumspec += spec
+
+    ref_max = cumspec.max()
```
craffel: If this is all you're using cumspec for, you should probably make a comment to this effect when fussing with it above.  (You could also just keep track of the max value seen in all spectrograms, which might simplify your code somewhat, but do whichever you prefer).

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+
+        color = next(ax._get_lines.prop_cycler)['color']
+        cmap = LinearSegmentedColormap.from_list(labels[i],
+                                                 [(1.0, 1.0, 1.0, 0.0),
+                                                  color])
+
+        ax.pcolormesh(times, freqs, spec,
+                      cmap=cmap,
+                      norm=LogNorm(vmin=ref_min, vmax=ref_max),
+                      shading='gouraud',
```
craffel: I don't know what's happening here but I like it (please comment).

mir_eval/display.py
```diff
@@ -0,0 +1,672 @@
...
+    legend_labels.extend(labels)
+
+    ax.legend(handles=handles, labels=legend_labels)
+
+    __expand_limits(ax, [times.min(), times.max()], which='x')
+    __expand_limits(ax, [freqs.min(), freqs.max()], which='y')
+    return ax
+
+
+def __ticker_midi_note(x, pos):
```
craffel: So, the main functionality here seems useful in its own right, so I would be happy to have it be non-private and factored out into a simply callable function which gave a pretty string for a given MIDI note number/float.  It'd be a bit misplaced, but useful, I think.

tests/test_display.py
```diff
@@ -0,0 +1,345 @@
...
+import mir_eval.display
+from mir_eval.io import load_labeled_intervals
+from mir_eval.io import load_valued_intervals
+from mir_eval.io import load_labeled_events
+from mir_eval.io import load_ragged_time_series
+from mir_eval.io import load_wav
+
+
+# Test fixtures will use seaborn-muted style
+GLOBAL_STYLE = 'seaborn-muted'
```
craffel: If a new version of seaborn uses slightly different colors, won't our tests all fail?  Is there any way to safeguard against this?

craffel:
Ok, just did a big review.  Thank you again for this.  This is going to be some truly useful (not to mention fun) stuff to play with. 

Apart from my specific comments:

- As per usual it would be great to get a lot more comments (you are doing a lot of elegant matplotlib wizardry that can be unclear even for a serious matplotlib user/abuser like myself); I made some specific suggestions for this in some places.  
- Did I already suggest adding a small note to the intervals functions mentioning how the colors are set and providing an example of how to change this?  I feel like people will care about that.  It would be cool to have examples for each function; if you do this, you could add that there.
- I think all the tests are regression tests, right?  It might be useful (albeit stupid) to also have some kind of unit-test-ish behavior, like checking axes limits and maybe the number of patches if such a thing is even possible.  I will by no means force you to do this but if you really like writing tests...
- It would be great to have other people mess around with it a bit and try to break it before merging.  Maybe you can bug specific annotation-savvy people to give it a try? E.g. @justinsalamon for pitch countours, @urinieto for (hierarchical) intervals, @ejhumphrey or @mattmcvicar  for overlapping intervals (chords), @rabitt for multipitch, @faroit or @dpwe or @carlthome for source separation, etc?  Let's get a display party going.

mir_eval/display.py
bmcfee: This should only happen for freshly constructed axes with no data attached.  I'll add a comment.

mir_eval/display.py
bmcfee: Good catch, yes.

mir_eval/display.py
bmcfee: Good catch.

mir_eval/display.py
bmcfee: No, it's for legend entries.  If you want to use a legend instead of the embedded text boxes, this is how you set that attribute.  We pop it after drawing the first one so you don't get redundant legend entries.

mir_eval/display.py
bmcfee: rect_kwargs gets re-used the next time that label is encountered in the loop.

mir_eval/display.py
bmcfee: Sure, but I couldn't think of a better solution here.  I think it's okay as is.

mir_eval/display.py
bmcfee: If by "nice" you mean "ugly" .. yeah.  This really ought to be done in relative coordinates with axspan, but that turned into a mess with multi-layer display.  I figure this is good enough.

mir_eval/display.py
bmcfee: This will be true after the label has been encountered for the first time.  We pop it to suppress redundant legend entries.

mir_eval/display.py
bmcfee: Yes, that's right.  I'll add comments.

mir_eval/display.py
bmcfee: Yeaaaaaaaaaah we shouldn't do that.

mir_eval/display.py
bmcfee: No, you need the max of the sum of the spectra, which you can't know without seeing all the data.

I'll comment it though.

mir_eval/display.py
bmcfee: :laughing:

mir_eval/display.py
bmcfee: Can we do that in a separate PR?

tests/test_display.py
bmcfee: No, we're using matplotlib's seaborn style sheet, not seaborn itself.  It will be stable across versions.

However, when matplotlib 2 rolls out, we will have to redo the regression test fixtures.

bmcfee:
> As per usual it would be great to get a lot more comments (you are doing a lot of elegant matplotlib wizardry that can be unclear even for a serious matplotlib user/abuser like myself); I made some specific suggestions for this in some places.

Yup, will do.  Thanks for the detailed cr!



> Did I already suggest adding a small note to the intervals functions mentioning how the colors are set and providing an example of how to change this? I feel like people will care about that. It would be cool to have examples for each function; if you do this, you could add that there.

This might be better accomplished with a demo notebook and/or example docstrings.  I doubt you want to go down the road of inline-docstring example rendering though, since it breaks rtd.  Your call.



> I think all the tests are regression tests, right? It might be useful (albeit stupid) to also have some kind of unit-test-ish behavior, like checking axes limits and maybe the number of patches if such a thing is even possible. I will by no means force you to do this but if you really like writing tests...

Yes, they're all regression tests -- they kind of have to be since the output is visual.  I don't want to go too overboard with that since they easily bloat the repo.  If there's something specific that seems dangerous though, we can totally add tests for it.



> t would be great to have other people mess around with it a bit and try to break it before merging. Maybe you can bug specific annotation-savvy people to give it a try?

I agree that would be great, but I'm also not holding my breath for any of those people to chime in on this... :cry:

mir_eval/display.py
bmcfee: It's bad, but documented.

It breaks because of the logic that infers label positioning from the axes.  In general, we want the same label to display at the same vertical position, eg when plotting overlapping chord or note segmentations.  But when you have a multi-level segmentation, you want each level to be a distinct set of labels.  BUT you also don't want to have to append the segment level index onto the labels.

Anyway, since there are not yet standard conventions for handling multi-level segmentations, I think we're in a decent position to require that labels are disjoint across levels.

mir_eval/display.py
bmcfee: I suppose that's reasonable.

mir_eval/display.py
bmcfee: Yes, but this implementation isn't wrong, and it's a bit more forgiving anyway, so I see no reason to change it.

mir_eval/display.py
craffel: Doesn't look like you changed this.  Do you disagree here?

mir_eval/display.py
craffel: Sure.  I assign you to making an issue to remind of this.

craffel:
> Yup, will do. Thanks for the detailed cr!

Anything you didn't address or is it mostly covered?

> This might be better accomplished with a demo notebook and/or example docstrings.

Demo notebook makes the most sense to me.

>  If there's something specific that seems dangerous though, we can totally add tests for it.

I can't think of anything.  It worries me a little that we are assuming that the output is correct without testing certain things explicitly (like maybe axes limits, labels, etc), but I can live with it.

> I agree that would be great, but I'm also not holding my breath for any of those people to chime in on this... 

Yes, you're probably right.  Well, I think at least one other person should play with it before merging, which I can do... eventually.  Alternatively, I will be in NYC for ICML next week, so maybe we could wrangle a few people and do a quick hack session, or something.

mir_eval/display.py
bmcfee: Huh, I guess this one got buried under "outdated diff" while I was working through it.

Yes, I can add this.

bmcfee:
> Anything you didn't address or is it mostly covered?

I think it's all covered.



> I can't think of anything. It worries me a little that we are assuming that the output is correct without testing certain things explicitly (like maybe axes limits, labels, etc), but I can live with it.

Agreed.. but most of this stuff is pretty transparent calls to mpl functions, so I feel comfortable with it.



> Well, I think at least one other person should play with it before merging, which I can do... eventually.

Sure.  It's not hard to fire up a notebook that loads test fixture data and kicks the tires on it.  I think it's all pretty simple stuff (user-wise) even if the guts are complicated.

bmcfee:
I guess the one remaining thing that might be useful to add is an event ticker (eg for times or beats).

We punted this idea before, since it's relatively trivial to do by hand with `plt.vlines`, but I think it makes sense to include for completeness (and it would also simplify the jams autodisplay module to have it).  If done in a similar style to `segments`, it would be a nice shortcut to drawing annotations over spectrograms or other feature representations.

I can jam this out in a few minutes (probably faster than folks will get around to testing this module anyway).

bmcfee:
Side note: I also fixed a quirk where axis ticks could disappear if you plot segments over an existing plot.  More generally, the display module now correctly detects when new axes are created, and only then will it mangle things like ticks.

craffel:
Just looked at it, looks nice.  One thing that I noticed which may apply in multiple places - if the user supplies `text_kw`, the defaults disappear:
```Python
    if text_kw is None:
        text_kw = dict(va='top',
                       clip_on=True,
                       bbox=dict(boxstyle='round', facecolor='white'))
```
So, for example, if the user passed `{'color': 'red'}` or something for `text_kw`, it would no longer be top-aligned, wouldn't have a box, etc.  I would think the user would want the defaults to remain the same unless they explicitly changed them, wouldn't you think so?  Which could be accomplished something like this:
```Python
    if text_kw is None:
        text_kw = dict()
    text_kw['va'] = text_kw.get('va', 'top')
    # etc...
```
What do you think?

mir_eval/display.py
```diff
@@ -0,0 +1,814 @@
...
+    ax
+        A handle to the matplotlib axes
+    '''
+
+    # This will break if a segment label exists in multiple levels
+    if levels is None:
+        levels = list(range(len(intervals_hier)))
+
+    # Get the axes handle
+    ax, new_axes = __get_axes(ax=ax)
```
craffel: Nitpick: `new_axes` is unused here, I think.

bmcfee:
Nits have been picked.

mir_eval/display.py
```diff
@@ -0,0 +1,816 @@
...
+    to [NOTE][OCTAVE]+[cents].
+    '''
+
+    NOTES = ['C', 'C#', 'D', 'D#', 'E', 'F', 'F#', 'G', 'G#', 'A', 'A#', 'B']
+
+    idx = int(x % 12)
+    cents = float(np.mod(x, 1.0))
+    if cents >= 0.5:
+        cents = cents - 1.0
+        idx += 1
```
wuaalb: Shouldn't centering cents increment note before wrapping octave? i.e.
```python
    cents = float(np.mod(x, 1.0))
    if cents >= 0.5:
        cents = cents - 1.0
        x += 1
    idx = int(x % 12)
```

bmcfee: Good catch.  This can be fixed by rounding instead of int-dividing.

wuaalb: Maybe we're talking about different edge cases or I misunderstood..

I was referring to the case where `idx` becomes out of bounds for `NOTES`,
e.g. `x = 47.8` resulted in `idx = 12`, `cents = -0.2`, `octave = 3`
with the code I posted `idx = 0`, `cents = -0.2`, `octave = 4`
with your last commit `idx = 12`, `cents = -0.2`, `octave = 4` (but octaves now wrap in the middle?)

bmcfee: > Maybe we're talking about different edge cases or I misunderstood..

Ah, quite right.  I think I misunderstood.  Serves me right for checking issues before breakfast...

bmcfee: > Related to this, maybe it would be nice if the lowest octave was configurable (I think -2 or -1 is more common than 0).

Really?  Isn't that well below the range of human pitch perception?

wuaalb: The calculation of the octave number still has to be reverted to how it was before.

I don't think there's a single definition of how octave numbers relate to absolute pitch in Hz, however MIDI note number 69 is usually C3 or C4, rather than the C5 in your current implementation.

bmcfee: Er, shouldn't 69 be A4, not C*?

wuaalb: Yes ofc, sorry :)

bmcfee: Doublederp on me for not regression testing against librosa's converter.

Fixing and updating tests now.

bmcfee:
TODO: one more tiny bug in separation limit setting

mir_eval/display.py
```diff
@@ -0,0 +1,819 @@
...
+        by transparency.
+
+        Otherwise, unvoiced pitch contours are omitted from the display.
+
+    ax : matplotlib.pyplot.axes
+        An axis handle on which to draw the intervals.
+        If none is provided, a new set of axes is created.
+
+    kwargs :
+        Additional keyword arguments to `labeled_intervals`.
```
wuaalb: Copy/paste mistake (intervals -> pitch contours, labeled_intervals -> plt.plot)
```python
'''...
    ax : matplotlib.pyplot.axes
        An axis handle on which to draw the pitch contours.
        If none is provided, a new set of axes is created.

    kwargs :
        Additional keyword arguments to `plt.plot`
'''
```

mir_eval/display.py
```diff
@@ -0,0 +1,819 @@
...
+
+    unvoiced : bool
+        If `True`, unvoiced pitches are plotted and indicated
+        by transparency.
+
+        Otherwise, unvoiced pitches are omitted from the display.
+
+    ax : matplotlib.pyplot.axes
+        An axis handle on which to draw the intervals.
+        If none is provided, a new set of axes is created.
```
wuaalb: Copy/paste mistake (intervals -> pitch contours)
```python
'''...
    ax : matplotlib.pyplot.axes
        An axis handle on which to draw the pitch contours.
        If none is provided, a new set of axes is created.
    ...
'''
```

mir_eval/display.py
```diff
@@ -0,0 +1,819 @@
...
+
+        # If we've already seen this label, remove it
+        # This way, it only appears once in the legend
+        if over_lab in seen:
+            seg_map[lab].pop('label', None)
+
+        seen.add(over_lab)
+        ax.add_patch(Rectangle((ival[0], seg_y[lab][0]),
+                               ival[1] - ival[0],
+                               seg_y[lab][1], **seg_map[lab]))
```
wuaalb: For me adding all rectangles at once as a `PatchCollection` helped performance (I tend to make these kind of visualizations somewhat interactive; zooming, panning, scrubbing, ..)

```python
rects = []
for ival, lab in zip(intervals, labels):
    ...
    rect = Rectangle((ival[0], seg_y[lab][0]),
                     ival[1] - ival[0],
                     seg_y[lab][1], **seg_map[lab])
    rects.append(r)
collection = PatchCollection(rects, match_original=True)
ax.add_collection(collection)
```

wuaalb:
Maybe it would be nice to add a ticker for x-axis in beats and bars?

Something like
```python
# from __future__ import division
def ticker_bars_and_beats(tempo, time_sig=(4, 4), ax=None):
    '''Set the x-axis of the given axes to bars (major ticks, 
    labeled) and beats (minor ticks, unlabeled)

    Parameters
    ----------
    tempo : float
        The tempo in BPM. Variable tempos are not supported.
    time_sig : tuple of ints
        The time signature as a (numerator, denominator) tuple.
        Variable time signatures are not supported.
        By default, uses 4/4.
    ax : matplotlib.pyplot.axes
        The axes handle to apply the ticker.
        By default, uses the current axes handle.
    '''
    ax, _ = __get_axes(ax=ax)

    def qn_to_sec(qn):
        return qn/tempo*60.0

    def sec_to_qn(sec):
        return sec/60.0*tempo

    def sec_to_bars(x, pos):
        bars = int(round(sec_to_qn(x)/4/(time_sig[0]/time_sig[1]))) + 1
        return '{:d}'.format(bars)

    bar_locator = MultipleLocator(qn_to_sec(4*time_sig[0]/time_sig[1]))
    bar_fmt = FuncFormatter(sec_to_bars)
    beat_locator = MultipleLocator(qn_to_sec(4/time_sig[1]))

    ax.xaxis.set_major_locator(bar_locator)
    ax.xaxis.set_major_formatter(bar_fmt)
    ax.xaxis.set_minor_locator(beat_locator)
```

mir_eval/display.py
bmcfee: That makes sense.  I tried this out, but it broke the label/legend construction.  I think the only way to do this and preserve functionality is to have a separate patch collection for each label.

bmcfee: I rewrote `labeled_intervals` using `BrokenBarHCollection`, since `PatchCollection` does not propagate labels correctly.

It's considerably simpler and faster now.

craffel:
Ok, spent some time playing with this.  It's great.  I had a lot of fun visualizing some `pretty_midi` stuff.  Go MIDI!  Here are some thoughts/questions/observations, some of which I think can be safely ignored, but were things that came up as I was messing around.

- I would expect the display functions to automatically make x and y axis labels (like in `librosa`).
- `mir_eval.display.events(events, base=1., height=2.)` displays nothing when using a new figure, because it does not change the boundaries from `[0, 1], [0, 1]` (same with other functions where you can change the plot range)
- `mir_eval.display.events(events, colors='k')` doesn't change the line color to black (but `plt.plot(events, colors='k')` does).
- `mir_eval.display.events(np.linspace(0, 10, 5)` sets the xlim to [0, 10], which makes the first and last lines invisible (a Matplotlib policy, I think).  I feel like it would be slick if it adjusted the range to always be slightly larger than the range of the events, for this reason.
- So what exactly does the `levels` kwarg do in `hierarchy`?  It doesn't seem to change the figure.
- Sort of similar to above - `mir_eval.display.labeled_intervals(intervals, labels, tick=False)` on a new graph keeps the y limits 0-1.
- If, someday, someone solves TF clustering for multipitch, that would be cool.
- `mir_eval.display.labeled_intervals(np.zeros((0, 2)), np.zeros((0,)))` gives a `ValueError`.  I think it would be a little more graceful if it just didn't plot anything, like `plt.plot([])`.  Same for the others - error shows up in `__expand_limits`.
- In `mir_eval.display.separation`'s docstring, `sources` is specified as a list of np.ndarray.  I think we should enforce that it be a np.ndarray of shape (n_src, n_sample), because if it's a ragged list, the function will puke.
- I wish that the output of `separation` was not quite so dependent on the order in which the sources are supplied.  Do you have any idea as to how we could adjust the threshold in terms of what gets turned transparent in each spectrogram (if that makes any sense)?  Here's an example of three sources from a MIDI file, the piano roll:

piano_roll

`mir_eval.display.separation(sources)`

sources

`mir_eval.display.separation([sources[2], sources[1], sources[0]])`

sources_2

- It is a little funky to me that only `separation` uses a legend, and also creates it automatically, whereas as fara s I can tell the other functions don't have a mechanism to create a legend at all.  Why is this?

bmcfee:
> I would expect the display functions to automatically make x and y axis labels (like in librosa).

In retrospect I think this is a bad convention.



> mir_eval.display.events(events, base=1., height=2.) displays nothing when using a new figure, because it does not change the boundaries from [0, 1], [0, 1] (same with other functions where you can change the plot range)

- [ ] Good call.  We can fix that.



> mir_eval.display.events(events, colors='k') doesn't change the line color to black (but plt.vlines(events, colors='k') does).

shouldn't it be `color='k'`?




> mir_eval.display.events(np.linspace(0, 10, 5) sets the xlim to [0, 10], which makes the first and last lines invisible (a Matplotlib policy, I think). I feel like it would be slick if it adjusted the range to always be slightly larger than the range of the events, for this reason.

That should be outside the scope of mir_eval; there's an open pr/discussion on matplotlib to do more intelligent automatic range limiting.

> So what exactly does the levels kwarg do in hierarchy? It doesn't seem to change the figure.

It's in the docstring.  If you put a legend on a hierarchy plot, `levels[i]` is the legend entry for the `i`th level.  This could be more clear.



> Sort of similar to above - mir_eval.display.labeled_intervals(intervals, labels, tick=False) on a new graph keeps the y limits 0-1.

What would you expect it to do?



> If, someday, someone solves TF clustering for multipitch, that would be cool.

I still think we shouldn't do that; instead, accept an input data structure that indexes tf values by contour id.



> ir_eval.display.labeled_intervals(np.zeros((0, 2)), np.zeros((0,))) gives a ValueError. I think it would be a little more graceful if it just didn't plot anything, like plt.plot([]). Same for the others - error shows up in __expand_limits.

- [ ] Good catch; will fix



> In mir_eval.display.separation's docstring, sources is specified as a list of np.ndarray. I think we should enforce that it be a np.ndarray of shape (n_src, n_sample), because if it's a ragged list, the function will puke.

I was going for consistency with the separation metrics, but sure.



> I wish that the output of separation was not quite so dependent on the order in which the sources are supplied. Do you have any idea as to how we could adjust the threshold in terms of what gets turned transparent in each spectrogram (if that makes any sense)? Here's an example of three sources from a MIDI file, the piano roll:

I tried a whole lot to work that out, but nothing seemed to do what I wanted.  Maybe it's possible if you render each spectrogram as RGBA image first, and then average in RGBA space.

craffel:
> In retrospect I think this is a bad convention.

Haha, I could see why you'd say that.  Fair enough.

> shouldn't it be color='k'?

From `vlines`' docstring:  "colors : array_like of colors, optional, default: 'k'"  So, either should work.  I'm not sure why `colors` isn't working in `events`.

> What would you expect it to do?

It's the same issue as above:

`mir_eval.display.labeled_intervals(intervals, labels)`

labeled_intervals

`mir_eval.display.labeled_intervals(intervals, labels, tick=False)`

labeled_intervals_tick_false

> I still think we shouldn't do that; instead, accept an input data structure that indexes tf values by contour id.

Yeah, or provide a kwarg which does that, or something.  But we don't need to worry about that now.

> I was going for consistency with the separation metrics, but sure.

I think they all specify e.g. `reference_sources : np.ndarray, shape=(nsrc, nsampl)` like [this](https://github.com/craffel/mir_eval/blob/master/mir_eval/separation.py#L55).

> I tried a whole lot to work that out, but nothing seemed to do what I wanted. Maybe it's possible if you render each spectrogram as RGBA image first, and then average in RGBA space.

It seems (from the point of view of an outsider who has no idea what's going on) that, under some threshold, alpha is set to 0.  Is there any easy way of modifying that threshold, so that more of the spectrogram blobs get made transparent?

bmcfee:
> .events(events, base=1., height=2.) displays nothing when using a new figure, because it does not change the boundaries

Can you give me an example that demonstrates this behavior?  I'm not sure what you mean by "new figure" (as opposed to new axes); when I do:

```python
fig = plt.gcf()
mir_eval.display.events(events, figure=fig)
```
it works as expected.  Are you doing something differently?



> From vlines' docstring: "colors : array_like of colors, optional, default: 'k'" So, either should work. I'm not sure why colors isn't working in events.

I think what's happening here is that `color` (popped from the cycler) supersedes `colors`; either will work in `vlines`, but `color` takes precedence.  This seems like a (documentation) error in matplotlib; i'm inclined to leave it as is, since it would otherwise take some pretty ugly contortions to work around with the axis cycler.

bmcfee:
> Is there any easy way of modifying that threshold, so that more of the spectrogram blobs get made transparent?

Not really.  The colormaps are generated by linear interpolation; what you're describing would require a nonlinear interpolation, and exposing that in a clean way to the user is non-trivial.

I also looked into making the maximum alpha value tuneable.  This is easy in mpl 2.0 via `mpl.colors.to_rgba()`, but in 1.5 it doesn't quite work.  (The difference comes in what kinds of inputs `to_rgba` supports in each version.)  We can revisit this in the future, but i think for now we should keep it as is.

bmcfee:
> It is a little funky to me that only separation uses a legend, and also creates it automatically, whereas as fara s I can tell the other functions don't have a mechanism to create a legend at all. Why is this?

Yeah, it sucks.  Basically, there's no way to attach legend entries (labels) to colormeshes, so we have to make proxy artists and label those instead.  To make that work, we need to first construct a legend, and there's no way to do that without drawing the legend.  (There might be a way to do it by making 0px-by-0px patches attached to the axes, but that seems ... perilous.)

At any rate, it doesn't bother me too much since the separation plots are completely uninterpretable without a legend, whereas the remaining plots may or may not be.

craffel:
> Can you give me an example that demonstrates this behavior? 

`mir_eval.display.events(np.arange(10))`

events

`mir_eval.display.events(np.arange(10), base=1, height=2)`

events_base_height

> I think what's happening here is that color (popped from the cycler) supersedes colors; either will work in vlines, but color takes precedence.

That makes sense.  The `colors` argument is useful in its own right though (allows for an array-like of len equal to `x` to be passed), so it may be confusing to end-users if it doesn't work with `events`.  Not a big deal though.

> We can revisit this in the future, but i think for now we should keep it as is.

Ok.

> Yeah, it sucks. Basically, there's no way to attach legend entries (labels) to colormeshes, so we have to make proxy artists and label those instead. To make that work, we need to first construct a legend, and there's no way to do that without drawing the legend.

Can we call `ax.legend().set_visible(False)` or something, and have things work the way we want them to?

bmcfee:
> Can we call ax.legend().set_visible(False) or something, and have things work the way we want them to?

Apparently not.  You can hide the legend that way, but then drawing it again doesn't work.  At best, it would have to be something like:

```python
ax = mir_eval.display.separation(some garbage)
ax.get_legend().set_visible(true)
```
instead of `ax.legend()`.  I'll take another stab at attaching proxies to the axis, but if it doesn't work, i don't see much of a problem in keeping it as is.

bmcfee:
Ok, got it working with 0x0 rect patches.  It's a brutal hack, but it works how you describe, and as an added bonus, makes it easier to manipulate the legend options.

Apart from the transparency thing in separation, I think I've hit all of @craffel's points.

bmcfee:
Okay, added a maximum alpha value for source separation display.  It doesn't achieve everything we'd want, but it does make it so that sources with comparable amplitudes (near max) don't mask each other out.

craffel:
> Okay, added a maximum alpha value for source separation display. It doesn't achieve everything we'd want, but it does make it so that sources with comparable amplitudes (near max) don't mask each other out.

Nice.  This and the legend looks nice.

Final nitpick: Setting `events`' `base` and `height` kwargs looks different than setting `vlines` equivalent `ymin` and `ymax` kwargs..  If you have a good reason it should be this way, that's ok, but it hurts my sensibilities a little bit.

`plt.vlines(np.arange(10), 10, 12)`

vlines

`mir_eval.display.events(np.arange(10), base=10, height=2)`

events

tests/test_display.py
```diff
@@ -346,6 +346,7 @@ def test_separation_label():
 
     plt.legend()
 
+
```
carlthome: image

bmcfee:
> Final nitpick: Setting events' base and height kwargs looks different than setting vlines equivalent ymin and ymax kwargs.. If you have a good reason it should be this way, that's ok, but it hurts my sensibilities a little bit.

It's done that way for consistency with `labeled_intervals` and `segments`.

craffel:
Sure, I'm just used to matplotlib fitting the axis limits to the content,
but we can do something different.  Unless there's anything left for you to
do, feel free to merge!

bmcfee:
> Sure, I'm just used to matplotlib fitting the axis limits to the content,
> but we can do something different.

Oh, I misunderstood -- I thought you were talking about the API.

I fixed the limit detection logic to do the right thing, but the API is designed for consistency across functions.

bmcfee:
@craffel btw, this should be ready to go.  If you want to run that last example through one more time, please do so before I merge.

craffel:
Merged, thank you again!

bmcfee:
:+1: :100:

Issue #192, opened 2016-05-16
Could bss_eval return min and max as well?


carlthome:
First of all, love this package! :dancer: 

**Curious if `bss_eval` could return the minimum and maximum SDR, SIR, SAR as well as the average?** I realize its based on Emmanuel Vincent's method but variance seems important in many cases. Lets say average SIR over a song is 10 dB. Roughly speaking, it could be the case that the chorus with lots of harmonic and percussive elements is easy to separate while the verse with, lets say less instrumentation, is harder to separate. Obviously one could pre-segment the signal but I figured asking doesn't hurt.

craffel:
Hey, thanks for the kind words.  Usually we focus on implementing evaluation metrics as widely used/specifically defined in research, so extensions like this are a little out of the scope of `mir_eval` unless they are starting to be utilized in publications or MIREX (or DCASE as the case may be).  Do you know of people reporting min/max too?  Otherwise, another goal of `mir_eval` is to make it very easy to modify the metrics so that they can be customized :)

carlthome:
Well, most recent papers seem to introduce their own additional evaluation aside from SDR, SIR, SAR, and many rely on listening tests still, but there's also this more recent framework (2011), [PEASS](http://bass-db.gforge.inria.fr/peass/), with objective and subjective metrics that promises to be more inline with perceptual quality, by the same people that proposed bss_eval originally back in 2006.

PEASS has three parts, one with objective metrics, one with subjective metrics derived from listening tests, and one part for making new listening tests. The first part could be a welcome addition to mir_eval maybe?

>Similarly to BSS Eval, the distortion signal is decomposed into three components: target distortion, interference, artifacts. These components are then used to compute four quality scores, namely OPS (Overall Perceptual Score), TPS (Target-related Perceptual Score), IPS (Interference-related Perceptual Score), APS (Artifact-related Perceptual Score). These scores better correlate with human assessments than the SDR/ISR/SIR/SAR measures of BSS Eval.

**I'd be pretty content with a minimum SDR, SIR and SAR though**, since that at least gives an idea of a lower bound on how poorly a separation method could be in a given time frame. I think many papers don't provide variance because they want to hide the fact that their model is not robust and has adversarial examples. For example when sources have similar energy distributions and the model discards phase information. Like, lets say a singer is particularly good at being in tune with the instrument and also just happens to display similar formants as the instruments' resonances. It seems to be surprisingly problematic, in my tiny experience.

[...]

Issue #68, opened 2014-07-25
Implement more separation metrics


craffel:
Although `bss_eval` is pretty standard, there are more:
http://bass-db.gforge.inria.fr/peass/
http://labrosa.ee.columbia.edu/projects/snreval/

faroit:
What about also supporting the [multichannel measures](http://bass-db.gforge.inria.fr/bss_eval/bss_eval_images.m)?

craffel:
Good idea!

carlthome:
PEASS would be a welcome addition, I think.

faroit:
Back to the state of bss_eval, I want trigger some discussion about what we could contribute to mir_eval and what would be the possible use cases, which is mainly the [SISEC Evaluation](https://sisec.inria.fr/home/2016-professionally-produced-music-recordings/) organised by @aliutkus this year.

#### Current implementation

* is a direct implementation of [bss_eval_sources v 3.0](http://bass-db.gforge.inria.fr/bss_eval/bss_eval_sources.m)

#### Issues/Missing features

##### Multichannel support 

Since many separation methods are now evaluating at least stereo mixes, supporting the `Image to Spatial distortion Ratio` would increase the usability of mir_eval a lot, since the current SISEC dataset is stereo and in evaluated using bss_eval_images

My student @ErikJohnsonCU and I have a working bss_eval_images pure python implementation which we could make ready for a PR into mir_eval. However, one of the issues that would need to be solved first, would be:

* how to do the unit tests regarding the test vectors which are mono. I would propose not add more test vectors to the repository, but instead generate fake stereo files from the existing mono files. 

* We can also add new json files for matlab compatibility comparisons

* @craffel how would you add the bss_eval_images? as a separate function or make a wrapper so that users can decide if they want the evaluate the separation on `sources` or `images`?

##### Frame wise evaluation

The [SISEC matlab based evaluation](https://github.com/faroit/dsd100mat) evaluates the audio signal frame wise instead of globally for the whole signal. We could add this to both the existing `sources` as well as to the new `images`. This would solve #192.

##### Speed

`bss_eval_sources.m` is slow. The mir_eval version is even slower. Would you be interested in a cython, numba... optimised version or would you prefer to have as few external dependencies as possible, hence simpler to install?

### DSD100

[...]

PR #199, opened 2016-06-08
Adding framewise evaluation for source separation

ErikJohnsonCU:
## Framewise Source Evaluation
A function has been added in separation.py (`bss_eval_sources_framewise`) that allows for framewise sources evaluation as discussed in #68, and which would be useful for min/max determination as discussed in #192.

## Optional Permutation Computation
For framewise evaluation the permutation computation is not terribly meaningful and an optional parameter (`compute_permutation`) was added to enable the permutation in `bss_eval_sources_framewise` (i.e. not computing permutations by default). An optional parameter (`compute_permutation`) was added to `bss_eval_sources` to disable permutation computation for applications not needing the permutation (makes the function considerably quicker!). The default for `bss_eval_sources` is entirely backwards compatible!

## Testing
Framewise regression and unit testing has been added to `test_separation.py` with modifications made to the `separation/output0*.json` files to allow them to have target results for both `bss_eval_sources` and `bss_eval_sources_framewise`.


carlthome:
Wonderful!

craffel:
Thanks for this!  I will get a chance to look it over soon-ish, I am in the midsf of a move.  If anyone else wants to chime in with code review and/or other comments, that would be cool too!

bmcfee:
@craffel I can CR this one if you'll do #196 :grin:

mir_eval/separation.py
```diff
@@ -107,7 +107,9 @@ def validate(reference_sources, estimated_sources):
                                                   MAX_SOURCES))
 
 
-def bss_eval_sources(reference_sources, estimated_sources):
+def bss_eval_sources(
+    reference_sources, estimated_sources, compute_permutation=True
+):
```
bmcfee: This seems like improper indenting -- does it pass pylint?

mir_eval/separation.py
```diff
@@ -161,28 +166,139 @@ def bss_eval_sources(reference_sources, estimated_sources):
...
+             reference_sources,
+    ...      estimated_sources)
+
+    Parameters
+    ----------
+    reference_sources : np.ndarray, shape=(nsrc, nsampl)
+        matrix containing true sources
+    estimated_sources : np.ndarray, shape=(nsrc, nsampl)
+        matrix containing estimated sources
+    win : int, optional
```
bmcfee: Is it required to have the same number of sources in reference and estimate?  If so, that should be explicitly stated.  If not, use different names in the shape descriptions.

mir_eval/separation.py
```diff
@@ -161,28 +166,139 @@ def bss_eval_sources(reference_sources, estimated_sources):
...
+
+    # compute the criteria across all windows
+    SDR = np.empty((nsrc, nwin))
+    SIR = np.empty((nsrc, nwin))
+    SAR = np.empty((nsrc, nwin))
+    perm = np.empty((nsrc, nwin))
+
+    for k in range(nwin):
+        K = slice(k * hop, k * hop + win)
+        SDR[:, k], SIR[:, k], SAR[:, k], perm[:, k] = bss_eval_sources(
```
bmcfee: It would be better to use a different variable name here.  `K` and `k` are too similar, and neither is descriptive.

mir_eval/separation.py
```diff
@@ -161,28 +166,139 @@ def bss_eval_sources(reference_sources, estimated_sources):
...
+            reference_sources[:, K],
+            estimated_sources[:, K],
+            compute_permutation
+        )
+
+    return SDR, SIR, SAR, perm
```
bmcfee: Maybe switch to lower-case variables to make lint happy an dbe consistent with the rest of the module.

mir_eval/separation.py
```diff
@@ -161,28 +166,139 @@ def bss_eval_sources(reference_sources, estimated_sources):
...
+        mean_sir = np.empty(len(perms))
+        dum = np.arange(nsrc)
+        for (i, perm) in enumerate(perms):
+            mean_sir[i] = np.mean(sir[perm, dum])
+        popt = perms[np.argmax(mean_sir)]
+        idx = (popt, dum)
+        return (sdr[idx], sir[idx], sar[idx], popt)
+    else:
+        # return the default permutation for compatibility
+        popt = range(nsrc)
```
bmcfee: Should this be `list(range(nsrc))` or `np.arange(nsrc)`?  In python 3, this will return a generator, which is probably not what you want here.

mir_eval/separation.py
```diff
@@ -161,28 +166,139 @@ def bss_eval_sources(reference_sources, estimated_sources):
...
+
+    Parameters
+    ----------
+    reference_sources : np.ndarray, shape=(nsrc, nsampl)
+        matrix containing true sources
+    estimated_sources : np.ndarray, shape=(nsrc, nsampl)
+        matrix containing estimated sources
+    win : int, optional
+        Window length for framewise evaluation
+    hop : int, optional
```
bmcfee: Docstring says `win` is optional, but it does not appear to be in the code.

mir_eval/separation.py
```diff
@@ -359,9 +485,22 @@ def evaluate(reference_sources, estimated_sources, **kwargs):
     # Compute all the metrics
     scores = collections.OrderedDict()
 
-    sdr, sir, sar, perm = util.filter_kwargs(bss_eval_sources,
-                                             reference_sources,
-                                             estimated_sources, **kwargs)
+    if win is not None and hop is not None:
```
bmcfee: What if only one of them is `None`?

mir_eval/separation.py
bmcfee: Any reason to call it `win` instead of `window`?  We use `window` throughout the rest of the package.

`hop` is fine.

tests/test_separation.py
```diff
@@ -68,6 +68,52 @@ def __unit_test_separation_function(metric):
...
+        assert str(w[-1].message) == ("estimated_sources is empty, "
+                                      "should be of size (nsrc, nsample).  "
+                                      "sdr, sir, sar, and perm will all be "
+                                      "empty np.ndarrays")
+        # And that the metric returns empty arrays
+        assert np.allclose(
+            metric(np.array([]), np.array([]), 40, 20), np.array([])
+        )
+
+    # Test for error when there is a silent reference/estimated source
```
bmcfee: Each of these failure tests should be a separate test function.

tests/test_separation.py
```diff
@@ -68,6 +68,52 @@ def __unit_test_separation_function(metric):
...
+
+    # Test for error when too many sources are provided
+    sources = np.random.random_sample((mir_eval.separation.MAX_SOURCES*2, 400))
+    nose.tools.assert_raises(ValueError, metric, sources, sources, 40, 20)
+
+    # Test for invalid win/hop parameter detection
+    est_sources = np.random.random_sample((4, 100))
+    nose.tools.assert_raises(
+        ValueError, metric, ref_sources, est_sources, 120, 20
+    )  # test with window larger than source lengths
```
bmcfee: Should we really consider this a value error, instead of just treating it as a degenerate case with a single (incomplete) frame?  I don't have an opinion on this, but it's not clear to me what the expected behavior is.

bmcfee:
@ErikJohnsonCU @craffel I did a quick pass over the PR, mainly checking for style and documentation.  Overall it looks pretty good, but I did spot one py2/py3 quirk (returning `range`) that should probably have a test case on it.

The implementation seems simple enough that I trust the accuracy of the regression tests, but @dawenl is probably a better person to validate that end of things.

mir_eval/separation.py
ErikJohnsonCU: It is required that the number of sources and samples be the same. I'll add a commenting stating that to the functions.

mir_eval/separation.py
ErikJohnsonCU: If window is used throughout the package, I'll replace win with window.

mir_eval/separation.py
ErikJohnsonCU: Good point! That would be a user mistake and I'll throw a descriptive error message.

tests/test_separation.py
ErikJohnsonCU: For this unit test function I was following the style of `__unit_test_separation_function` on line 37. I could divide them both into multiple failure tests if you want?

tests/test_separation.py
ErikJohnsonCU: A window with only a single frame would most likely be a user mistake and I think it would be better to throw an error than evaluate a frame that doesn't capture all the data. But maybe someone with experience using this type of evaluation can give some input. @faroit or @carlthome ?

mir_eval/separation.py
faroit: It does (see https://www.python.org/dev/peps/pep-0008/#indentation) and I like new new lines a lot...
Anyway, I changed it now so that the indentation style matches the rest of the package 😄

tests/test_separation.py
carlthome: I have no particular preference. :sunglasses: 

I think the user could be forced to pad with silence etc. just to make sure they realize what's happening, but perhaps it makes more sense to do as @bmcfee proposes (and its more consistent with librosa, mir_eval, jams' expected behavior overall :+1: ) considering that's typical in other fields (computer vision and applying image kernels, comes to mind) and just providing a warning instead that SDR might be a little too good.

mir_eval/separation.py
carlthome: In a better world: https://github.com/google/yapf :sunglasses:

ErikJohnsonCU:
Thanks for the feedback @bmcfee! I think I've addressed all your concerns in the latest commit (except for separating the unit testing which I'll wait for feedback on before I separate the existing unit test). Do we want to squash the commits before merging? @craffel

tests/test_separation.py
bmcfee: Fair enough; I'll defer to @craffel on this.

In general, I would expect each unit test to have exactly one call to the function being tested (`metric`).  I count at least seven in this test, and they're each testing slightly different failure modes.

mir_eval/separation.py
```diff
@@ -108,7 +108,7 @@ def validate(reference_sources, estimated_sources):
 
 
 def bss_eval_sources(
-    reference_sources, estimated_sources, compute_permutation=True
+        reference_sources, estimated_sources, compute_permutation=True
```
bmcfee: Still bad indent style :grin: 

should be

```ptython
def bss_eval_sources(reference_sources, estimated_sources,
                     compute_permutation=True, ...
```

but i won't be a stickler on this.

mir_eval/separation.py
```diff
@@ -207,17 +209,12 @@ def bss_eval_sources(
         return (sdr[idx], sir[idx], sar[idx], popt)
     else:
         # return the default permutation for compatibility
-        popt = range(nsrc)
+        popt = list(range(nsrc))
```
bmcfee: Docs say `ndarray`, but this returns a list.

mir_eval/separation.py
faroit: yes, sorry, had this in 8e1313 but that commit was accidentally reverted

mir_eval/separation.py
ErikJohnsonCU: Whoops, sorry, I was too focused on fixing the iterator mistake... everything should be an ndarray.

ErikJohnsonCU: Currently (ie. on line 207 and in master) `popt` is returned as a `tuple`, not an `ndarray`. Should the documentation be changed to reflect this or should the return type be changed to match the documentation? I'd lean towards a documentation change to minimize impact on anyone currently using with function. Either choice I will change the default permutation to match this type. @craffel

carlthome:
Ready for merge?

ErikJohnsonCU:
I want to fix the return type issue on the permutations before this merge goes through. The current return type is `tuple` (despite saying `ndarray` in the documentation). Should I change the permutation return type to `ndarray` (as the doc string says) or change the doc string? @craffel

bmcfee:
My vote would be for returning the permutation as an `ndarray`, because then you can use it directly as an index to the source array.  Tuples will not work for this, but I think it's reasonable expected behavior to use the permutation this way.

@craffel dis/agree?

craffel:
Sorry again for the delay on this.  I will be able to look it over more thoroughly next week.  I agree that an `ndarray` makes most sense.  Thanks for catching this.

mir_eval/separation.py
```diff
@@ -205,10 +205,10 @@ def bss_eval_sources(reference_sources, estimated_sources,
             mean_sir[i] = np.mean(sir[perm, dum])
         popt = perms[np.argmax(mean_sir)]
         idx = (popt, dum)
-        return (sdr[idx], sir[idx], sar[idx], popt)
+        return (sdr[idx], sir[idx], sar[idx], np.asarray(popt))
     else:
         # return the default permutation for compatibility
-        popt = list(range(nsrc))
+        popt = np.asarray(list(range(nsrc)))
```
faroit: maybe `np.arange(nsrc)` would be easier here

tests/test_separation.py
ErikJohnsonCU: Yeah, you're right... The code duplication is bad enough for having 2 unit tests, but it will get much worse if/when we add image and image framewise evaluation.. Having multiple test cases allows them to be reused across the eventual 4 functions. Good catch!

ErikJohnsonCU:
Have a look at this handling of the unit tests @bmcfee and let me know if that looks better. I'm calling metric more than once in a few of the tests, but for those it logically makes sense to group the calls together.

mir_eval/separation.py
```diff
@@ -88,19 +88,10 @@ def validate(reference_sources, estimated_sources):
                          'source to be non-silent, having a silent estiamted '
                          'source will result in an underdetermined system.')
 
-    if estimated_sources.shape[0] > MAX_SOURCES:
+    if estimated_sources.shape[0] > MAX_SOURCES or \
+            reference_sources.shape[0] > MAX_SOURCES:
```
bmcfee: Use parentheses for continuations instead of backslash escaping.

bmcfee:
> Have a look at this handling of the unit tests @bmcfee and let me know if that looks better. I'm calling metric more than once in a few of the tests, but for those it logically makes sense to group the calls together.

LGTM.

I added one more comment on the code itself, but it's an easy one.

bmcfee:
I'd suggest running pylint over the modified files just to catch anything else that I may have missed, but otherwise LGTM.

carlthome:
Status? Ready for merge?

ErikJohnsonCU:
@bmcfee pylint shows nothing new vs master

@carlthome I'd say it's ready, but I just really want to proceed to adding support for images. :smile:

craffel:
Hey all, sorry for being MIA.  Finally have some time on my plate so I will take a quick look tonight.  Thanks again for this.

mir_eval/separation.py
```diff
@@ -52,9 +52,9 @@ def validate(reference_sources, estimated_sources):
 
     Parameters
     ----------
-    reference_sources : np.ndarray, shape=(nsrc, nsampl)
+    reference_sources : np.ndarray, shape=(nsrc1, nsampl1)
```
craffel: Since these are required to be the same shape (as per the `ValueError` when `reference_sources.shape != estimated_sources.shape`), why change the docstring like this?

mir_eval/separation.py
```diff
@@ -88,26 +88,18 @@ def validate(reference_sources, estimated_sources):
                          'source to be non-silent, having a silent estiamted '
                          'source will result in an underdetermined system.')
 
-    if estimated_sources.shape[0] > MAX_SOURCES:
+    if (estimated_sources.shape[0] > MAX_SOURCES or
+            reference_sources.shape[0] > MAX_SOURCES):
         raise ValueError('The supplied matrices should be of shape (n_sources,'
-                         ' n_samples) but estimated_sources.shape[0] = {} '
-                         'which is greater than '
+                         ' n_samples) but n_sources = {} which is greater than'
```
craffel: This wording is less clear than stating that one of the arguments has the wrong shape, and we haven't defined "n_sources" anywhere (only nsrc).  Since we already checked that `estimated_sources.shape == reference_sources.shape`, if you want to collapse these two checks, instead of writing "n_sources", write "estimated_sources.shape[0] = {} and reference_sources.shape[0] = {}".

mir_eval/separation.py
```diff
@@ -161,28 +158,137 @@ def bss_eval_sources(reference_sources, estimated_sources):
...
-    perms = list(itertools.permutations(range(nsrc)))
-    mean_sir = np.empty(len(perms))
-    dum = np.arange(nsrc)
-    for (i, perm) in enumerate(perms):
-        mean_sir[i] = np.mean(sir[perm, dum])
-    popt = perms[np.argmax(mean_sir)]
-    idx = (popt, dum)
-    return (sdr[idx], sir[idx], sar[idx], popt)
+    # does user desire permutations?
+    if compute_permutation:
```
craffel: Can we collapse these two if-statement blocks into one?  E.g. only have one `if compute_permutation`.  I think we can also simplify the logic a bit to avoid most of the special-casing, but we can table that.

mir_eval/separation.py
```diff
@@ -161,28 +158,137 @@ def bss_eval_sources(reference_sources, estimated_sources):
...
+        idx = (popt, dum)
+        return (sdr[idx], sir[idx], sar[idx], np.asarray(popt))
+    else:
+        # return the default permutation for compatibility
+        popt = np.arange(nsrc)
+        return (sdr, sir, sar, popt)
+
+
+def bss_eval_sources_framewise(reference_sources, estimated_sources,
+                               window, hop, compute_permutation=False):
```
craffel: Shouldn't `window` and `hop` have default values?

mir_eval/separation.py
```diff
@@ -328,7 +434,8 @@ def _safe_db(num, den):
     return 10 * np.log10(num / den)
 
 
-def evaluate(reference_sources, estimated_sources, **kwargs):
+def evaluate(reference_sources, estimated_sources,
+             window=None, hop=None, **kwargs):
```
craffel: Assuming `window` and `hop` indeed can/should have default values, `window` and `hop` should not be kwargs, they should be passed in **kwargs and you should use `util.filter_kwargs` (see other `evaluate` functions for examples of this).  Then you can avoid all of the special-casing, and just iterate over all of the functions in the submodule.  The way it is is a little incon

craffel:
Just posted a few comments and questions. I apologize if they have already been discussed, please point me in the right direction if they have.  Most are minor, my main concern is the non-default `window` and `hop` arguments which were created.  There aren't any other metrics in `mir_eval` which have non-default parameters, other than the actual annotations (correct me if I'm wrong) - this is a conscious decision, because of course the output of the metric could change significantly if two users passed different parameters, and forcing every user to come up with these parameters is sure to result in inconsistent evaluation.  I am guessing you're doing this because we don't know the sampling rate (if there's another reason, let me know).  In that case, I'd suggest requiring that the sampling rate be specified, and then compute a consistent window/hop size based on that - the sampling rate is arguably a part of the "annotations" themselves.  Alternatively, if you want to just set defaults for `window` and `hop`, you can add a note to the docstring saying "These values were chosen assuming a sampling rate of xyz, if you're supplying data with another sampling rate, you should correct them".  Anyways, let's discuss this and determine the best way forward.  Thanks!

mir_eval/separation.py
ErikJohnsonCU: You're right; my thinking here was confused.

ErikJohnsonCU: My thinking was that we haven't yet verified this condition (which is the purpose of this function) so stating they are the same size would invalidate the purpose of verifying their sizes.

mir_eval/separation.py
ErikJohnsonCU: You're right this is relatively unclear (and I don't know where n_sources and n_samples came from).. I'll use the format you mention and correct the references to (n_sources, n_samples).

ErikJohnsonCU:
Thanks for the feedback! I've addressed the comments not relating to the `window` and `hop` parameters for `bss_eval_sources_framewise` function. The issue here is that the `window` and `hop` parameters are application dependent, not just sample rate dependent. While we could say that the default window should be 1s at 44.1kHz and the default hop should be 0.5s at 44.1kHz, this would not work for arrays with `nsampl<66150` (i.e. 2 windows) and would take a prohibitive length of time for audio over a minute. This evaluation method would likely be used by those with knowledge of source separation and the meaning of the metrics they would receive.

carlthome:
Default arguments are sweet as a guideline for documenting sensible values though, if nothing else. It improves the documentation, IMHO.

mir_eval/separation.py
craffel: Well, you could make the same argument for all of the metrics functions themselves - we have not verified them either.  To me, the docstring typically specifies "What input to this function is expected and will not cause an exception?" which is the convention through `mir_eval`.

craffel:
> The issue here is that the window and hop parameters are application dependent, not just sample rate dependent.

Following convention here, we should have multiple functions with sensible defaults for each application.  `mir_eval` tries to implement community-standard metrics, and that standardization includes parameterization.  If there are no agreed-upon parameters in the community, that's somewhat problematic.

> This evaluation method would likely be used by those with knowledge of source separation and the meaning of the metrics they would receive.

We can't really make this assumption - while I would expect people calling this function to at least know what a window and hop size are, I wouldn't expect them to know the conventions as far as how to set them (and in different applications).  I mean, I have worked on source separation a bit, and I would not know what the standard setting for them was in pretty much any application.


All this being said, I recognize that source separation is a bit different from the rest of `mir_eval`, so if there is general agreement that having default parameter values for the hop and window doesn't make any sense, I'm happy to defer to community judgement.

aliutkus:
hi there!
I could answer maybe to this point as the chair of the [SiSEC evaluation](https://sisec.inria.fr/home/2016-professionally-produced-music-recordings/) for source separation.

Although any length is arguably good or wrong, we took the decision for the MUS task since 2015 to settle on 30s frames with 15s overlap !

it turned out we were satisfied with this choice. It would be nice if mireval had this as a default value indeed !

cheers

ErikJohnsonCU:
Perfect! Thanks @aliutkus! In that case, using those lengths with a sample rate of 44.1kHz gives default values for `window` and `hop` or 1323000 and 661500 respectively; however, I'm unclear on how the `evaluate` function would work in that case.. how could we detect the user's intention to use `bss_eval_sources` or `bss_eval_sources_framewise`?

mir_eval/separation.py
ErikJohnsonCU: Yeah, my logic for that sounds hollow to me. I agree, that should be the convention for the docstring. I'll change that back.

carlthome:
How about a bool argument `global=True` so it's completely backwards compatible? Framewise only if False.

aliutkus:
I would say it makes no sense to use 'no framing' for very long signals (e.g. several minutes).
Personnally, I would ALWAYS use the framing, and average the scores obtained in the different  windows if I wanted a single statistics (that would then come along a std).

I think that users should be free to run with no windowing all the same, but what I would do myself is the following: 
in the bsseval_images or bsseval_sources, if the input is more than 30s, I would generate a warning stating that the signal is very long and that using framing should be considered. And then proceed using no framing

ErikJohnsonCU:
@carlthome I think the idea of a bool argument that allows for backwards compatibility is a good one (and will cause minimal test impacts)!

@aliutkus I like the idea of generating a warning for long input (if that's consistent with `mir_eval`'s conventions). For determining what's a "very long" signal, is assuming a sample rate of 44.1kHz (as done with the framewise parameters) acceptable?

aliutkus:
very long= more than 30s (the default for the windowed version)

please also make sure than when calling the windowed version on short signals, the output is the same than the non-windowed version

ErikJohnsonCU:
Currently calling the windowed version on a signal where only a single window would be used raises an exception (e.g. with 30s window and 15s hop that would be a signal less than 45s). Do you think this is the right thing to do? It would require a user to explicitly call the non-windowed version for short signals.

aliutkus:
yeah I think  calling the nonwindowed version for short signals should be automatic, without even raising a warning

ErikJohnsonCU:
I think the danger with that is a user might expect the result to be only using a single window instead of the whole signal. Then the result from calling windowed will differ from if they manually sliced a window with their desired length and used the non-windowed evaluation. Documentation could aid in clarifying this, but I think it would be dangerous to potentially cause this confusion. Raising an error/warning, forces the user to explicitly use the non-windowed evaluation on their segment of interest.

aliutkus:
anyways, you have the problem for the last frame right ? How handling that one ? zero padding ?

ErikJohnsonCU:
This is a good point. Currently no fully complete frame is evaluated.. should the last frame be 0 padded? I'm not 0 padding as that was what the dsd100mat SISEC code did (I believe, please correct me if I'm wrong here).

aliutkus:
you should not be zero padding, because that will bias the result towards good scores: 0 groundtruth, 0 estimate is perfect ! So really, you should process the short frames using the non-windowed version silently, well at least that's my thinking about it

ErikJohnsonCU:
To clarify: I said the current implementation (and the SISEC implementation) are NOT zero padding. I'm sorry for causing confusion.

mir_eval/separation.py
```diff
@@ -328,7 +438,8 @@ def _safe_db(num, den):
     return 10 * np.log10(num / den)
 
 
-def evaluate(reference_sources, estimated_sources, **kwargs):
+def evaluate(reference_sources, estimated_sources,
+             framewise=False, window=30*44100, hop=15*44100, **kwargs):
```
craffel: 1. Since you are using `**kwargs` and `filter_kwargs`, you should not add `window` and `hop` as separate keyword arguments.  They already can be supplied thanks to `**kwargs`, and they will not be passed to functions which do not accept them thanks to `filter_kwargs`.

1. Why add a flag for computing framewise metrics or not?  The `mir_eval` convention is that `evaluate` calls all of the different evaluation functions, so in this case we should just call (and return scores for) both the framewise and non-framewise functions.

craffel:
Thanks for this excellent discussion and the updates. I just made two comments about the `evaluate` function.  Once we wrap those up, we should be all set to go.  Thanks!

mir_eval/separation.py
ErikJohnsonCU: 1. Ahh! Sorry, I was confused about where the default should be defined.. I'll make that change now.

2. Okay, I'll have it call both. I'll do a bit of reworking on the framewise function to incorporate the feedback from @aliutkus vis-a-vis handling short frames.

ErikJohnsonCU: For 2. is it better if I return a nested dictionary or should each part have it's own name? (Also, this approach will not work for images support, as the existing function cannot be called on images...)

craffel: > For 2. is it better if I return a nested dictionary or should each part have it's own name?

It should return a flat dictionary, each metric having its own name, e.g. `'Source to Distortion (framewise)'`, etc.

> (Also, this approach will not work for images support, as the existing function cannot be called on images...)

Why is that?  We can discuss this elsewhere (probably in the images issue).

ErikJohnsonCU:
@craffel If you're happy with how I'm returning the metrics, I'd say this PR is ready.

craffel:
Great!  Looks good to me - since this ended up being a lot of commits from two different people, I squashed into 6 commits rather than one, which hopefully are somewhat coherent/atomized.  Merged in https://github.com/craffel/mir_eval/compare/5887679a519f313cd275017cff7d4ce87c23cf00...4425853f27d9039bd48c985fe4ef0c29b53a9385 Thank you again for your work, looking forward to getting images merged too!

Issue #202, opened 2016-06-10
sorting intervals -- OR -- interpolate_intervals bug


bmcfee:
[porting over from https://github.com/marl/jams/issues/116 ]

I ran into the following error:

```python
>>> import mir_eval
>>> import numpy as np
>>> ints = np.asarray([[1, 2], [0, 1]])
>>> labs = list('ab')
>>> mir_eval.segment.evaluate(ints, labs, ints, labs)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-9-dc90e2bfeebe> in <module>()
      3 labs = list('ab')
      4 
----> 5 mir_eval.segment.evaluate(ints, labs, ints, labs)

/home/bmcfee/git/mir_eval/mir_eval/segment.py in evaluate(ref_intervals, ref_labels, est_intervals, est_labels, **kwargs)
   1099                                                         ref_labels,
   1100                                                         est_intervals,
-> 1101                                                         est_labels, **kwargs)
   1102 
   1103     # Rand index

/home/bmcfee/git/mir_eval/mir_eval/util.py in filter_kwargs(_function, *args, **kwargs)
    832             filtered_kwargs[kwarg] = value
    833     # Call the function with the supplied args and the filtered kwarg dict
--> 834     return _function(*args, **filtered_kwargs)
    835 
    836 

/home/bmcfee/git/mir_eval/mir_eval/segment.py in pairwise(reference_intervals, reference_labels, estimated_intervals, estimated_labels, frame_size, beta)
    382 
    383     # Find where they agree
--> 384     matches = np.logical_and(agree_ref, agree_est)
    385     n_matches = (matches.sum() - len(y_ref)) / 2.0
    386 

ValueError: operands could not be broadcast together with shapes (20,20) (10,10) 

```
I think the error is that `util.interpolate_intervals()` assumes the intervals are sorted, but this is not stated in the docstring.

In general, I don't think we should require the intervals to be sorted, since the boundaries are all explicit.  It would be easy to add a sorting step to `interpolate_intervals` and the rest should fix itself.

@craffel what you say?

bmcfee:
This issue actually pops up again in all the segment metrics because `validate_structure` requires `intervals[0, 0] == 0`.

[...]

PR #203, opened 2016-06-10
support unsorted segment intervals

bmcfee:
This is a bugfix for issue #202.

### DIFF
--- a/mir_eval/segment.py
+++ b/mir_eval/segment.py
@@ -45,13 +45,14 @@ Metrics
   estimated segments
 '''
 
+import collections
+import warnings
+
 import numpy as np
 import scipy.stats
 import scipy.sparse
 import scipy.misc
 import scipy.special
-import collections
-import warnings
 
 from . import util
 
@@ -131,7 +132,7 @@ def validate_structure(reference_intervals, reference_labels,
         # Check only when intervals are non-empty
         if intervals.size > 0:
             # Make sure intervals start at 0
-            if not np.allclose(intervals[0, 0], 0.0):
+            if not np.allclose(intervals.min(), 0.0):
                 raise ValueError('Segment intervals do not start at 0')
 
     if reference_intervals.size == 0:
@@ -140,8 +141,8 @@ def validate_structure(reference_intervals, reference_labels,
         warnings.warn("Estimated intervals are empty.")
     # Check only when intervals are non-empty
     if reference_intervals.size > 0 and estimated_intervals.size > 0:
-        if not np.allclose(reference_intervals[-1, 1],
-                           estimated_intervals[-1, 1]):
+        if not np.allclose(reference_intervals.max(),
+                           estimated_intervals.max()):
             raise ValueError('End times do not match')
 
 
@@ -348,7 +349,6 @@ def pairwise(reference_intervals, reference_labels,
         F-measure of detecting whether frames belong in the same cluster
 
     """
-
     validate_structure(reference_intervals, reference_labels,
                        estimated_intervals, estimated_labels)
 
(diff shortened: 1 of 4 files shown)


bmcfee:
TODO:

- Support unsorted intervals in `util.adjust_intervals`
- In general, replace implicit start time `intervals[0, 0]` and end-time `intervals[-1, 1]` by calls to `min()` and `max()` unless we've explicitly sorted the data already.
- Rewrite `util.interpolate_intervals` to support discontinuities; this may be better done as a separate PR though.

bmcfee:
I think we should punt on rewriting `interpolate_intervals` since it's non-essential to fixing the current issue with implicitly ordered timings.  AFAIK, there's no need for label interpolation with incomplete interval coverage, so we can deal with that if/when it comes up.

Otherwise, I think this one's ready for CR from @craffel or @urinieto .  (Should be an easy one.)

bmcfee:
(PS: speedy resolution of this would be appreciated, as it's stalling a project i'm working on.)

mir_eval/segment.py
```diff
@@ -131,7 +132,7 @@ def validate_structure(reference_intervals, reference_labels,
         # Check only when intervals are non-empty
         if intervals.size > 0:
             # Make sure intervals start at 0
-            if not np.allclose(intervals[0, 0], 0.0):
+            if not np.allclose(intervals.min(), 0.0):
```
craffel: This, and the max call below, could be made 2x faster as `intervals[:, 0].min()` and `intervals[:, 1].max()` respectively.

mir_eval/util.py
```diff
@@ -138,6 +138,8 @@ def interpolate_intervals(intervals, labels, time_points, fill_value=None):
         The ``i`` th interval spans time ``intervals[i, 0]`` to
         ``intervals[i, 1]``.
 
+        Intervals are assumed to be disjoint.
```
craffel: Is this what you mean?  You wrote "Rewrite util.interpolate_intervals to support discontinuities".  From that statement it sounds like the problem is that the intervals are assumed to fully cover the track, not that they are assumed to be disjoint.  Maybe I am confused.

bmcfee: Those are two different things.

Currently, the interpolator assumes that intervals are disjoint and cover some range `[t_min, t_max]`; the fill value gets applied outside the interval.

Disjointness is required for the labeling process to be a function, so I just made that requirement explicit in the docstring.

In the future, I see no reason to require full coverage in general, ie, a gap in the middle should be fine.  However, at present it will do the wrong thing here and just propagate the label from the interval preceding the gap.

mir_eval/segment.py
bmcfee: Nope :grin: the slice construction dominates runtime for the range of values we care about.

```python
N = 100
x = np.random.randn(N, 2)

%timeit x.min()
100000 loops, best of 3: 1.9 µs per loop

%timeit x[:, 0].min()
100000 loops, best of 3: 1.92 µs per loop
```

craffel: Well, there you go.

mir_eval/util.py
craffel: Ok.

craffel:
Looks reasonable to me.  I am not an expert on `interpolate_intervals`, so it might be useful if another user chimed in to make sure the change makes sense, but if it's doing what you want it to do it seems ok to me.

bmcfee:
Cool, thanks.

I didn't make any substantive changes to `interpolate_intervals` (all previous tests still pass).  The function itself is strange and a little broken IMO, but this isn't the right place to fix it.

The logic is: for each time point `t`, find the first interval `k` that starts after `t`, go back to the previous segment `k-1`, and apply that segment's label to `t`.  This obviously requires ordered intervals to work, ie, `interval[k-1]` immediately precedes `interval[k]`.

craffel:
Ok, cool.

bmcfee:
@craffel thanks for the CR.  Unless you object, I think this is merge-ready.

craffel:
Feel free to squash and merge.

PR #207, opened 2016-07-06
Bss eval images

ErikJohnsonCU:
## Images Separation Evaluation

A function has been added in `separation.py` (`bss_eval_images`) that allows for image separation evaluation as discussed in #68. This implementation is identical to that of the `BSS_eval` MATLAB implementation.

## Implementation Testing

In order to validate the consistency of `bss_eval_images` with that of the MATLAB implementation, comparison testing has been performed. This comparison testing is available for anyone wishing to test it [here](https://github.com/ErikJohnsonCU/mir_eval_images_compare) (please let me know if there are any issues installing/running it!).

## Regression Testing

A suite of tests have been added to `test_separation.py` for regression testing. These tests follow the same style as for `bss_eval_sources` and `bss_eval_sources_framewise` and the `output0*.json` files have been updated accordingly. Generation of dual channel audio is done by expanding both references and estimates to the correct number of channels and then manipulating the estimates in order to generate non-identical channels (see the `__generate_multichannel` function in `test_separation.py`).

**All feedback is welcome!**


ErikJohnsonCU:
I thought I'd create this PR before #199 goes through so you can all start having a look at it. I'll fold back any changes relating to #199 into this PR as well.

craffel:
Thanks for this.  I will give it a closer look once we merge #199, since that will make the diff easier to parse too.

First though, a question - so the images version essentially allows for/requires multichannel sources to be evaluated, whereas "sources" only allows mono sources?  If that's the case, can't we infer whether to call `images` or `sources` depending on whether the supplied numpy arrays have 2 or 3 dimensions?  And, is there a reason why there is not a framewise version for images?  Thanks.

ErikJohnsonCU:
I thought I'd include the framewise version of images in a separate PR so that it simplifies this one a little bit and we can focus on the details of full evaluation. I have a framewise version of images tested against the SISEC MATLAB implementation that just needs regression testing before a PR is submitted. I'll PR that when this one is close to going through.

The issue with using `images` for stereo (and higher numbers of channels) and `sources` for mono is that images can be used for mono sources and results in different values. This would not be a problem; however, *I BELIEVE* (correct me if I'm wrong @faroit @carlthome @aliutkus) that `images` is fairly often used to evaluate mono metrics due to their differences as explained by @aliutkus in #68 ([specifically here](https://github.com/craffel/mir_eval/issues/68#issuecomment-223808531))

ErikJohnsonCU:
I guess what I could do is always evaluate `images`, and only evaluate `sources` if the input has 2 dimensions (i.e. is a mono signal).

carlthome:
Honestly, having read https://github.com/craffel/mir_eval/issues/68#issuecomment-223808531 I think evaluating _sources_ seems terribly uninteresting for MIR, and that _images_ are the only sensible thing to estimate. I'd rather see dereverberation and stuff handled separately from source separation. Many datasets only provide stems with effects applied (reverb, dyncomp, etc.) anyway so having a reference source is rare.

craffel:
> I guess what I could do is always evaluate images, and only evaluate sources if the input has 2 dimensions (i.e. is a mono signal).

This seems reasonable to me.

ErikJohnsonCU:
I'm having a lot of difficulty recreating the failure occurring in the Travis CI build.. only one test case is failing on only two of the four measures. On my computer and @faroit's, these tests pass. I ran the test cases in a Docker build of a Travis CI worker (which is not identical to the one used here; I couldn't find a way to recreate this exact VM) and they passed there as well. Any thoughts on what might be causing this? (The failures are legitimate mismatches with the BSS_Eval MATLAB implementation.)

craffel:
Hmm, yes, that is quite strange.  Thank you for being so thorough trying to track down the reason.  It looks like the scores are _very_ different though: 
`[9.04398573049726, -4.324186590576983, 4.845951990328276, 3.996466612358616]`
vs
`[9.099154328454752, -0.6395994716351378, 4.933804894359991, 18.322131574769077]`
and
`[12.195644841027203, 0.6632345592643484, 4.898079314490429, 5.0964513442276695]`
vs.
`[12.290493375669527, 1.2044404549739196, 4.942972818383563, 13.322032007557937]`

I assume on your system the `mir_eval` scores are much closer to the expected scores?

ErikJohnsonCU:
On my system (and @faroit's) the answers are within `1e-12' (the `ATOL` value in `test_separation.py`).

`[9.099154328454752, -0.6395994716351378, 4.933804894359991, 18.322131574769077]`
and
`[12.290493375669527, 1.2044404549739196, 4.942972818383563, 13.322032007557937]`

Which are exactly the same as the comparisons. This makes sense since I generated the baseline scores on my system.

aliutkus:
what is the characteristics of these tests ? are there zeros in the sources for more than 30s somewhere ?

ErikJohnsonCU:
@aliutkus All the test audio is 1s long at 8kHz sampling. The second channel is generated by duplicating the mono audio and scaling it (by `0.25` in the case of the failing test) and reversing some of them. There are test cases where a silent channel is generated. It is important to note that this scaling and reversing is only applied to the estimates; the references are simply duplicated.

craffel:
Yes, this is bizarre and frustrating.  One guess: is there any way to confirm that the versions of packages being used on the Travis server and your local copies/Docker instance are exactly the same?

ErikJohnsonCU:
Yeah, the Travis build prints the version numbers for the packages and I've made sure they are the same.

craffel:
I have no idea, I think you have already done everything I'd check for.  @bmcfee any ideas? You have more Travis experience than me...

ErikJohnsonCU:
I'm in a bind with respect to coveralls now... it is impossible to reach 100% coverage for `separation.py` due to the necessity of excepting indeterminate systems in the code and it is impossible to determine input that would generate an indeterminate system. This is the only thing in `separation.py` that is not covered and tests I add for other modules don't count towards this coverage evaluation.

craffel:
> I'm in a bind with respect to coveralls now... it is impossible to reach 100% coverage for separation.py due to the necessity of excepting indeterminate systems in the code and it is impossible to determine input that would generate an indeterminate system. This is the only thing in separation.py that is not covered and tests I add for other modules don't count towards this coverage evaluation.

That's fine, we can still merge if coverage decreased 0.003% :) Is this otherwise ready for code review?

bmcfee:
> That's fine, we can still merge if coverage decreased 0.003% :)

Fun fact: coveralls lets you set absolute and relative thresholds for this sort of thing.

ErikJohnsonCU:
Yep, ready for code review!

mir_eval/separation.py
```diff
@@ -66,11 +66,27 @@ def validate(reference_sources, estimated_sources):
...
+    elif (reference_sources.ndim == 2 and
+            np.any(np.all(reference_sources == 0, axis=1))):
+        raise ValueError('All the reference sources should be non-silent (not '
+                         'all-zeros), but at least one of the reference '
+                         'sources is all 0s, which introduces ambiguity to the'
+                         ' evaluation. (Otherwise we can add infinitely many '
+                         'all-zero sources.)')
+    elif (reference_sources.ndim > 2 and
+            np.any(np.all(np.sum(reference_sources, axis=-1) == 0,
+                          axis=1))):
```
craffel: Can you combine the above two if-statements since they produce the same Exception?  It seems to me that you are effectively checking whether the trailing dimension is all-zeros, right?  If so, I am sure we can figure out some numpy wizardry to do this all in one case.  It also looks like you are only testing ndim >= 2, but isn't ndim == 1 possible too (which then gets converted to 2D by adding an axis)?

mir_eval/separation.py
```diff
@@ -81,7 +97,16 @@ def validate(reference_sources, estimated_sources):
...
-    elif np.any(np.all(estimated_sources == 0, axis=1)):
+    elif (estimated_sources.ndim == 2 and
+            np.any(np.all(estimated_sources == 0, axis=1))):
+        raise ValueError('All the estimated sources should be non-silent (not '
+                         'all-zeros), but at least one of the estimated '
+                         'sources is all 0s. Since we require each reference '
+                         'source to be non-silent, having a silent estiamted '
+                         'source will result in an underdetermined system.')
+    elif (estimated_sources.ndim > 2 and
+            np.any(np.all(np.sum(estimated_sources, axis=-1) == 0,
```
craffel: Same here with collapsing if-statements.

mir_eval/separation.py
```diff
@@ -300,6 +325,141 @@ def bss_eval_sources_framewise(reference_sources, estimated_sources,
...
+        vector of Sources to Artifacts Ratios (SAR)
+    perm : np.ndarray, shape=(nsrc,)
+        vector containing the best ordering of estimated sources in
+        the mean SIR sense (estimated source number perm[j] corresponds to
+        true source number j)
+        Note: perm will be (1,2,...,nsrc) if compute_permutation is False
+
+    """
+
+    # make sure the input has 3 dimensions
```
craffel: You should be able to replace all of the following code (up to line 391) with a single call to `np.atleast_3d`.

mir_eval/separation.py
```diff
@@ -337,6 +497,51 @@ def _bss_decomp_mtifilt(reference_sources, estimated_source, j, flen):
...
 
 
+def _bss_decomp_mtifilt_images(reference_sources, estimated_source, j, flen,
+                               Gj=None, G=None):
+    """Decomposition of an estimated source image into four components
+    representing respectively the true source image, spatial (or filtering)
+    distortion, interference and artifacts, derived from the true source
+    images using multichannel time-invariant filters.
+    Adapted version to work with multichannel sources.
+    Improved performance can be gained by passing Gj and G parameters.
```
craffel: In that case, maybe give a quick note as to what they are/do here since those are not descriptive names?

mir_eval/separation.py
```diff
@@ -422,6 +703,33 @@ def _bss_source_crit(s_true, e_spat, e_interf, e_artif):
     return (sdr, sir, sar)
 
 
+def _bss_image_crit(s_true, e_spat, e_interf, e_artif):
+    """Measurement of the separation quality for a given image in terms of
+    filtered true source, spatial error, interference and artifacts.
+
+    Parameters
```
craffel: You don't need to include these in an underscore function if you aren't specifying them.  Where they appear elsewhere was due to some auto-docstring-formatting code that we run at some point.

tests/test_separation.py
```diff
@@ -28,14 +28,37 @@ def __load_and_stack_wavs(directory):
...
         stacked_audio_data.append(audio_data)
     return np.vstack(stacked_audio_data)
 
 
+def __generate_multichannel(mono_sig, nchan=2, gain=1.0, reverse=False):
+    ''' Turn a single channel (ie. mono) audio sample into a multichannel
+    (e.g. stereo)
+    Note: to achieve channels of silence pass gain=0
+    '''
+    # add the channels dimension
```
craffel: Should be achievable by `np.atleast3d` too.

craffel:
Just made some minor comments.  All is good otherwise, I assume the impementation tests still check out?  Thank you again!

mir_eval/separation.py
ErikJohnsonCU: The reason this is a complicated check is that we want to accept any channel being entirely 0 (in the `ndim >= 2` case). Hence, the addition of the `np.sum(reference_sources, axis=-1)` which sums all the channels for the `ndim >= 2` case. Since the trailing dimension for `ndim >= 2` is the channels, a check of trailing dimension all-zeros will almost always be true (one sample has zeros on both channels).

ErikJohnsonCU: If there's a way to reduce an nD array to 2D by summing higher dimensions and then check the last dimension this could replace the two separate cases.

mir_eval/separation.py
ErikJohnsonCU: Is there a way to add a dimension at the beginning and at the end with `np.atleast_3d`? That's necessary for the `ndim == 1` case.

mir_eval/separation.py
ErikJohnsonCU: You mean I should remove the docstrings for all underscore functions? Why not leave them since they don't hurt and could help others understand the code. I could remove the parameters though since they aren't described.

mir_eval/separation.py
ErikJohnsonCU: That is exactly what it does! I guess I should have read the documentation more carefully for that function... Thanks!

tests/test_separation.py
ErikJohnsonCU: Now using `np.atleast_3d`.

mir_eval/separation.py
craffel: You got it.

mir_eval/separation.py
craffel: Sorry, I meant just the parameters.  The docstring description part is still useful!

mir_eval/separation.py
craffel: Like this?
```Python
In [1]: import numpy as np

In [2]: a = np.empty((2, 3, 4, 5, 6, 7, 8, 9))

In [3]: a.sum(axis=tuple(range(2, a.ndim))).shape
Out[3]: (2, 3)

In [4]: a = np.empty((2, 3))

In [5]: a.sum(axis=tuple(range(2, a.ndim))).shape
Out[5]: (2, 3)
```

mir_eval/separation.py
ErikJohnsonCU: Ahhhh! Okay, I'll take the parameter lists out for underscore functions.

mir_eval/separation.py
ErikJohnsonCU: That works perfectly! Thanks!

mir_eval/separation.py
ErikJohnsonCU: Have a look at the description I added and let me know if that's clear enough.

ErikJohnsonCU:
Implementation tests still check out. My pleasure!

tests/test_segment.py
```diff
@@ -67,6 +67,11 @@ def __unit_test_structure_function(metric):
         # And that the metric is 0
         assert np.allclose(score, 0)
 
+    # Test for non-matching numbers of intervals and labels
```
craffel: Haha, just noticed this - was this to try to maintain coverage?  Happy to include this but it's a bit off-topic.

mir_eval/separation.py
```diff
@@ -474,26 +702,40 @@ def evaluate(reference_sources, estimated_sources, **kwargs):
...
-    scores['Sources - Source to Artifact'] = sar.tolist()
-    scores['Sources - Source permutation'] = perm
+    scores['Images - Source to Distortion'] = sdr.tolist()
+    scores['Images - Image to Spatial'] = isr.tolist()
+    scores['Images - Source to Interference'] = sir.tolist()
+    scores['Images - Source to Artifact'] = sar.tolist()
+    scores['Images - Source permutation'] = perm.tolist()
+
+    # Verify we can compute sources on this input
+    if reference_sources.ndim < 3 and estimated_sources.ndim < 3:
```
craffel: It's probably worth making a note in the docstring about how sources will only be evaluated when the nchan dimension is not provided.

mir_eval/separation.py
craffel: So, since you're using np.atleast_3d you can omit the if statement!  np.atleast_3d will leave things as-is if the input is 3d.

craffel:
Couple more minor nitpicks - after that, ready to go, thanks againf or your patience!

tests/test_segment.py
ErikJohnsonCU: Yeah... :laughing: I guess coveralls is too smart to allow me to do that since I didn't change the associated source file. It should improve overall coverage though.

mir_eval/separation.py
ErikJohnsonCU: This is so true. Good catch again!

mir_eval/separation.py
ErikJohnsonCU: I added a note in the `evaluate` docstring. Let me know if that's clear and detailed enough.

craffel:
LGTM. Thanks again, merging!

carlthome:
@ErikJohnsonCU, testing this out now. It's working great. I'm just wondering one thing about `compute_permutation`. The documentation says

> Passing True for compute_permutation will much improve the performance of the evaluation; however, it is not always appropriate and is not the way that the BSS_EVAL Matlab toolbox computes bss_eval_images.

and I assume _performance_ refers to separation performance (and not computational performance considering `compute_permutation=False` is O(n) and `compute_permutation=True` is O(n²)) but why would it be affected? It's only if one actually aggregates the local scores, right? Then one is "cheating" because for every evaluated frame a sorting based on SIR to achieve the largest possible global measure is done. Could you perhaps elaborate a bit in the documentation (like, provide an example when `compute_permutation=True` is not appropriate) or point me in the right direction where I can read up on this?

ErikJohnsonCU:
@carlthome Sorry, I may not have been clear enough in the comment: the performance I'm talking about is ONLY computation performance; that is, the results will be identical just without determining the best ordering of sources. If estimate x corresponds to reference x (as is normally the case), the results with both values of `compute_permutation` will be identical.

faroit:
@carlthome 

>  It's only if one aggregates the local scores, right? Then one is "cheating" because for every evaluated frame a sorting based on SIR to achieve the largest possible global measure is done. Could you perhaps elaborate a bit in the documentation, like provide an example when compute_permutation=True is not appropriate and when it is?

I think it's all in [[Vincent 2006]](https://hal.inria.fr/inria-00544230/PDF/vincent_TASLP06bis.pdf):
screen shot 2016-07-28 at 12 21 41

 You might call it "cheating" others call it "evaluating the _oracle_
 performance" ;-) Anyway, for all supervised source separation approaches, you want to switch of permutation. The sorting is not done framewise but globally. (@ErikJohnsonCU is this reflected in #212 ?)

carlthome:
@ErikJohnsonCU, oh, ok. Thanks! 

@faroit, hehe, alright, "evaluating oracle performance" then! :dancers: :smile:

PR #208, opened 2016-07-10
refactored file-or-str opening. fixes #191

### LINKED ISSUE #191: use context managers for open file handles
In ```io.py``` ```load_delimited``` and ```load_ragged_delimited```:

[This line](https://github.com/craffel/mir_eval/blob/master/mir_eval/io.py#L59) could leak file descriptors if there's an error anywhere below (before the call to close). Replace with ```with my_open(buf_or_str, 'r') as fdisc: ....```

bmcfee:
This PR implements #191, and is mainly lifted (and simplified) from a similar helper in JAMS.

Changes:

- A new (private) helper function `_open` consolidates repeated code for handling `string-or-file-like` inputs in IO.
- The new function uses `contextlib` to properly handle file descriptor cleanup.
- The exception type for load failure was changed from `ValueError` to `IOError`; tests were updated accordingly

### DIFF
--- a/mir_eval/io.py
+++ b/mir_eval/io.py
@@ -2,6 +2,7 @@
 Functions for loading in annotations from files in different formats.
 """
 
+import contextlib
 import numpy as np
 import re
 import warnings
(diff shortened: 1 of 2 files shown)


craffel:
Awesome, thank you.  I assume this is tried and true since it appeared in JAMS first.  The only pedantic comment I have is that `_open` is not a very descriptive function name, I would call it something like `open_file_or_str` (or `_open_file_or_str` if you insist on the pseudo-private function/not supplying a typical docstring).  Change that if you want, then feel free to squash and merge!

Issue #183, opened 2016-02-27
Create loader for tempo data


craffel:
There's currently no loader for tempo annotations in the MIREX format.  You can load them in with `io.load_delimited`, but it's cumbersome:

```Python
data = mir_eval.io.load_delimited(filename, [float, float, float])
tempi = np.array([data[0][0], data[1][0]])
weight = data[2][0]
```

It should be very simple to create this and will look a lot like `io.key`.

PR #210, opened 2016-07-21
Tempo loader

bmcfee:
implements #183


mir_eval/io.py
```diff
@@ -463,6 +464,50 @@ def load_key(filename, delimiter=r'\s+'):
...
+    denotes the relative confidence of the first value compared to the
+    second (in the range [0, 1]). The file should contain only one row.
+
+    Parameters
+    ----------
+    filename : str
+        Path to the annotation file
+    delimiter : str
+        Separator regular expression.
+        By default, lines will be split by any amount of whitespace.
```
craffel: Nitpick: "lines" is a bit of a misnomer here, since we require that there only be a single line.

craffel:
Thank you!  Made one tiny docstring nitpick, otherwise LGTM.  Feel free to fix and merge at will.

mir_eval/io.py
bmcfee: This should be tokens/columns, right?

craffel: Depends on your perspective I guess. Is it the line that's getting split into tokens/columns, or are the tokens/columns being solit out of the line or something?

bmcfee: Er, I stole that comment from the key loader... sounds like your fault :grin: 

But in general, it's the line(s) being split into columns.  I think the existing description is correct.

craffel:
Merged, thanks again.

Issue #184, opened 2016-02-28
Changelog


bmcfee:
There doesn't seem to be a changelog/release notes in the documentation.

We should definitely add this.

bmcfee:
I looked into doing this, and it seems like we don't have milestone tags for most merged PRs.  This makes it really difficult to figure out which PRs merged for which releases.

@craffel could you create retroactive milestones (0.3, 0.2, 0.1) and we can then tag old PRs to make this easy?

craffel:
> I looked into doing this, and it seems like we don't have milestone tags for most merged PRs. This e
makes it really difficult to figure out which PRs merged for which releases.
> @craffel could you create retroactive milestones (0.3, 0.2, 0.1) and we can then tag old PRs to make this easy?

There already are milestones for all of those releases: https://github.com/craffel/mir_eval/milestones?state=closed Apart from that, I don't think all contributions have been in PRs (except probably for 0.3) since lots of people have write access to the repo and we weren't doing PRs initially.  If you want to create a changelog retroactively, it probably makes more sense to just look at the list of commits (and the release notes) for each release: https://github.com/craffel/mir_eval/releases

bmcfee:
> don't think all contributions have been in PRs (except probably for 0.3) since lots of people have write access to the repo and we weren't doing PRs initially.

Can we do it by issue instead?  I really don't want to rifle through the commit log.

craffel:
> Can we do it by issue instead? I really don't want to rifle through the commit log.

Potentially, but then we might miss changes, as I don't think every change had an accompanying issue (or at least there may not be a 1-1 correspondence).  But you can already see issues assigned to each milestone:
https://github.com/craffel/mir_eval/milestone/6?closed=1
Or do you mean you want me to go back and assign each issue without a milestone to a milestone?

bmcfee:
> Or do you mean you want me to go back and assign each issue without a milestone to a milestone?

Yes.

craffel:
Ok, I will try to make a 1-1 correspondence before the next release.

bmcfee:
That might be overkill, but as long as we catch the important issues, that's fine.

I don't mind doing the sphinxification.

bmcfee:
I started the PR: i don't think we need a changelog for anything before 0.1.

In terms of issue assignment:

- 0.2 looks pretty complete
- 0.3 is totally missing
- 0.4 looks good so far.

[...]

PR #211, opened 2016-07-22
[WIP] adds changelog to documentation

bmcfee:
implements #184

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,0 +1,41 @@
+Changes
+=======
+
+v0.4
+----
+
+v0.3
+----
+
+v0.2
+----
+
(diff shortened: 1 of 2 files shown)


craffel:
What's the current status?  Anything I can help out with?

bmcfee:
I think it's up to date.

bmcfee:
... near as I can tell, the changelog covers all of the major issues for each release now.  I skipped over some of the minor ones that only affect things between versions.  For example, frame-wise bss didn't exist in 0.3, so I added an entry for the corresponding initial PR, but not for any of the subsequent patches prior to 0.4.

If that seems reasonable, I think we can merge and :shipit:

craffel:
Merged, thanks!

PR #212, opened 2016-07-26
Adding framewise evaluation for bss_eval_images

ErikJohnsonCU:
## Framewise Images Evaluation

The last of the series: A function has been added in the separation module (`bss_eval_images_framewise`) that is identical in functionality to `bss_eval_sources_framewise`, but calls `bss_eval_images` in place of `bss_eval_sources`. This is in response to discussions in #68 and would also be useful for #192.

As was done with `bss_eval_sources_framewise`, if the window and hop parameters will result in fewer than 2 windows being computed the function will simply return the result of `bss_eval_images` over the signals.

## Testing

The required unit and regression testing has been added. The unit tests reuse the functions added for `bss_eval_images` and `bss_eval_sources_framewise` with slight modifications to accommodate the differences. Regression test data has been added in the `test/data/separation/output0*.json` files.


mir_eval/separation.py
```diff
@@ -442,6 +442,101 @@ def bss_eval_images(reference_sources, estimated_sources,
...
+        return np.array([]), np.array([]), np.array([]), np.array([])
+
+    nsrc = reference_sources.shape[0]
+
+    nwin = int(
+        np.floor((reference_sources.shape[1] - window + hop) / hop)
+    )
+    # if fewer than 2 windows would be evaluated, return the images result
+    if nwin < 2:
+        return bss_eval_images(reference_sources,
```
craffel: Wouldn't this result in the function returning arrays of shape `(nsrc,)`?  It seems for consistency that we should reshape so that it returns arrays of shape `(nsrc, 1)`.  But I'm open to discussion about it!

tests/test_separation.py
```diff
@@ -151,20 +154,38 @@ def __unit_test_default_permutation(metric):
...
-                                                            est_sources,
-                                                            False),
-                       atol=A_TOL)
-    # Test with hop larger than source length
-    assert np.allclose(metric(ref_sources, est_sources, window=20, hop=120),
-                       mir_eval.separation.bss_eval_sources(ref_sources,
-                                                            est_sources,
-                                                            False),
-                       atol=A_TOL)
+    if metric == mir_eval.separation.bss_eval_sources_framewise:
```
craffel: You can collapse this test and the one below by doing something like
```Python
if metric == mir_eval.separation.bss_eval_sources_framewise:
    collapse_fn = mir_eval.separation.bss_eval_sources
elif metric == mir_eval.separation.bss_eval_images_framewise:
    collapse_fn = mir_eval.separation.bss_eval_images
else:
    raise ValueError('Unknown metric {}'.format(metric))
```
then below
```Python
assert np.allclose(metric(ref_sources, est_sources,
                          window=20, hop=120),
                   collapse_fn(ref_sources, est_sources, False),
                   atol=ATOL)
```

craffel:
Thank you again!  Of course, looks good overall; had one comment about the "not enough frames" case (which may apply to sources framewise) and another about collapsing some of your test code, but otherwise seems ready!

tests/test_separation.py
ErikJohnsonCU: That's a good point! I've made that change and added the failure on an invalid metric to all the unit tests.

mir_eval/separation.py
ErikJohnsonCU: I think that's a good idea for both the framewise functions.

faroit:
@carlthome @ErikJohnsonCU @craffel @aliutkus

To continue the discussion on #207 concerning the permutation indices: we actually think it doesn't make sense to provide a framewise permutation, on the other hand it doesn't make sense to provide an aggregated measure (like median) as well.. So we decided, to output the permutations, but we should add a smart docstring for this to make sure users are aware of this problem.

For SISEC (which is framewise images bss), permutation is not computed, so I would vote for changing the default values to `False` for all bss scores in mir_eval.

mir_eval/separation.py
```diff
@@ -442,6 +443,102 @@ def bss_eval_images(reference_sources, estimated_sources,
...
+    ...      estimated_sources,
+             window,
+    ....     hop)
+
+    Parameters
+    ----------
+    reference_sources : np.ndarray, shape=(nsrc, nsampl)
+        matrix containing true sources (must have the same shape as
+        estimated_sources)
+    estimated_sources : np.ndarray, shape=(nsrc, nsampl)
```
faroit: this should be ```np.ndarray, shape=(nsrc, nsampl, nchan)```

mir_eval/separation.py
```diff
@@ -442,6 +443,102 @@ def bss_eval_images(reference_sources, estimated_sources,
         return (sdr, isr, sir, sar, popt)
 
 
+def bss_eval_images_framewise(reference_sources, estimated_sources,
+                              window=30*44100, hop=15*44100,
+                              compute_permutation=False):
+    """Framewise computation of bss_eval_images
+
```
faroit: what about adding the matlab reference here:

```an reimplementation of http://bass-db.gforge.inria.fr/bss_eval/bss_eval_images.m by Emanuel Vincent``

File: mir_eval/separation.py (hunk 3 above)
Comment (ErikJohnsonCU) [inline mir_eval/separation.py]: True! I'll correct that.

File: mir_eval/separation.py (hunk 4 above)
Comment (ErikJohnsonCU) [inline mir_eval/separation.py]: It would make most sense to put that in the `bss_eval_images` docstring. Are we sure that link will be valid for a long time? Maybe mention that link and the paper in case the link goes belly up?

Comment (faroit) [inline mir_eval/separation.py]: hm, that might be better, yes...

it's

```
Emmanuel Vincent, Shoko Araki, Fabian J. Theis, Guido Nolte, Pau Bofill,
Hiroshi Sawada, Alexey Ozerov, B. Vikrham Gowreesunker, Dominik Lutter
and Ngoc Q.K. Duong, "The Signal Separation Evaluation Campaign
(2007-2010): Achievements and remaining challenges", Signal Processing,
92, pp. 1928-1936, 2012.
Emmanuel Vincent, Hiroshi Sawada, Pau Bofill, Shoji Makino and Justinian
P. Rosca, "First stereo audio source separation evaluation campaign:
data, algorithms and results," In Proc. Int. Conf. on Independent
Component Analysis and Blind Source Separation (ICA), pp. 552-559, 2007.
```
but maybe the 1st reference is sufficient

Comment (ErikJohnsonCU) [inline mir_eval/separation.py]: Okay, I added the first as a reference in the `bss_eval_sources` and `bss_eval_images` docstrings.

**Comment by craffel** [comment]
> So we decided, to output the permutations, but we should add a smart docstring for this to make sure users are aware of this problem.

Ok, being explicit in the docstring sounds right to me.

> For SISEC (which is framewise images bss), permutation is not computed, so I would vote for changing the default values to False for all bss scores in mir_eval.

So is the consensus that `permutation=True` is wrong in general, or just for framewise?  Is there any precedent for `permutation=True` or `permutation=False` for non-framewise anywhere else?  I hesitate to change this default because it will break backwards compatibility, but if there's consensus that it's wrong in general, that is ok.

**Comment by faroit** [comment]
> So is the consensus that permutation=True is wrong in general, or just for framewise? Is there any precedent for permutation=True or permutation=False for non-framewise anywhere else? I hesitate to change this default because it will break backwards compatibility, but if there's consensus that it's wrong in general, that is ok.

no it's not wrong in general, it's just not useful for `framewise` computation. Changing the default to `false` is just because of very few people do actually need the permutation. But yes, we should not break compatibility here, so maybe lets leave it as it is

**Comment by craffel** [comment]
> no it's not wrong in general, it's just not useful for framewise computation. Changing the default to false is just because of very few people do actually need the permutation. But yes, we should not break compatibility here, so maybe lets leave it as it is

We could also make it default `False` for framewise functions and `True` for the others, but this would be a little messy/confusing.

**Comment by carlthome** [comment]
How about computing a global SIR sorting in the framewise functions as well? That might make more sense. Or perhaps it's just an unnecessary limitation.

**Comment by faroit** [comment]
> How about computing a global SIR sorting in the framewise functions as well (e.g. concatenating the frames before determining the permutation ndarray)?

that would indeed make sense but this would also make the evaluation much slower. In fact, for very long items, consisting of many sources like in DSD100 or MedleyDB, people use framewise to make it computational efficient...

**Comment by carlthome** [comment]
For computational performance, how about doing a histogram first and using that for calculating the sorting? I.e. summing the frames to one (__or a few__) and then use that to determine the "global" permutation? Assuming the interference is evenly distributed across an estimated signal, the sum shouldn't mess that up, I think.

Basically, just something like
```Python
sorts = Enum('local fast_global true_global')
sort = ...  # Parameter
if sort == sorts.fast_global:
    frame = np.average(frames)
elif sort == sorts.true_global:
    frame = np.concatenate(frames)
# SIR sort estimated sources...
```

Or maybe this is just silly... Especially if people use framewise evaluation with all songs in DSD100 and so on, in which case separation results could be really different for songs due to different instrumentation etc.

faroit:
@carlthome, I like this idea, but I would let the users decide on this and just output framewise values for now.

ErikJohnsonCU:
> We could also make it default False for framewise functions and True for the others, but this would be a little messy/confusing.

@craffel This is what is currently being done. If the default was True and the permutation changed during the evaluation without the user being aware this would result in grossly incorrect results.

ErikJohnsonCU:
> So we decided, to output the permutations, but we should add a smart docstring for this to make sure users are aware of this problem.

@craffel @faroit @carlthome How does the following sound as a smart docstring comment?
Please be aware that this function does not compute permutations on the possible relations between reference_sources and estimated_sources due to the dangers of a changing permutation. Therefore (by default), it assumes that reference_sources[i] corresponds to estimated_sources[i]. To enable computing permutations please set compute_permutation to be True and check that the returned perm is identical for all windows.

craffel:
Sounds reasonable to me!

carlthome:
I wouldn't presume that people don't want to SIR sort per frame though. I'd just add a little to the `compute_permutation` parameter, like:

>Note: Permutations are computed independently for every frame. For `reference_sources[i]` to properly correspond to `estimated_sources[i]`, verify that the permutation result is the same for every frame.

ErikJohnsonCU:
> I wouldn't presume that people don't want to SIR sort per frame though.

But if a different permutation were computed for one frame, that would artificially buff the results by assuming a different correlation between the references and estimates than previous frames. Changing the `compute_permutation` parameter to default `True` could lead to a "blind" user (not reading the code/documentation) accepting the measures when they are incorrect.

A user can still pass `compute_permutation = True` and sort every frame, but it requires them accepting they will need to verify no change in permutation across all frames.

carlthome:
Sorry, I was unclear. I absolutely think the default value should be False. I just don't think the docstring should assume there's no use case for framewise permutations.

ErikJohnsonCU:
@carlthome I totally agree with you. There is definitely a use case for framewise permutations and the docstring should make it clear how to safely compute them.

craffel:
LGTM, thank you all again!

PR #215, opened 2016-08-07
Raise exception when piano roll has neither pitches nor midi

### LINKED ISSUE #214: mir_eval.display.piano_roll should raise a ValueError when both pitches and midi are None
As specified [here](https://github.com/craffel/mir_eval/blob/master/mir_eval/display.py#L659).  There is no such check; it just throws a confusing AtributeError [when `None` is passed to `hz_to_midi`](https://github.com/craffel/mir_eval/blob/master/mir_eval/display.py#L675).  I think this is the only place this is true in `display`.

bmcfee:
fixes #214

### DIFF
--- a/mir_eval/display.py
+++ b/mir_eval/display.py
@@ -672,6 +672,10 @@ def piano_roll(intervals, pitches=None, midi=None, ax=None, **kwargs):
     '''
 
     if midi is None:
+        if pitches is None:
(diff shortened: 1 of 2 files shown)


craffel:
I think your turnaround on the PR was faster than Travis' turnaround on the CI.  :+1 Thanks!

Issue #213, opened 2016-07-29
Framewise separation evaluation where some frames are silent


faroit:
While evaluating the framewise separation methods that have been added #199 and are almost ready in #212 , I've figured out, the validation routines are a bit too strict: Let's assume there is silence in one of the frames being evaluated:

```python
nsrc = 2
nchan = 2

empty_source = np.zeros((nsrc, 20, nchan))
estimates = np.random.random_sample((nsrc, 200, nchan))

estimated_sources = np.concatenate(
    (empty_source, estimates),
    axis=1
)

mir_eval.separation.bss_eval_images_framewise(
    estimated_sources,
    estimated_sources,
    window=10,
    hop=5
)
```

this results in 

```ValueError: All the reference sources should be non-silent (not all-zeros), but at least one of the reference sources is all 0s, which introduces ambiguity to the evaluation. (Otherwise we can add infinitely many all-zero sources.)```

and therefore would not continue evaluating the other (non-silent) frames. I think it would make sense to return `np.nan` or `None` for those (2) frames in the framewise evaluation. I don't know how to handle these results later when it comes to compute aggregated measures, but thats more of a discussion for SISEC (@aliutukus ?) 

@craffel @ErikJohnsonCU @carlthome what do you think?

carlthome:
Yes, I agree. I don't think an exception should be raised. It should continue (but maybe output the same text as a warning). It should be left to the user to think about what material they're evaluating and whether it makes sense.

craffel:
> I think it would make sense to return np.nan or None for those (2) frames in the framewise evaluation.

`np.nan` is better than `None` because a `np.ndarray` can't store a `None` without being of dtype `object`.  I will leave it up to you all to decide what the best validation/metric conventions are!

faroit:
@craffel okay, cool. Shall we wait for #212 and then add the `nan` routine in a separate PR that changes all framewise calculations?

craffel:
Sounds good to me!

ErikJohnsonCU:
This issue is corrected in PR #216. I'd invite further discussion on that webzone. I used the `np.nan` approach since it seemed that was unanimously decided to be a better approach.

faroit:
🎉 @craffel @bmcfee @carlthome @dpwe 

[...]

PR #216, opened 2016-08-08
Avoiding framewise silence issues

ErikJohnsonCU:
## Frame Silence Detection
This is to solve issue #213. A silent frame is detected by using the same check from validation before attempting to evaluate the function on the frame. This allows the silence of either the reference slice or estimate slice to be determined and then the results set as np.nan (decided as the best approach in #213).

## Unit Testing
Unit testing has been added and tests silence for both reference and estimate sources. No regression testing is required, as no values are determined. Fully silent sources will still fail the framewise evaluation since `validate` is called at the beginning of the framewise function.

#### The `if` statement
The style of the `if` statement is pretty horrid ([have a look at it here](https://github.com/faroit/mir_eval/blob/8c0ef33e1236b2661aa360ce887795974b5213b3/mir_eval/separation.py#L330)), but it passes pep8 checks for me. If anyone has advice on cleaning it up and make it easier to understand, that would be great!

**As always, all feedback is welcome!**

### DIFF
--- a/mir_eval/separation.py
+++ b/mir_eval/separation.py
@@ -109,7 +109,7 @@ def validate(reference_sources, estimated_sources):
 
 
 def _any_source_silent(sources):
-    """Returns true if the parameter sources has any silent first dimensions."""
+    """Returns true if the parameter sources has any silent first dimensions"""
     return (np.any(np.all(np.sum(sources,
                                  axis=tuple(range(2, sources.ndim))) == 0,
                           axis=1)))
--- a/mir_eval/separation.py
+++ b/mir_eval/separation.py
@@ -324,11 +324,32 @@ def bss_eval_sources_framewise(reference_sources, estimated_sources,
     # k iterates across all the windows
     for k in range(nwin):
         win_slice = slice(k * hop, k * hop + window)
-        sdr[:, k], sir[:, k], sar[:, k], perm[:, k] = bss_eval_sources(
-            reference_sources[:, win_slice],
-            estimated_sources[:, win_slice],
-            compute_permutation
-        )
+        ref_slice = reference_sources[:, win_slice]
+        est_slice = estimated_sources[:, win_slice]
+        # check for a silent frame
+        if not (np.any(np.all(np.sum(ref_slice,
+                                     axis=tuple(range(2,
+                                                      ref_slice.ndim))) == 0,
+                              axis=1))) \
+                and not (np.any(np.all(np.sum(est_slice,
+                                              axis=tuple(
+                                                    range(
+                                                        2,
+                                                        est_slice.ndim
+                                                    )
+                                                )) == 0,
+                                axis=1))):
+            sdr[:, k], sir[:, k], sar[:, k], perm[:, k] = bss_eval_sources(
+                ref_slice,
+                est_slice,
+                compute_permutation
+            )
+        else:
+            # if we have a silent frame set results as np.nan
+            sdr[:, k] = np.nan
+            sir[:, k] = np.nan
+            sar[:, k] = np.nan
+            perm[:, k] = np.nan
 
     return sdr, sir, sar, perm
 
@@ -568,12 +589,34 @@ def bss_eval_images_framewise(reference_sources, estimated_sources,
     # k iterates across all the windows
     for k in range(nwin):
         win_slice = slice(k * hop, k * hop + window)
-        sdr[:, k], isr[:, k], sir[:, k], sar[:, k], perm[:, k] = \
-            bss_eval_images(
-                reference_sources[:, win_slice, :],
-                estimated_sources[:, win_slice, :],
-                compute_permutation
-            )
+        ref_slice = reference_sources[:, win_slice, :]
+        est_slice = estimated_sources[:, win_slice, :]
+        # check for a silent frame
+        if not (np.any(np.all(np.sum(ref_slice,
+                                     axis=tuple(range(2,
+                                                      ref_slice.ndim))) == 0,
+                              axis=1))) \
+                and not (np.any(np.all(np.sum(est_slice,
+                                              axis=tuple(
+                                                    range(
+                                                        2,
+                                                        est_slice.ndim
+                                                    )
+                                                )) == 0,
+                                axis=1))):
+            sdr[:, k], isr[:, k], sir[:, k], sar[:, k], perm[:, k] = \
+                bss_eval_images(
+                    ref_slice,
+                    est_slice,
+                    compute_permutation
+                )
+        else:
+            # if we have a silent frame set results as np.nan
+            sdr[:, k] = np.nan
+            isr[:, k] = np.nan
+            sir[:, k] = np.nan
+            sar[:, k] = np.nan
+            perm[:, k] = np.nan
 
     return sdr, isr, sir, sar, perm
 
--- a/mir_eval/separation.py
+++ b/mir_eval/separation.py
@@ -337,10 +337,7 @@ def bss_eval_sources_framewise(reference_sources, estimated_sources,
             )
         else:
             # if we have a silent frame set results as np.nan
-            sdr[:, k] = np.nan
-            sir[:, k] = np.nan
-            sar[:, k] = np.nan
-            perm[:, k] = np.nan
+            sdr[:, k] = sir[:, k] = sar[:, k] = perm[:, k] = np.nan
 
     return sdr, sir, sar, perm
 
@@ -593,11 +590,7 @@ def bss_eval_images_framewise(reference_sources, estimated_sources,
                 )
         else:
             # if we have a silent frame set results as np.nan
-            sdr[:, k] = np.nan
-            isr[:, k] = np.nan
-            sir[:, k] = np.nan
-            sar[:, k] = np.nan
-            perm[:, k] = np.nan
+            sdr[:, k] = sir[:, k] = sar[:, k] = perm[:, k] = np.nan
 
     return sdr, isr, sir, sar, perm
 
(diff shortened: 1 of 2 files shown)


dpwe:
Regarding the if statement:

It looks like

np.sum(ref_slice, axis=tuple(range(2, ref_slice.ndim)))

is trying to reduce ref_slice to 2D by summing across all the dimensions
beyond the 2nd.  But as far as I can see from the function comments,
reference_sources  is always (nsrc, nsampl), so there are no further
dimensions, and ref_slice will also be 2D.

Moreover, in the case of ref_slice.ndim == 2 (which I think it does), axis
ends up set to (), which is a special case for np.sum meaning sum across
all dimensions, so that the test amounts to:

if not (np.sum(ref_slice) == 0 or np.sum(est_slice) == 0):

Am I missing something?

  DAn.

ErikJohnsonCU:
This is the general check that also works for the images case where `reference_sources` and `estimated_sources` are of form `(nsrc, nsampl, nchan)`. I used this general case in both functions since it is known to work and has been well tested.

dpwe:
On Tue, Aug 9, 2016 at 4:25 AM, Erik <notifications@github.com> wrote:

> This is the general check that also works for the images case where
> reference_sources and estimated_sources are of form (nsrc, nsampl, nchan).
> I used this general case in both functions since it is known to work and
> has been well tested.
>
That's interesting.  But if the function is supposed to work for 3D
tensors, the function comment needs updating at the very least.  I haven't
looked at the tests, but I'd want them to cover both 2D and 3D inputs if
that's what we intend.

  DAn.

ErikJohnsonCU:
@dpwe The function I believe you are referring to (`bss_eval_sources_framewise`) is not supposed to work for 3 dimension tensors. That is the other function I updated (`bss_eval_images_framewise`). I'm only using the general check for the reasons presented in my comment (known to work, well tested).

ErikJohnsonCU:
@dpwe This generalized check was introduced to `validate` in #207 after @craffel provided a good way to sum down to a 2D ndarray. This was done so that the `validate` function could work for both `bss_eval_sources` and `bss_eval_images`.

dpwe:
I think "face readability" should trump "tried and true" in this case, but
I'm just a driver-by in this instance, so please feel free to ignore my
opinions.

I was particularly concerned that np.sum(matrix_2d, axis=()) (what we get
for a 2D input) might behave as np.sum(matrix_2d, axis=None) (scalar sum of
all elements), but it doesn't. np.sum(matrix_2d, axis=()) is simply a no-op.

But if you're asking how to simplify a statement that, according to the
constraints implied by the surrounding code, has a lot of no-op machinery,
I think it's worth considering changing the idiom, e.g.:

assert ref_slice.ndim == 2 and est_slice.ndim == 2
if not (np.any(np.all(ref_slice == 0, axis=1))) and not
(np.any(np.all(est_slice == 0, axis=1))):
("if neither ref_slice nor est_slice contain any columns that are all 0")

Also, I think this is equivalent to:

if np.all(np.any(ref_slice > 0, axis=1) and np.all(np.any(est_slice > 0,
axis=1))):
("if every column of both ref_slice and est_slice have at least one value
greater than zero")

I think the gloss of the first form is actually a little clearer, but
avoiding the negation makes the second expression clearer.

Finally, if you want to retain the version that retains the capacity to
accept arbitrary sample submatrices, I would recommend folding some part of
the duplicated idiom into a helper function, e.g.

def zero_valued_column_submatrices(tensor):
  """Returns a boolean vector of size tensor.shape[1] that is true if all
values in that slice are zero."""
  return np.all(np.sum(tensor, axis=tuple(range(2, tensor.ndim))) ==
0, axis=1)

then the original code becomes:

if (not np.any(zero_valued_column_submatrices(ref_slice) and
    not np.any(zero_valued_column_submatrices(est_slice)):

  DAn.

ErikJohnsonCU:
@dpwe I do like the idea of collapsing to a helper function. That way we get to keep the tested generality and reduce a lot of complex code duplication.

mir_eval/separation.py
```diff
@@ -114,6 +108,13 @@ def validate(reference_sources, estimated_sources):
                                                 MAX_SOURCES))
 
 
+def _any_source_silent(sources):
+    """Returns true if the parameter sources has any silent first dimensions"""
+    return (np.any(np.all(np.sum(sources,
+                                 axis=tuple(range(2, sources.ndim))) == 0,
+                          axis=1)))
```
craffel: Since you asked about code style (otherwise I would not nitpick this... probably), I think it's much cleaner to do
```Python
return np.any(np.all(np.sum(
    sources, axis=tuple(range(2, sources.ndim))) == 0, axis=1))
```
Beyond that, I think it should be about-as-fast and slightly cleaner to use `np.arange` rather than `tuple(range)` in this case:
```Python
return np.any(np.all(np.sum(
    sources, axis=np.arange(2, sources.ndim)) == 0, axis=1))
```

mir_eval/separation.py
```diff
@@ -324,11 +325,22 @@ def bss_eval_sources_framewise(reference_sources, estimated_sources,
...
         win_slice = slice(k * hop, k * hop + window)
-        sdr[:, k], sir[:, k], sar[:, k], perm[:, k] = bss_eval_sources(
-            reference_sources[:, win_slice],
-            estimated_sources[:, win_slice],
-            compute_permutation
-        )
+        ref_slice = reference_sources[:, win_slice]
+        est_slice = estimated_sources[:, win_slice]
+        # check for a silent frame
+        if not _any_source_silent(ref_slice) \
```
craffel: Again a code style nitpick, but parentheses are preferred over `\` line continuation:
```Python
if (not _any_source_silent(ref_slice)
        and not _any_source_silent(est_slice)):
```

mir_eval/separation.py
```diff
@@ -568,12 +580,24 @@ def bss_eval_images_framewise(reference_sources, estimated_sources,
...
-        sdr[:, k], isr[:, k], sir[:, k], sar[:, k], perm[:, k] = \
-            bss_eval_images(
-                reference_sources[:, win_slice, :],
-                estimated_sources[:, win_slice, :],
-                compute_permutation
-            )
+        ref_slice = reference_sources[:, win_slice, :]
+        est_slice = estimated_sources[:, win_slice, :]
+        # check for a silent frame
+        if not _any_source_silent(ref_slice) \
```
craffel: Same here if you choose to change `\`.

mir_eval/separation.py
```diff
@@ -324,11 +325,22 @@ def bss_eval_sources_framewise(reference_sources, estimated_sources,
...
+                and not _any_source_silent(est_slice):
+            sdr[:, k], sir[:, k], sar[:, k], perm[:, k] = bss_eval_sources(
+                ref_slice,
+                est_slice,
+                compute_permutation
+            )
+        else:
+            # if we have a silent frame set results as np.nan
+            sdr[:, k] = np.nan
```
craffel: Code style nitpicks: You can collapse all of this into one line!
```Python
           sdr[:, k] = sir[:, k] = sar[:, k] = perm[:, k] = np.nan
```

craffel:
I like the helper function version more too.  I made some code style nitpicks (because you asked for them ;) ).  Only other question I have - is `np.nan` appropriate for `perm` with silent sources? Or should we just set it to a uniform mapping?  @faroit any input?

mir_eval/separation.py
ErikJohnsonCU: Ooooh, very true! That is much more pythonic!

mir_eval/separation.py
ErikJohnsonCU: That is a much clearer way to look at it!

mir_eval/separation.py
```diff
@@ -324,11 +324,19 @@ def bss_eval_sources_framewise(reference_sources, estimated_sources,
...
-            estimated_sources[:, win_slice],
-            compute_permutation
-        )
+        ref_slice = reference_sources[:, win_slice]
+        est_slice = estimated_sources[:, win_slice]
+        # check for a silent frame
+        if (not _any_source_silent(ref_slice) and
+                not _any_source_silent(est_slice)):
+            sdr[:, k], sir[:, k], sar[:, k], perm[:, k] = bss_eval_sources(
+                ref_slice,
```
craffel: One last style nitpick for good measure - here and below, you can just put all the arguments on a single line.

mir_eval/separation.py
```diff
@@ -114,6 +108,12 @@ def validate(reference_sources, estimated_sources):
                                                 MAX_SOURCES))
 
 
+def _any_source_silent(sources):
+    """Returns true if the parameter sources has any silent first dimensions"""
+    return np.any(np.all(np.sum(
+        sources, axis=np.arange(2, sources.ndim)) == 0, axis=1))
```
craffel: Oops, looks like `np.arange` doesn't work here when `sources.ndim == 2`, my mistake.  Kind of weird and disappointing.  Looks like you will need to go back to tuple(range).

ErikJohnsonCU: It must return something other than an empty list for the null case.. I'll change that back and make the other style correction.

dpwe: np.arange(2,2) returns [], but the axis arg to np.sum requires an int or a
tuple (not a list).
axis=tuple(np.arange(...)) will work, I think.

  DAn.

craffel: > np.arange(2,2) returns [], but the axis arg to np.sum requires an int or a
tuple (not a list).

This is what I consider weird and disappointing.  I don't really see a reason they can accept tuple but not a list (or ndarray, or other iterable).

> axis=tuple(np.arange(...)) will work, I think.

`tuple(range(...))` is better.

craffel:
Looks good to me, merging!  Thank you again for all your contributions!

PR #218, opened 2016-08-26
Speed up resample_melody_series

rabitt:
Originally ```resample_melody_series``` would resample no matter what, and this caused things to be slow when calling melody metrics many times on small segments (my use case).

Now, ```resample_melody_series``` first checks if the timebase is the same, and if it is simply returns back the original arrays rather than going through the resampling.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -12,6 +12,7 @@ v0.4
 - `#208`_: refactored file/buffer loading
 - `#210`_: added `io.load_tempo`
 - `#212`_: added frame-wise blind-source separation evaluation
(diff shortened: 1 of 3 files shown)


craffel:
Nice, thank you!  You are in luck, because doing the docstring audit for 0.4 is taking me forever, so we can still squeeze this into 0.4.  Can you

1. Add a note to the docstring which says that times and times_new are equivalent, no resampling will be performed
1. Add a note to the changelog?

Thanks!

rabitt:
Done!

craffel:
LGTM, merging!

PR #222, opened 2016-08-31
added int cast for inferred length in sonify.clicks. fixes #221

### LINKED ISSUE #221: Warning in sonify.clicks
I get this a lot:
```
/home/bmcfee/git/mir_eval/mir_eval/sonify.py:45: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
  click_signal = np.zeros(length)
```
whenever `length` is a float type.  I can cast it myself, but I think it's better to handle that within `clicks`.

bmcfee:


craffel:
TY!

PR #224, opened 2016-10-10
Sonification: Linear interpolation

stefan-balke:
Hi,

just a quick hack on the sonification module.

I had many clicks in my sonifications. Most of them stem from the discontinuities between the blocks.
To overcome this issue, we added a little ramp to the value of the next frame.

At the moment this solved my problem and I thought I'll share that. Thoughts are welcome. @justinsalamon?

### DIFF
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -131,12 +131,18 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
         # Get a waveform of length samples at this frequency
         wave = _fast_synthesize(frequency)
         # Scale each time interval by the piano roll magnitude
-        for m, (start, end) in enumerate((times * fs).astype(int)):
+        for m, (start, end) in enumerate((times[:-1, :] * fs).astype(int)):
             # Clip the timings to make sure the indices are valid
             start, end = max(start, 0), min(end, length)
+            hop_size = end - start
+            xover_size = int(hop_size / 8)
+
+            # get linear increasing weights
+            weights = np.linspace(gram[n, m], gram[n, m+1], xover_size)
 
             # Sum into the aggregate output waveform
-            output[start:end] += wave[start:end] * gram[n, m]
+            output[start:end-xover_size] += wave[start:end-xover_size] * gram[n, m]
+            output[end-xover_size:end] += wave[end-xover_size:end] * weights
 
     # Normalize
     output /= np.abs(output).max()


mir_eval/sonify.py
```diff
@@ -131,12 +131,18 @@ def _fast_synthesize(frequency):
         # Get a waveform of length samples at this frequency
         wave = _fast_synthesize(frequency)
         # Scale each time interval by the piano roll magnitude
-        for m, (start, end) in enumerate((times * fs).astype(int)):
+        for m, (start, end) in enumerate((times[:-1, :] * fs).astype(int)):
```
craffel: This means that the final column in `gram` will never be accessed, I think.  You need to special-case this after the for loop.

mir_eval/sonify.py
```diff
@@ -131,12 +131,18 @@ def _fast_synthesize(frequency):
...
             # Clip the timings to make sure the indices are valid
             start, end = max(start, 0), min(end, length)
+            hop_size = end - start
+            xover_size = int(hop_size / 8)
```
craffel: The 8 should probably be user-settable, unless we are sure this will work well in all the cases we care about.  I am concerned, for example, that if `times` had any really long-duration intervals, it would sound kind of over-the-top.

mir_eval/sonify.py
```diff
@@ -131,12 +131,18 @@ def _fast_synthesize(frequency):
...
+
+            # get linear increasing weights
+            weights = np.linspace(gram[n, m], gram[n, m+1], xover_size)
 
             # Sum into the aggregate output waveform
-            output[start:end] += wave[start:end] * gram[n, m]
+            output[start:end-xover_size] += wave[start:end-xover_size] * gram[n, m]
```
craffel: I think it would be better if the fade was zero-centered.

craffel: Also, this line is too long, which is why Travis is failing.

mir_eval/sonify.py
stefan-balke: Yes, user-settable for sure.

We could make it in milliseconds? Would be more DAW-like. This setting corresponded to ~10 ms.

bmcfee:
I'm not sure about the approach here.  The clicks are coming in because the additive synthesis is catching the synthesized wave at nonzero values.  Ramping does fix it, but it seems to me like it'll be slow, and more complex than it needs to be.

Just my $2e-2, but I think it would be both cleaner and faster to bisect the wave position back (or forward) to the closest zero.  Since we only actually synthesize a few cycles of the wave and then do in-place tiling, it's actually pretty quick to compute the location of zeros.

Given the zeros, we have code for zc-bisection in [amen](https://github.com/algorithmic-music-exploration/amen/blob/master/amen/timing.py#L45-L73), but I suspect it's simpler to rewrite it if that's how you want to go.

craffel:
> Just my $2e-2, but I think it would be both cleaner and faster to bisect the wave position back (or forward) to the closest zero. Since we only actually synthesize a few cycles of the wave and then do in-place tiling, it's actually pretty quick to compute the location of zeros.

May be a matter of taste, but I think adding zero-crossing code and then adjusting the snippets boundaries to those locations sounds more complex.  Would be happy to see both implemented and we can decide which is better.

bmcfee:
> but I think adding zero-crossing code and then adjusting the snippets boundaries to those locations sounds more complex.

But you don't have to detect the zc's.  Under the assumption that the input waveform is periodic and `f(0)=0` (eg, sine, square, triangle, but not cosine), then you can find the zeros by rounding to the nearest multiple of the period.

craffel:
> Under the assumption that the input waveform is periodic and f(0)=0

We can't make that assumption in general as long as we have the `function` argument.  We could put some kind of disclaimer that when passing your own periodic function you may get clicks, which is a little bit of a bummer.  But, I am still not convinced that adjusting-interval-to-nearest-end-of-period is cleaner/more efficient than cross-fading.  Would be happy to see an implementation of the former to compare, though to be honest I don't think there would be a strong winner either way.

bmcfee:
> We can't make that assumption in general as long as we have the function argument.

Don't we already make the periodicity assumption?



> But, I am still not convinced that adjusting-interval-to-nearest-end-of-period is cleaner/more efficient than cross-fading.

If all we do is round to the nearest period boundary, that should be negligible overhead compared to the clickypoppy implementation... but anyway, it's nbd.

craffel:
> Don't we already make the periodicity assumption?

Yes, by "that assumption" I meant f(0) = 0.

rabitt:
Adding my two cents here -- I've started using `sonify.time_frequency` pretty extensively and I hear what @stefan-balke is talking about. Namely, I hear something that sounds related to the frame rate of the TFgram. Example here:
[sonify_tf_example.wav.zip](https://github.com/craffel/mir_eval/files/1139615/sonify_tf_example.wav.zip)
Depending on the TF representation, that "hissing" is more or less prominent, but it can sometimes overpower the pitch content being sonified.

craffel:
Happy to merge this or another solution after requested changes are addressed!

craffel:
Superseded by #255

Issue #201, opened 2016-06-08
T-measure acceleration


bmcfee:
[Note: This isn't an immediate issue, but I want to jot down some thoughts here for future reference.]

The `hierarchy.tmeasure` implementation involves a [summation](https://github.com/craffel/mir_eval/blob/master/mir_eval/hierarchy.py#L146) that takes `O(n * w**2)` time, where `n` is the number of frames and `w` is the window length.  This calculation derives from counting the number of disordered triples `(q, i, j) || (q, j, i)` between estimate and reference.  We use a naive algorithm to do this via numpy ufuncs.

For small values of `w`, this is acceptably fast.  (There are other reasons to use small `w` for t-measures, as we describe in the paper, so this doesn't really come up in practice yet.)  However, when `w` is large -- or in the case of a new metric I'm developing for label hierarchies, `w==n` -- this approaches cubic time and is generally bad news.

This calculation can be reduced to `O(n w log(w))` with a divide-and-conquer algorithm for inversion counting.  For large `w`, this becomes `O(n^2 log(n))`, which is within a `log(n)` factor of the other frame-wise segmentation metrics (nce, pairwise-f, etc).

I'm implementing this enhancement in the fork that includes the label hierarchy measures.  Unless there are objections, I'll just include it with that PR, but we could also do it separately if there's sufficient need.

bmcfee:
For posterity, here's a bit more detail on how t-measures can be computed more efficiently.

After computing the least-common-ancestor matrices for `ref` and `est`, we iterate over query frames `q`.

For each `q`, we slice `ref_scores = lca_ref[q, window]` and `est_scores = lca_est[q, window]` where `window` is a slice around `q`.

Then, for each `i, j` (not counting `q`), we want to know how often the following occurs:
```
ref_scores[i] < ref_scores[j] => est_scores[i] < est_scores[j]
```
The slow algorithm for this is quadratic in the window size, since it looks at all pairs `i,j`.

This can be done more efficiently by the following method:

1. Sort `ref_scores`; let `idx` be the sorting index.
2. For each unique value `h` in `ref_scores`:
    a. Let `s, t` be the slice such that `ref_scores[s:t] == h`
    b. Find and sort the corresponding estimate scores: `pred[h] = sort(est_scores[idx][s:t])`
[...]

PR #225, opened 2016-10-12
hierarchy improvements: Accelerated T-measures and L-measures

bmcfee:
This PR combines two modifications of the segment heirarchy module:
1. As mentioned in #201 , the implementation of T-measures can be accelerated significantly by using a more clever implementation of inversion counting.
2. The T-measures can be generalized by accepting segment labels (for each layer) rather than assigning each segment a unique identifier.  This allows us to compute hierarchical segment label agreement.  The more generic version is called `L-measure` (L for label and/or lattice, since the hierarchy is no longer a tree), and is derived in an upcoming article.

So far, this PR only includes regression tests to verify that the T-measures are still numerically equivalent after the re-implementation.  I still need to expand the tests to cover L-measure regressions, and unit tests for the helper functions.


craffel:
Sweet, speedups are cool with regression tests.  For L-measure, I think we should wait until we have tests for it, or maybe even until it's published.  `mir_eval` convention is to only implement "standard" metrics for a task; we bent this rule a little for hierarchy eval, but I think having something which is not written down anywhere may bend it a bit too far.  What's the expected publication date?

bmcfee:
> convention is to only implement "standard" metrics for a task; we bent this rule a little for hierarchy eval, but I think having something which is not written down anywhere may bend it a bit too far.

Agreed -- I'm in no rush to merge, but I wanted to get the PR open so we can track regressions.

Article should be out some time in 2017 ; submission deadline is at the end of this month.

bmcfee:
Quick note: the regression tests for t-measures have slightly different values now, but this is entirely down to rounding error in the JSON output.  The new versions have one fewer digit, but are in fact identical when computed.

craffel:
Overall looks reasonable to me. The T-measure accelerations are fine as long as they pass regression tests (json rounding aside).  The L-measure, I don't know what it is/what it computes, but I assume it's correct.  Is the pub out?  In October you wrote
> Article should be out some time in 2017; submission deadline is at the end of this month.

bmcfee:
> Overall looks reasonable to me. The T-measure accelerations are fine as long as they pass regression tests (json rounding aside).

Yup, they do.  You can eyeball the diffs and see that the differences in values are only in the 15th digits.

>The L-measure, I don't know what it is/what it computes, but I assume it's correct. 

It's essentially the same as the T-measure applied to a different type of segment agreement matrix.  T-measure is actually just the special case where each segment within a level gets a unique label.

> Is the pub out? In October you wrote

Not yet. :cry: It's still under review.  I can send you the draft if you want to give it a skim.

craffel:
Ok to wait til the pub comes out?  Maybe I'm wrong but it doesn't seem like people will be rushing to use this until they have had the chance to read about it.

bmcfee:
> Ok to wait til the pub comes out? Maybe I'm wrong but it doesn't seem like people will be rushing to use this until they have had the chance to read about it.

Sure -- I just wanted to have the PR in place.  OTOH, if you're planning a release ahead of ismir, it might be nice to include since the t-measure accelerations could be useful.  It's totally your call though.

craffel:
Should be able to do another release before ISMIR.  We can aim this to put this in there.

bmcfee:
> another release before ISMIR

to be clear, I meant in advance of the deadline, so that people can use it for submissions.

craffel:
I could see that happening.

bmcfee:
Quick note on hobgoblins:

Could we at least move the pep8 check to after the test execution?  It's not helpful to have style failures precede correctness failures while developing.

craffel:
> Could we at least move the pep8 check to after the test execution? It's not helpful to have style failures precede correctness failures while developing.

Sure, either way is fine.

bmcfee:
The paper is now accepted, final print should appear soon: http://journal.frontiersin.org/article/10.3389/fpsyg.2017.01337/abstract

craffel:
Congrats!  Where does this PR stand?  Last comment you were still working on tests for the L-measure version, but that was some time ago.  I assume that the coverage increasing indicates this is ready for review at least?

bmcfee:
Yeah, it's done and ready for CR.

mir_eval/hierarchy.py
```diff
@@ -245,6 +321,105 @@ def _gauc(ref_lca, est_lca, transitive, window):
...
+        if a[i] < b[j]:
+            i += 1
+        elif a[i] >= b[j]:
+            inversions += np.sum(a_counts[i:]) * b_counts[j]
+            j += 1
+
+    return inversions
+
+
+def _compare_frame_rankings(ref, est, transitive=False):
```
craffel: Requesting simple sanity-check unit test.

mir_eval/hierarchy.py
```diff
@@ -245,6 +321,105 @@ def _gauc(ref_lca, est_lca, transitive, window):
     return score
 
 
+def _count_inversions(a, b):
```
craffel: Requesting simple sanity-check unit test.

mir_eval/hierarchy.py
```diff
@@ -143,6 +167,65 @@ def _lca(intervals_hier, frame_size):
     return lca_matrix.tocsr()
 
 
+def _meet(intervals_hier, labels_hier, frame_size):
```
bmcfee: ok, I can do that.

bmcfee: done

mir_eval/hierarchy.py
```diff
@@ -143,6 +167,65 @@ def _lca(intervals_hier, frame_size):
...
+    '''Compute the (sparse) least-common-ancestor (LCA) matrix for a
+    hierarchical segmentation.
+
+    For any pair of frames ``(s, t)``, the LCA is the deepest level in
+    the hierarchy such that ``(s, t)`` are contained within a single
+    segment at that level.
+
+    Parameters
+    ----------
+    intervals_hier : list of ndarray
```
bmcfee: done

mir_eval/hierarchy.py
bmcfee: done

bmcfee: done

mir_eval/hierarchy.py
bmcfee: done

mir_eval/hierarchy.py
```diff
@@ -143,6 +169,69 @@ def _lca(intervals_hier, frame_size):
     return lca_matrix.tocsr()
 
 
+def _meet(intervals_hier, labels_hier, frame_size):
+    '''Compute the (sparse) least-common-ancestor (LCA) matrix for a
+    hierarchical segmentation.
+
+    For any pair of frames ``(s, t)``, the LCA is the deepest level in
+    the hierarchy such that ``(s, t)`` are contained within a single
```
craffel: Maybe add a note here (or elsewhere, wherever you think is appropriate) that levels start from level 1 (as opposed to level 0).

craffel:
@urinieto can you look this over and confirm the changes match your expectations?  I don't have the expertise to sign off on the new metrics.

mir_eval/hierarchy.py
bmcfee: I don't think that matters? All the comparisons are relative, so starting at 1 or 0 won't make a difference.

bmcfee:
@urinieto ?

mir_eval/hierarchy.py
```diff
@@ -210,29 +298,22 @@ def _gauc(ref_lca, est_lca, transitive, window):
...
 
-        # Compute normalization constant
-        normalizer = float(ref_rank.sum())
+        ref_score = np.concatenate((ref_score[:idx], ref_score[idx+1:]))
+        est_score = np.concatenate((est_score[:idx], est_score[idx+1:]))
 
-        # Add up agreement for frames
-        if normalizer > 0:
-            score += np.sum(np.logical_and(ref_rank, est_rank)) / normalizer
+        # If there are no comparisons to be made, move to the next frame
```
urinieto: This comment seems out of place. I don't think that's that what `_compare_frame_rankings` does. Also, I would remove the next empty line.

mir_eval/hierarchy.py
```diff
@@ -98,6 +100,29 @@ def _hierarchy_bounds(intervals_hier):
...
+def _align_intervals(int_hier, lab_hier, t_min=0.0, t_max=None):
+    '''Align a hierarchical annotation to span a fixed start and end time.
+
+    Parameters
+    ----------
+    int_hier : list of list of intervals
+    lab_hier : list of list of str
+        Hierarchical segment annotations, encoded as a
+        list of list of intervals (int_hier) and list of
+        list of strings (lab_hier)
```
urinieto: The function arguments `t_min` and `t_max` are missing in the docstring.

mir_eval/hierarchy.py
```diff
@@ -98,6 +100,29 @@ def _hierarchy_bounds(intervals_hier):
...
+
+    Returns
+    -------
+    intervals_hier, labels_hier
```
urinieto: I would clearly specify the types of the returned values. I would also enumerate them separately.

bmcfee:
Ok, I think I've addressed Uri's comments, so this ought to be gtg.

urinieto:
Sweet, thanks! Just approved my review.

craffel:
Merged, thanks!

Issue #226, opened 2016-10-21
RFC: fixing the segment.nce metrics?


bmcfee:
[warning: opinions follow.]
### Summary

I think the normalized conditional entropy (NCE) scores for segmentation are not properly defined, and we could fix it.
### Background

The NCE scores are defined in terms of conditional entropy between the label distributions estimated from the reference and estimated segmentations.  From Lukashevich's paper:

![image](https://cloud.githubusercontent.com/assets/1190540/19609604/1e137fce-97a6-11e6-9111-dcc1a7c808c3.png)

where:
- `H(E|A)` is the entropy of the estimated label given the "annotated" (reference) label
- `N_e` is the number of unique labels in the estimated segmentation

(and similarly for `H(A|E)` and `N_a`).

The intuition here is that `log N_e` is the maximum possible entropy of any distribution over `N_e` labels (ie, the uniform distribution), so it's a good way to normalize the `S_o`.
### What's wrong

The uniform distribution has nothing to do with either the reference or estimate, and can be easily skewed when we have to pad additional segments to an estimate so that it spans the same time extent as the reference.  (This is covered in the mir_eval paper, as it explains the inflated deviation from the mirex implementation.)

A better way, I think, is to normalize by the marginal entropy `H(E)`:
![image](https://cloud.githubusercontent.com/assets/1190540/19775465/7394f3a0-9c3e-11e6-99ce-f0e5773e333e.png)

This would capture the actual decrease in uncertainty in predicting the estimated label when the reference label is provided.  Since `H(E) <= H(E|A)` for any distribution `A`, it still provides a valid normalization.

If we use the marginal entropy instead of uniform, then new labels due to tiny padding segments do not significantly change the normalization on the conditional entropy, so the results should be more stable.
### Proposed modification

Add a flag to the NCE metrics, false by default, which changes the normalization from uniform entropy to the marginal entropy.

Incidentally, this modification would render NCE equivalent to the [v-measure](http://www.cs.brandeis.edu/~marc/misc/proceedings/acl-2007/EMNLP-CoNLL2007/pdf/EMNLP-CoNLL200743.pdf), which was originally included in the segment module, but removed since it's not used by anyone in MIR or MIREX.

craffel:
As someone outside of segment eval, this seems reasonable to me.  Do you have any intuition for how much this changes the score in practice?  

> Add a flag to the NCE metrics, false by default, which changes the normalization from uniform entropy to the marginal entropy.

[...]

PR #227, opened 2016-11-01
added marginal flag to segment.nce

bmcfee:
Implements #226


bmcfee:
@craffel do you have any opinions on how tests for this one should look?  Some options:

1. Add a simple unit test that calls with a simple test case that you can check manually for `marginal=True`.  Something like `[A, B], [a, b, c, b]`.

2. Add a wrapper function `vmeasure` (or whatever we decide to call it from #226) that overrides `marginal` and calls the nce metrics.  Then add full regression test outputs as before.  I'm confident enough in the correctness of the implementation that correctness doesn't worry me.

If we do go the regression test route, I think we'd need to add the wrapper function so that the results are kept distinct from the existing NCE metrics.

What do you think?

craffel:
Regression would be good, both would be good too. Are you planning on making it so that the `evaluate` method calls the function both with `marginal` True and False? (Sorry, may have to jog my memory on this a bit)

bmcfee:
> Are you planning on making it so that the evaluate method calls the function both with marginal True and False?

Your call.  If we do, I'd advocate doing it via a wrapper/rename, instead of having two separate output dict entries.  It will make it easier to keep things separate down the road.

bmcfee:
Added bonus of the wrapper solution: we can encourage people to deprecate nce in favor of v-measure, and all they have to do is ignore columns (rather than change defaults).

craffel:
If the goal is to eventually deprecate `marginal=False`, then I agree that it doesn't make sense to compute both in `evaluate`.  But, then you need to decide whether `evaluate` will call it with `marginal=True` or `marginal=False`.

Regarding the test, whichever is the non-default behavior (i.e. whichever is NOT called in evaluate) is OK to just have a unit test.  The other should be regression tested, of course.

Does that seem reasonable?

bmcfee:
> If the goal is to eventually deprecate marginal=False, then I agree that it doesn't make sense to compute both in evaluate.

I guess it's a matter of whether we deprecate `marginal=False` and still call it NCE, or deprecate NCE entirely in favor of v-measure.  I prefer the latter, since it would reduce possibility for mis-interpretation.

craffel:
> I guess it's a matter of whether we deprecate marginal=False and still call it NCE, or deprecate NCE entirely in favor of v-measure. I prefer the latter, since it would reduce possibility for mis-interpretation.

When do you propose we do the deprecation?  What was the community concensus?

bmcfee:
> When do you propose we do the deprecation? What was the community concensus?

Predictably, there was no obvious consensus.  I think the best thing to do is implement it now, then pitch it in the unconference at the next ismir so that people have to actually respond.

craffel:
SG.  I would advocate for a unit test for v-measure and regression test for the current behavior, and then switching to regression for v-measure and unit for current behavior once it's deprecated.

bmcfee:
> I would advocate for a unit test for v-measure and regression test for the current behavior, and then switching to regression for v-measure and unit for current behavior once it's deprecated.

Cool.  OTOH, we could also just regression-test both if it's going to become another entry in the `evaluate` output.  I only propose this because it will be the minimal modification to the existing test structure, and LIS, the change is simple enough that I feel confident in the correctness without a specific unit test.

craffel:
>  OTOH, we could also just regression-test both if it's going to become another entry in the evaluate output

I thought we just decided it wasn't going to become another output of `evaluate`.

bmcfee:
> I thought we just decided it wasn't going to become another output of evaluate.

Hah, I thought we just decided the opposite?  Let me be totally clear about what I think makes the most sense:

1. Add a new function `vmeasure` that's equivalent to `nce(..., margin=True)`.  This way, we don't cause confusion between the new and old style.
2. Add `vmeasure` to the `evaluate` outputs.  Old-style NCE is still included.
3. Add regression tests for the extended dictionary.
4. Eventually, deprecate `nce` and keep `vmeasure`.  (Or, if not deprecate in the library, encourage people not to use it.)

craffel:
Ok, sounds fine.  For `evaluate`, I assume you will just make a new key in the output dict which is `v_measure` or something.

craffel:
What is left on this other than wait for the other tests to stop failing?

bmcfee:
> What is left on this other than wait for the other tests to stop failing?

Well, I'd need to actually implement the separate wrapper function and top-level `evaluate` hooks.  I've been holding off until separation rights itself.

bmcfee:
Okay, this one's up to speed.  I added `vmeasure` as a function, hooked it into `evaluate`, and updated the tests and fixtures.

I noticed that the reference section is missing in the segmentation docstring.  Do you want me to add one while we're at it?

bmcfee:
@craffel I had to move the `scipy` version requirement here because it turns out `scipy.stats.entropy` did not exist prior to 0.14.  At this point, that release is almost 3 years old, so I doubt this is a huge deal.

mir_eval/segment.py
```diff
@@ -958,6 +961,10 @@ def nce(reference_intervals, reference_labels, estimated_intervals,
         beta for F-measure
         (Default value = 1.0)
 
+    marginal : bool
+        If `False`, normalize conditional entropy by uniform entropy.
+        If `True`, normalize conditional entropy by the marginal entropy.
+
     Returns
     -------
     S_over
```
craffel: Should we change the math in this part of the docstring to note that it can be different according to the value of `marginal`?  For example denominator should be `H(y_ref)` right?  Or at least, we should change them in the `v_measure` docstring right?

craffel:
> I noticed that the reference section is missing in the segmentation docstring. Do you want me to add one while we're at it?

Sure!

craffel:
> @craffel I had to move the scipy version requirement here because it turns out scipy.stats.entropy did not exist prior to 0.14. At this point, that release is almost 3 years old, so I doubt this is a huge deal.

No problem.

mir_eval/segment.py
```diff
@@ -1009,25 +1016,99 @@ def nce(reference_intervals, reference_labels, estimated_intervals,
...
 
-    # The following scipy.stats.entropy calls are equivalent to
-    # scipy.stats.entropy(contingency, base=2)
-    # However the `base` kwarg has only been introduced in scipy 0.14.0
-    true_given_est = p_est.dot(scipy.stats.entropy(contingency) / np.log(2))
-    pred_given_ref = p_ref.dot(scipy.stats.entropy(contingency.T) / np.log(2))
+    true_given_est = p_est.dot(scipy.stats.entropy(contingency, base=2))
+    pred_given_ref = p_ref.dot(scipy.stats.entropy(contingency.T, base=2))
+
+    if marginal:
```
bmcfee: done below

bmcfee:
I think I've hit all the review points.  @craffel last pass?

craffel:
Last thing to check - did you rebuild the docs and make sure there are no new errors/glitches now that the refs are there?

bmcfee:
I have built the docs locally and didn't notice anything bad, but I wasn't looking too carefully.

craffel:
> I have built the docs locally and didn't notice anything bad, but I wasn't looking too carefully.

Look carefully at the changes caused by this PR and I will be happy :)

bmcfee:
> Look carefully at the changes caused by this PR and I will be happy :)

Done.  Fixed a couple of minor formatting issues, but I think it's solid.

craffel:
Merged, thanks!

Issue #197, opened 2016-06-01
Chroma transcription metrics


craffel:
To cover all evaluation done in MIREX, we also need to add the ability to evaluate transcription annotations after mapping the pitch values to a chroma (single octave) scale, as discussed in #180.

craffel:
Punting to 0.5.

chf2117:
I'd like to take a stab at this. Seems like the way to do this is add a flag in precision_recall_f1_overlap and match_notes?

ejhumphrey:
not sure if it'll be of any help, but there's machinery in [mir_eval.chord](https://github.com/craffel/mir_eval/blob/master/mir_eval/chord.py) that _might_ be useful??

craffel:
> I'd like to take a stab at this. 

Cool, contributions welcome.  @justinsalamon will know best what is necessary.  

> Seems like the way to do this is add a flag in precision_recall_f1_overlap and match_notes?

Seems that way to me too.

justinsalamon:
This should be relatively straight forward. The metrics (excluding ones that only consider onsets or offsets) rely on [`match_notes`](https://github.com/craffel/mir_eval/blob/master/mir_eval/transcription.py#L336), which computes pitch distances [here](https://github.com/craffel/mir_eval/blob/master/mir_eval/transcription.py#L426) and checks for pitch matches [here](https://github.com/craffel/mir_eval/blob/master/mir_eval/transcription.py#L428). You'd have to add a flag which, if set, calls an alternative version of these lines which checks for matches in (cyclic) chroma space.

chf2117:
I've got a 3 line solution that makes sense to me [here](https://github.com/chf2117/mir_eval/blob/master/mir_eval/transcription.py#L430). Still need to test.

Don't know a good way to generate test data. Any advice @justinsalamon?

justinsalamon:
@chf2117 that looks right to me (but should be tested of course). To test that it really does what you expect it to you need to add unit tests to [`test_transcrpition.py`](https://github.com/craffel/mir_eval/blob/master/tests/test_transcription.py). The toy data for the unit tests is hard coded [here](https://github.com/craffel/mir_eval/blob/master/tests/test_transcription.py#L17), if it's sufficient for covering all the chroma cases great, if not you should create more hard coded data (but don't add to / change the existing data as that will break the existing tests!).

[...]

PR #170, opened 2016-02-09
Transcription [wip]

justinsalamon:
This will (eventually) contain all the code for transcription evaluation (and tests)

PR #232, opened 2017-01-09
Chroma transcription metrics

chf2117:
Implements and tests #197

### DIFF
--- a/mir_eval/transcription.py
+++ b/mir_eval/transcription.py
@@ -430,7 +430,7 @@ def match_notes(ref_intervals, ref_pitches, est_intervals, est_pitches,
     if chroma:
         #1200 cents is one octave, mod by 1200 to remove octave information
         pitch_distances = np.mod(pitch_distances,1200)
-        pitch_distances = np.min(np.abs(pitch_distances-1200),pitch_distances)
+        pitch_distances = np.minimum(1200-pitch_distances,pitch_distances)
     pitch_hit_matrix = cmp_func(pitch_distances, pitch_tolerance)
 
     # check for offset matches if offset_ratio is not None
--- a/mir_eval/transcription.py
+++ b/mir_eval/transcription.py
@@ -430,6 +430,7 @@ def match_notes(ref_intervals, ref_pitches, est_intervals, est_pitches,
     if chroma:
         #1200 cents is one octave, mod by 1200 to remove octave information
         pitch_distances = np.mod(pitch_distances,1200)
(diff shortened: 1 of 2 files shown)


craffel:
Thanks Caleb.  Apart from PEP8 stuff, you'll also need to add separate evaluation for chroma metrics to `mir_eval.transcription.evaluate`, e.g. forcing the value of `chroma` in the `kwargs` dict to be `True`, then `False`, so that both are evaluated (right @justinsalamon)?

Do we have any implementations to test this against?  It's a small enough change that I am ok with not testing it against something existing, though we should be very certain in its correctness then :)

justinsalamon:
> Thanks Caleb. Apart from PEP8 stuff, you'll also need to add separate evaluation for chroma metrics to mir_eval.transcription.evaluate, e.g. forcing the value of chroma in the kwargs dict to be True, then False, so that both are evaluated (right @justinsalamon)?

yes

> Do we have any implementations to test this against?

yes, see [here](https://github.com/craffel/mir_eval/issues/197#issuecomment-271113346).

chf2117:
>yes, see [here.](https://github.com/craffel/mir_eval/issues/197#issuecomment-271113346)

The music-ir link is dead for me. Do you know of any others?

justinsalamon:
> The music-ir link is dead for me. Do you know of any others?

The MIREX website is temporarily down, my understanding is that they're working on it. I'll check if I have the result files stored locally somewhere, otherwise we'll have to wait till it's back up.

craffel:
Now that the MIREX website is back up, can we move forward on this?

chf2117:
Yes, I'll try to wrap it up in the next couple days

chf2117:
I'm a bit confused about comparing to MIREX results. The existing output*.json files don't match the differences from [this](https://github.com/craffel/mir_eval/pull/170#issuecomment-186446767) comment, so I'm not sure what to look for when comparing the results.

craffel:
> I'm a bit confused about comparing to MIREX results. The existing output*.json files don't match the differences from this comment, so I'm not sure what to look for when comparing the results.

@justinsalamon can you provide some guidance here?

bmcfee:
Bumping this one, since it seems like everyone disappeared after summer?

TGabor:
Is it still planned to integrate this functionality and does anybody know which points are still open?

craffel:
This needs to be picked up again. I don't think anyone is planning on doing so.

PR #234, opened 2017-02-09
Mpl2 display

### LINKED ISSUE #233: matplotlib 2.0 may break regression tests on display module
Just putting this issue here, since the new release might change some of the default settings in ways that cause the regression tests to fail.

My advice is to update the build requirements for testing to `matplotlib >= 2.0.0` and re-generate the test cases accordingly.  People can still run code on 1.5, but testing should stay up to date.

bmcfee:
This PR updates the test suite for matplotlib 2.0 compatibility, and fixes #233 .

Note: the display still works on 1.5, but for testing purposes, we now use 2.0 (and 2.0 default styles).

I had to change a couple of small things in the test setup to get things working -- notably, removing `tests/__init__.py` (tests does not need to be a module), and added an extras_require for testing.


bmcfee:
Note: this PR needs to get merged before any of the others, because display tests are currently breaking everything else.

bmcfee:
It looks like the source separation metrics are getting some minor errors now.  I can't tell if it's due to numerical precision or an actual change in the data.  Note: i didn't change any data here, only locations.

@dawenl got a minute to investigate?

bmcfee:
follow-up: I don't see any of these separation errors when i run the tests locally.  Only notable difference I see between my env and travis is that I have openblas-numpy, and travis is vanilla.

craffel:
> EDIT: confirmed that these separation failures are independent of the changes in this PR. I recommend we make a separate PR to fix the separation fixtures.

@faroit or @ecmjohnson may be the best people to ask since they touched the codebase most recently.

faroit:
@craffel I will have look this weekend

bmcfee:
@faroit @rabitt pointed out that this might be due to randomness in the bss_eval functions.

If so, this can be resolved by seeding the RNG before each test function executes.  librosa does this by a helper function [here](https://github.com/librosa/librosa/blob/master/tests/test_core.py#L37-L39), example [here](https://github.com/librosa/librosa/blob/master/tests/test_core.py#L865-L877).

bmcfee:
As mentioned over in #236, I think this one's good to go.  The only remaining failures are on separation metrics.

A quick CR would be nice, since this PR does change a little of how the tests are executed (beyond just the display submodule).  It does not change any functionality though.

.travis.yml
```diff
@@ -27,13 +27,13 @@ before_install:
...
 install:
-    - pip install -e .[display]
+    - pip install -e .[display,testing]
 
 before_script:
     - pep8 mir_eval evaluators tests
 
 script:
-    - nosetests -v --with-coverage --cover-package=mir_eval
+    - nosetests -v --with-coverage --cover-package=mir_eval -w tests
```
craffel: So the `-w tests` tells `nosetests` it's running within the `tests` folder, and this is why you removed the `tests` prefix from all the paths?

craffel:
Looked it over, looks good.  Just want to reiterate how awesome the display module is!!

.travis.yml
bmcfee: Yep.  This way you can also execute from within the `tests` directory, and the tests don't have to know that they live within the `tests/` directory of the repo.

craffel: Cool.

bmcfee:
@craffel thanks! merge away?

bmcfee:
@faroit did you have a chance to look at the source sep fixtures?

PR #236, opened 2017-02-12
forced integer division in beat pscore

### LINKED ISSUE #235: float-valued indexing will break everything on numpy >= 1.12
This warning is popping up in the current tests on python 3:
```
test_beat.test_beat_functions(<function p_score at 0x7f9f3449a8c8>,) ... /home/travis/build/craffel/mir_eval/mir_eval/beat.py:409: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
```
The [offending code](https://github.com/craffel/mir_eval/blob/master/mir_eval/beat.py#L405-L409) is due to a float-divison `/` rather than integer division `//` in computing `middle_lag`.  This is why the warning does not show up in py2.7.

This gives a warning in numpy <= 1.11, but in 1.12 and above, it will throw TypeErrors.  Note that 1.12 has not made it into conda yet, so this isn't affecting our test builds, but we should get out ahead of it.  Probably there are other points where this comes up, and we should do a thorough audit.

bmcfee:
This should fix #235 , or at least the one problem that was hitting me over in JAMS land.

### DIFF
--- a/mir_eval/beat.py
+++ b/mir_eval/beat.py
@@ -402,7 +402,7 @@ def p_score(reference_beats,
     # Get full correlation
     train_correlation = np.correlate(reference_train, estimated_train, 'full')
     # Get the middle element - note we are rounding down on purpose here
-    middle_lag = train_correlation.shape[0]/2
+    middle_lag = train_correlation.shape[0]//2
     # Truncate to only valid lags (those corresponding to the window)
     start = middle_lag - win_size
     end = middle_lag + win_size + 1


craffel:
Looks good, I think we can just merge despite the failing tests addressed in other issues, yeah?

bmcfee:
> Looks good, I think we can just merge despite the failing tests addressed in other issues, yeah?

I think this also applies to #234, if you want to take a quick skim on the error log.

PR #240, opened 2017-02-27
Repairing unit tests for source separation

### LINKED ISSUE #239: Fix source separation unit tests
Tagging @faroit @ecmjohnson 

At some point in the last six months, the regression tests for source separation have started failing, specifically on fixtures `output00` (Images Frames - Source to Interference and Artifact) and `output04` (Images - Source to Interference and Artifact).

Anyone know what's going on here?  The results are off by up to 1e-2 in some cases, so I don't think this is up to json encoding round-off error.  Maybe something changed under the hood in numpy or scipy since the fixtures were generated?  If that's all it is, then it should be a simple matter of revising the expected scores.

It looks like these fixtures were most recently updated by @ecmjohnson in #212 .

bmcfee:
This PR should resolve #239 by relaxing the pass condition on frame-wise source separation metrics.

Again, this is a result of instability introduced by solve/lstsq having different implementations depending on the underlying linear algebra system (blas/mkl/openblas/etc).


craffel:
Appears to still fail :(

bmcfee:
Dang.  I rolled the threshold up until it passed locally.  I guess travis is a different story; trying again at 1e-2.

bmcfee:
apparently 1e-2 is still too much.  Maybe it's worth removing the offending fixtures instead?

craffel:
It looks like the two remaining failing tests are still being called with atol 1e-12.

bmcfee:
Aha -- the problem is actually that some of the failures are not framewise metrics.

At this point, I recommend reverting back to strict test thresholds, but dropping examples 00 and 04 from the fixtures.  Were these generated specifically to test any particular failure modes, or are they just random clips?

faroit:
> Were these generated specifically to test any particular failure modes, or are they just random clips?
I think those (failing) scores are quite low which seem to increase the probability of numerical instabilities. However, they are not edge cases where one of the estimated targets is zero or identical to the target.

bmcfee:
@faroit so would we lose any case coverage by dropping these examples?

faroit:
short answer: no

longer answer:

04 is a nice example of one target (bass) having a really bad estimate particular in the first frames. But it's not the only item with bass separation. So it would still be okay to drop it. 

screen shot 2017-02-28 at 08 35 40

also it would be interesting to so if increasing the metric window (currently set to `120`) could average out the precision error?

bmcfee:
Okay, tests pass now that we killed the two offending cases.

craffel:
Merged, thank you sir.

faroit:
thanks for taking care of this!

PR #245, opened 2017-03-24
added labeled interval formatter

### LINKED ISSUE #244: use proper ticker locators for display.labeled_intervals
This is a minor peeve that I'll probably fix this afternoon:

In interactive mode, the `labeled_intervals` plots do not show the correct vertical axis information in the mouse coordinate display (lower-right of the display window).  This is because tick labels are set directly to the axis, and not run through a `Locator` object from matplotlib.

It should be trivial to swap the tick labeling code out for a `FixedLocator` object to fix this.  `librosa.display` does this for tonnetz plots, so we have a model to work from.

This change should have no perceivable difference in static plots.

bmcfee:
This fixes #244.  There should be no regression in the static test plots; interactive mode works as expected.  I think this is good to go.

### DIFF
--- a/mir_eval/display.py
+++ b/mir_eval/display.py
@@ -7,8 +7,9 @@ import numpy as np
 from scipy.signal import spectrogram
 
 import matplotlib.pyplot as plt
-from matplotlib.patches import Rectangle, Patch
+from matplotlib.patches import Rectangle
 from matplotlib.ticker import FuncFormatter, MultipleLocator
+from matplotlib.ticker import Formatter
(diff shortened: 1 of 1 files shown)


bmcfee:
For the record, screenshot of the new hotness in action:

image

bmcfee:
Some additional notes here:

`FixedFormatter` didn't work out of the box, so I had to write a new dummy class that interpolates positions properly.  Since this new class is purely internal, I didn't document it thoroughly.  We might want to take measures to hide it in the docs, or not, it's up to @craffel .

craffel:
Thanks!

PR #247, opened 2017-03-27
do not sonify negative amplitudes in tfgram

### LINKED ISSUE #246: sonify.chord, X should be silent but sounds like garbage
Title says it all.

Background: the chord module distinguishes two "non-chord" labels:

- `N` (no-chord) used to indicate things like silence, crowd noise, or sounds that have no harmony
- `X` (unknown) used to indicate, as far as I can tell, out-of-gamut chords.

The `X` label mainly comes up when a predictor can only operate in a reduced label space (eg, is not allowed to predict 9ths), but you still want to model those observations as distinct from `N`.

When these are encoded via [`chord.encode`](https://github.com/craffel/mir_eval/blob/master/mir_eval/chord.py#L99-L102), the `N` label maps to an empty set of pitch classes.  The `X` label maps to a vector of all `-1`, which gets sonified as button-mashing the keyboard.

Proposed change: `sonify.time_frequency` thresholds `gram` at 0 so that negative amplitudes are treated as silence.  This would solve the problem indirectly, and prevent similar mishaps from popping up elsewhere.

bmcfee:
This fixes #246 by not sonifying negative values in tfgram.

I added a simple test for `X` chord sonification, and gave it a quick spin locally.  Everything should check out.

### DIFF
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -68,6 +68,9 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
     gram : np.ndarray
         ``gram[n, m]`` is the magnitude of ``frequencies[n]``
         from ``times[m]`` to ``times[m + 1]``
+
+        Non-positive magnitudes are interpreted as silence.
+
     frequencies : np.ndarray
         array of size ``gram.shape[0]`` denoting the frequency of
         each row of gram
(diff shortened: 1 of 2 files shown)


bmcfee:
Note: this also fixed an uncaught divide-by-zero bug in `time_frequency` if the input signal was empty.  I added a numerical precision safety check before normalization of the waveform.

craffel:
Thanks.

PR #249, opened 2017-04-14
fix #248, support gaps in labeled interval interpolation

### LINKED ISSUE #248: mir_eval.util.intervals_to_samples expected behavior
What's the intended behavior for `mir_eval.util.intervals_to_samples`?

In the following example:
```python
import numpy as np
import mir_eval

intervals = np.array(
    [[0.2, 0.49],
    [0.8, 0.9]]
)
labels = ['a', 'b']
sample_times, sample_labels = mir_eval.util.intervals_to_samples(intervals, labels, offset=0, sample_size=0.1, fill_value=None)
```
I expected something close to:
```python
sample_times = np.array([0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8])
sample_labels = np.array([None, None, 'a', 'a', 'a', None, None, None, 'b'])
```
(i.e. time periods where there is no labeled interval get filled in with `None`)
but instead get:
```python
sample_times = np.array([0.0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8])
sample_labels = np.array([None, None, 'a', 'a', 'a', 'a', 'a', 'a', 'b'])
```
Am I missing something?

bmcfee:
This PR rewrites `util.interpolate_intervals` to support gaps in the intervals properly.

For now, I added documentation to indicate that `time_points` should be in ascending order.  We can either tighten this by adding a validation check, or loosen it and extend the function to support unordered time grids.

### DIFF
--- a/mir_eval/util.py
+++ b/mir_eval/util.py
@@ -145,7 +145,8 @@ def interpolate_intervals(intervals, labels, time_points, fill_value=None):
         The annotation for each interval
 
     time_points : array_like, shape=(m,)
-        Points in time to assign labels.  These must be in ascending order.
+        Points in time to assign labels.  These must be in
+        non-decreasing order.
 
     fill_value : type(labels[0])
         Object to use for the label with out-of-range time points.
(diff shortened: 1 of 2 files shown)


bmcfee:
Note: coverage percentage decrease here is because the rewrite of `interpolate_intervals` is shorter than the previous implementation.

bmcfee:
> Nice, elegant solution. Can you add a test which checks that it has the expected behavior with non-contiguous intervals?

Done.

craffel:
Thanks!

PR #250, opened 2017-05-29
Expose evaluators as command-line scripts

carlthome:
When installing mir_eval with PyPI, conda, or just a git clone it can be hard to find the evalutors/*.py scripts, particularly on CI builds and so on. This pull request exposes all the evalutors so they're available on the CLI if mir_eval has been installed. [For more, see this.](http://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point)


craffel:
Does this work?  E.g. if you `python setup.py install` in a clean environment can you then run the evaluators as binaries?

I am not opposed to this, although I don't personally appreciate when Python packages install pseudo-binaries for me without my asking.  Can we make this installation optional?  Happy to have my vote overridden here by other people's feelings/preferences.

carlthome:
Yeah, it should work. I tried with `pip install mir_eval` and could call the scrips as "binaries". It has some precedent in the MIR/Python world as well, e.g. [madmom](https://github.com/CPJKU/madmom/tree/master/bin). It also looks to me like a lot of MIREX stuff is CLI oriented.

> I don't personally appreciate when Python packages install pseudo-binaries for me without my asking.

Suppose `pip install mir_eval[cli]` could be an okay alternative (though it's not obvious to me how `extras_require` could be used for that, I'm sure it would be possible)?

bmcfee:
I think a better way to handle this is to implement `main` routines so that each submodule can run as a script.  For example,

```
python -m mir_eval.chord reference.csv estimate.csv -o scores.json
```

craffel:
> Suppose pip install mir_eval[cli] could be an okay alternative 

Yeah, something like this.  Since I don't have a strong opinion on this, I am generally favoring simplicity.

> I think a better way to handle this is to implement main routines so that each submodule can run as a script.

Hmm, this could be useful.  Would be open to this as long as we could write a utility function to call `evaluate` appropriately to avoid code duplication.

craffel:
> Suppose pip install mir_eval[cli] could be an okay alternative 

Yeah, something like this.  Since I don't have a strong opinion on this, I am generally favoring simplicity.

> I think a better way to handle this is to implement main routines so that each submodule can run as a script.

Hmm, this could be useful.  Would be open to this as long as we could write a utility function to call `evaluate` appropriately to avoid code duplication.

carlthome:
> I think a better way to handle this is to implement main routines so that each submodule can run as a script.

Wouldn't that make evaluators pretty redundant unless they were imported and run from each mir_eval submodule's main routine (which AFAIK is only possible if evaluators/ is made into its own package, or moved into mir_eval/)?

I've pushed an example with mir_eval.beat. What do you think? I prefer console_scripts even though polluting the global namespace is bad.

```sh
python -m mir_eval.beat

/home/carl/anaconda3/lib/python3.6/runpy.py:125: RuntimeWarning: 'mir_eval.beat' found in sys.modules after import of package 'mir_eval', but prior to execution of 'mir_eval.beat'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
usage: beat.py [-h] [-o OUTPUT_FILE] reference_file estimated_file
beat.py: error: the following arguments are required: reference_file, estimated_file
```

Also, what's with the RuntimeWarning?

mir_eval/beat.py
```diff
@@ -768,3 +768,8 @@ def evaluate(reference_beats, estimated_beats, **kwargs):
                                                     **kwargs)
 
     return scores
+
+
+if __name__ == '__main__':
+    from evaluators.beat_eval import main
```
carlthome: This import should be done at the top of the file but this is just an examle.

bmcfee:
> Wouldn't that make evaluators pretty redundant unless they were imported and run from each mir_eval submodule's main routine (which AFAIK is only possible if evaluators/ is made into its own package (weird), or moved into mir_eval/ (breaking change))?

Sorry, dropped the thread on this one.

What I had in mind was moving all eval logic into the mir_eval package, so that the evaluator scripts are just thin wrappers that call the corresponding submodule's evaluator, eg:
```python
# evaluators/beat_eval.py
import mir_eval.beat

if __name__ == '__main__:
    mir_eval.beat.main()
```

So yeah, they'd be redundant, but I don't think it causes any problems wrt import order or circular dependencies.  I would probably deprecate the evaluator scripts if the equivalent code gets promoted up into the package proper.

carlthome:
> What I had in mind was moving all eval logic into the mir_eval package, so that the evaluator scripts are just thin wrappers that call the corresponding submodule's evaluator.

That sounds good.

carlthome:
By the way, there's no evaluator for mir_eval/key.py.

carlthome:
I also think print_evaluation and save_evaluation should be inlined so each main() is its own thing.

carlthome:
Works on 3.5 but fails on 3.4 and 2.7 (circular import caused by tempo.py now needing io.py and io.py already importing tempo.py, I think).

carlthome:
Same problem would go for key.py if it had a CLI.

bmcfee:
> Works on 3.5 but fails on 3.4 and 2.7 (circular import caused by tempo.py now needing io.py and io.py already importing tempo.py, I think).

Ah, that's annoying.  It looks like this is because the `io` module needs to know how to validate key/tempo data.

@craffel what do you think about migrating all of the validator functions into `util`?

craffel:
> @craffel what do you think about migrating all of the validator functions into util?

This violates my mental model of how `mir_eval` is organized, but I won't block it if it's the only workable solution.

bmcfee:
> This violates my mental model of how mir_eval is organized, but I won't block it if it's the only workable solution.

Got any other ideas for how to organize it such that both task submodules and io can access validators?

craffel:
Other ideas:
- Move task-specific i/o to the submodule itself (e.g. load_tempo should be moved to tempo) but keep generic i/o (e.g. load_delimited) in io.py.  (I think this is what I am leaning towards)
- Move all validators to a validators.py file.

bmcfee:
> Move all validators to a validators.py file.

I like this idea.

craffel:
In my mental model anything specific to one particular task goes in the task submodule, shared functionality goes in util, and i/o goes in io.  So task-specific i/o into task submodules makes more sense to me but I'm not married to the idea, I'd be happy with whatever as long as there were compelling reasons.

carlthome:
> task-specific i/o into task submodules makes more sense to me but I'm not married to the idea
:+1:

carlthome:
This fixes the circular import but introduces a breaking change for users of `io.load_key` and `io.load_tempo` (now `key.load` and `tempo.load`).

Keeping an `io.load_key` around that calls `key.load` with a `DeprecationWarning` doesn't work (circular import reintroduced) so I propose just breaking it with an error and letting users fix it as they go, if that's fine with you?

Alternatively, since there's so little code to juggle, I could inline task specific validation code into `io.py` while the old functions are being deprecated (fugly) with a minor expense of some code duplication.

craffel:
> Keeping an io.load_key around that calls key.load with a DeprecationWarning doesn't work (circular import reintroduced) so I propose just breaking it with an error and letting users fix it as they go, if that's fine with you?

Yes, please leave those functions in `io` but replace their body with something like
```Python
raise NameError('mir_eval.io.load_tempo has moved to mir_eval.tempo.load_tempo')
```

carlthome:
It looks like coverage has decreased but that's actually not true. Rather, there were no tests for evaluators/*.py previously. I added basic checks that they can be started (one was broken actually), although nothing is asserted because mir_eval functionality is already tested.

Ready for code review.

craffel:
I'll review this shortly - but just a heads up that you will have to revert all of your whitespace-around-operator (e.g. `*`->` * `) changes before we merge.  These things heavily pollute the blame/diff and are no-ops in terms of both Python and PEP-8.

carlthome:
The whitespace was actually introduced by `autopep8` which has `E226 - Fix missing whitespace around arithmetic operator.` on by default, but I'll revert it if it's a preference. Maybe E226 should be added to the ignore list in a `setup.cfg`?

craffel:
Yeah, I don't think that behavior is consistent with PEP-8, 

> If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). Use your own judgment; however, never use more than one space, and always have the same amount of whitespace on both sides of a binary operator.

> Yes:
```
i = i + 1
submitted += 1
x = x*2 - 1
hypot2 = x*x + y*y
c = (a+b) * (a-b)
```

> No:
```
i=i+1
submitted +=1
x = x * 2 - 1
hypot2 = x * x + y * y
c = (a + b) * (a - b)
```

`pep8`, the tool used in Travis, is the gold standard we are using for checking PEP-8, not running through `autopep8` (though anything run through `autopep8` should pass through `pep8` without issue).

carlthome:
Btw, the `pep8` tool was renamed `pycodestyle` by [royal decree](https://github.com/PyCQA/pycodestyle/issues/466) apparently, though there seems to be slight differences because `pycodestyle mir_eval` also complains about  `mir_eval/io.py:96:17: E722 do not use bare except` which `pep8` doesn't mention.

mir_eval/io.py
```diff
@@ -581,3 +503,45 @@ def load_ragged_time_series(filename, dtype=float, delimiter=r'\s+',
             values.append(converted_value)
 
     return np.array(times), values
+
+
+def save_evaluation(results, output_file):
+    # TODO Use NumPy style docstrings like the rest of mir_eval.
```
craffel: Please fix.

craffel: ...also, I think it would be ok to not functionify this.  It's short enough.

mir_eval/io.py
```diff
@@ -581,3 +503,45 @@ def load_ragged_time_series(filename, dtype=float, delimiter=r'\s+',
...
+            the corresponding scores
+        - output_file : str
+            Path to .json file to write to
+    '''
+    with open(output_file, 'w') as f:
+        json.dump(results, f)
+
+
+def print_evaluation(results):
+    # TODO Use NumPy style docstrings like the rest of mir_eval.
```
craffel: Please fix.

mir_eval/io.py
```diff
@@ -581,3 +503,45 @@ def load_ragged_time_series(filename, dtype=float, delimiter=r'\s+',
...
+    max_len = max([len(key) for key in results])
+    for key, value in results.items():
+        if type(value) == float:
+            print('\t{:>{}} : {:.3f}'.format(key, max_len, value))
+        else:
+            print('\t{:>{}} : {}'.format(key, max_len, value))
+
+
+def load_tempo(filename, delimiter=r'\s+'):
+    raise NameError("load_tempo moved to mir_eval.tempo.load")
```
craffel: Keep name load_tempo.

mir_eval/io.py
```diff
@@ -581,3 +503,45 @@ def load_ragged_time_series(filename, dtype=float, delimiter=r'\s+',
...
+
+
+def load_key(filename, delimiter=r'\s+'):
+    raise NameError("load_key moved to mir_eval.key.load")
```
craffel: Keep name load_key.

mir_eval/key.py
```diff
@@ -61,6 +66,43 @@ def validate(reference_key, estimated_key):
         validate_key(key)
 
 
+def load(filename, delimiter=r'\s+'):
```
craffel: Keep name load_key.

mir_eval/io.py
```diff
@@ -581,3 +503,45 @@ def load_ragged_time_series(filename, dtype=float, delimiter=r'\s+',
...
+            Results dictionary, where keys are metric names and values are
+            the corresponding scores
+        - output_file : str
+            Path to .json file to write to
+    '''
+    with open(output_file, 'w') as f:
+        json.dump(results, f)
+
+
+def print_evaluation(results):
```
craffel: This function should go in util, not io.  It doesn't involve I/O.

mir_eval/tempo.py
```diff
@@ -65,6 +71,49 @@ def validate(reference_tempi, reference_weight, estimated_tempi):
         raise ValueError('Reference weight must lie in range [0, 1]')
 
 
+def load(filename, delimiter=r'\s+'):
```
craffel: Keep name load_tempo

mir_eval/io.py
carlthome: Arguable IMO as `sys.stdout` is treated as an always open file object in Python, no? But more importantly, it should be possible to remove the function entirely (maybe pandas instead?).

mir_eval/io.py
carlthome: Done

mir_eval/io.py
carlthome: Ye, I'm a bit OOP damaged, sry.

mir_eval/io.py
carlthome: Done

carlthome:
Fixed comments and everything should be green now but a seemingly unrelated error happens with Python 2.7  :confused:

```
ERROR: Failure: ImportError (cannot import name KnownFailureTest)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/travis/build/craffel/mir_eval/tests/test_display.py", line 14, in <module>
    from mpl_ic import image_comparison
  File "/home/travis/build/craffel/mir_eval/tests/mpl_ic.py", line 26, in <module>
    from matplotlib.testing.noseclasses import KnownFailureTest, \
ImportError: cannot import name KnownFailureTest
```
@craffel thoughts?

carlthome:
"matplotlib.testing.noseclasses is deprecated and will be removed in 2.3"

carlthome:
Though it looks like matplotlib==2.1.0 is used in the failing build unless I'm reading the log wrong.

carlthome:
Reopened to get a new build just to see where we're at. :smile:

carlthome:
Closing because stale.

Issue #251, opened 2017-07-04
Extended chords not mapped correctly?


fdlm:
- Does `mir_eval` handle extended chords correctly? It seems to ignore everything beyond the seventh:
```
In [5]: mir_eval.chord.encode('A:maj(9)')
Out[5]: (9, array([1, 0, 0, 0, 1, 0, 0, 1, 0, 0, 0, 0]), 0)

In [6]: mir_eval.chord.encode('A:maj(9)', reduce_extended_chords=True)
Out[6]: (9, array([1, 0, 0, 0, 1, 0, 0, 1, 0, 0, 0, 0]), 0)
```
compared to
```
In [8]: mir_eval.chord.encode('A:maj(2)')
Out[8]: (9, array([1, 0, 1, 0, 1, 0, 0, 1, 0, 0, 0, 0]), 0)
```
- Does it make sense to treat chords such as `A:maj(9)` and `A:maj(2)` differently in the first place? Both comprise the same pitch classes, and I doubt that one can distinguish between the two in the sound mixture of e.g. a pop/rock song.

ejhumphrey:
That second line (`In [6]`) is the bug, and should produce the same as `In [8]`. I've confirmed locally, and I'm a little surprised that bug got so far ... but there definitely aren't any tests for that kind of behavior. 😞  

I'll try to patch soon, and will tag you in the PR.

ejhumphrey:
oh, and to your second question ... I don't think `mir_eval` should take a philosophical stance on the difference between an add2 and a 9 chord, and the flag `reduce_extended_chords` lets the user do whichever they need. That said, since the second returned object is a bit-vector of active pitch classes, perhaps `reduce_extended_chords=True` is a more appropriate default. 

That decision, however, is going to need some input from at least @craffel.

ejhumphrey:
also also thanks for raising the issue 😄

bmcfee:
> I don't think mir_eval should take a philosophical stance on the difference between an add2 and a 9 chord,

I'll chime in and disagree here (of course :grin:).  I think it would be useful to extend the chord encoding vector to a double-octave representation, so that above-octave extensions can be coded unambiguously (and distinct from, say, add2).  Of course, we could retain the default behavior by collapsing out the second octave, but having the double-octave intermediate form might be useful down the line.

fdlm:
[...]

PR #252, opened 2017-07-05
Ejh 20170704 extended chord mapping

ejhumphrey:
Resolves issue #251.

**Diagnosis:** The method `scale_degree_to_bitmap` in `mir_eval.chord` was only ever dropping extended scale degrees on the floor, regardless of whether or not `reduce_extended_chords` was set. This resulted in unexpected behavior when `reduce_extended_chords=True`, because one expects that the extended scale degrees would be folded into the bit-vector representation.

**Fix:** Two arguments were added to `scale_degree_to_bitmap`: `modulo` and `length`, defaulting to `False` and `12` respectively, which preserves the same behavior of `mir_eval` overall. Then, in the `encode` method, `reduce_extended_chords` is passed on to `modulo`. This also preserves the consistent behavior of `mir_eval`, while fixing the expected behavior of these methods.

This will be of interest to @fdlm and @bmcfee, having weighed in on the issue.

### DIFF
--- a/mir_eval/chord.py
+++ b/mir_eval/chord.py
@@ -414,6 +414,7 @@ def split(chord_label, reduce_extended_chords=False):
 
     if reduce_extended_chords:
         quality, addl_scale_degrees = reduce_extended_quality(quality)
+        print(quality, addl_scale_degrees)
         scale_degrees.update(addl_scale_degrees)
 
     return [chord_root, quality, scale_degrees, bass]
--- a/mir_eval/chord.py
+++ b/mir_eval/chord.py
@@ -177,6 +177,9 @@ def scale_degree_to_semitone(scale_degree):
     semitone : int
         Relative semitone of the scale degree, wrapped to a single octave
 
+    Raises
+    ------
+    InvalidChordException if `scale_degree` is invalid.
     """
     semitone = 0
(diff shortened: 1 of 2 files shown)


ejhumphrey:
for convenience, I believe [this line](https://github.com/craffel/mir_eval/pull/252/files#diff-d39185c0311a57dc7670155643e8893aR202) is that check.

fdlm:
Yes, works for me!

mir_eval/chord.py
```diff
@@ -204,21 +208,25 @@ def scale_degree_to_bitmap(scale_degree):
     ----------
     scale_degree : str
         Spelling of a relative scale degree, e.g. 'b3', '7', '#5'
+    modulo : bool, default=True
+        If a scale degree exceeds the length of the bit-vector, modulo the
+        scale degree back into the bit-vector; otherwise it is discarded.
+    length : int, default=12
+        Length of the bit-vector to produce
```
bmcfee: should be positive int

bmcfee: also, does it make sense to support anything other than multiples of 12?

craffel:
Thanks @ejhumphrey !

PR #253, opened 2017-07-06
Ival to sample gaps

rabitt:
It looks like this PR never actually got opened -- I ran into the same issue again and would love the gap functionality :)


rabitt:
Derp -- this got merged jk. Closing!

PR #255, opened 2017-07-17
Fix crackle in sonify.time_frequency

rabitt:
This addresses @craffel's comments on #224.

The main changes are:
1) Replaces @stefan-balke's asymmetric ramp between frames with an interpolator so the amplitudes will change smoothly (and symmetrically) over time.
2) Fixes a second cause of noisiness/buzzing I found in `_fast_synthesize`. The line:
`n_samples = int(10.0 * fs / frequency)`
is meant to generate 10 periods of the wave at `f0=frequency`, and the generated wave is copied. For non-integer values of `frequency`, `10. * fs / frequency` is not necessarily an integer, so casting it to an integer results in generating less than 10 periods of the wave, and then there is a discontinuity when  the wave is copied. I did a hack to fix this problem and force `n_samples` to always be an integer.

If this is merged, #224 can be closed. cc @stefan-balke

### DIFF
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -60,7 +60,8 @@ def clicks(times, fs, click=None, length=None):
     return click_signal
 
 
-def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
+def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None,
+                   n_dec=1):
     """Reverse synthesis of a time-frequency representation of a signal
 
     Parameters
@@ -84,6 +85,9 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
     length : int
         desired number of samples in the output signal,
         defaults to ``times[-1]*fs``
+    n_dec : int
+        the number of decimals used to approximate each sonfied frequency.
+        Defaults to 1 decimal place. Higher precision will be slower.
 
     Returns
     -------
@@ -102,17 +106,24 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
     times, _ = util.adjust_intervals(times, t_max=length)
 
     # Truncate times so that the shape matches gram
-    times = times[:gram.shape[1]]
+    n_times = gram.shape[1]
+    times = times[:n_times]
 
     def _fast_synthesize(frequency):
         """A faster way to synthesize a signal.
             Generate one cycle, and simulate arbitrary repetitions
             using array indexing tricks.
         """
-        frequency = float(frequency)
+        # hack so that we can ensure an integer number of periods and samples
+        # rounds frequency to 1st decimal, s.t. 10 * frequency will be an int
+        frequency = np.round(frequency, n_dec)
 
-        # Generate ten periods at this frequency
-        n_samples = int(10.0 * fs / frequency)
+        # Generate 10*frequency periods at this frequency
+        # Equivalent to n_samples = int(n_periods * fs / frequency)
+        # n_periods = 10*frequency is the smallest integer that guarantees
+        # that n_samples will be an integer, since assuming 10*frequency
+        # is an integer
+        n_samples = int(10.0**n_dec * fs)
 
         short_signal = function(2.0 * np.pi * np.arange(n_samples) *
                                 frequency / fs)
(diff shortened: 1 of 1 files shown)


rabitt:
@craffel I think this is good to go (on my end).

mir_eval/sonify.py
```diff
@@ -102,17 +102,24 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None):
...
 
     def _fast_synthesize(frequency):
         """A faster way to synthesize a signal.
             Generate one cycle, and simulate arbitrary repetitions
             using array indexing tricks.
         """
-        frequency = float(frequency)
+        # hack so that we can ensure an integer number of periods and samples
+        # rounds frequency to 1st decimal, s.t. 10 * frequency will be an int
+        frequency = np.round(frequency, 1)
```
bmcfee: Doesn't this mean that only integer frequencies are sonifiable?

rabitt: No -- but i'm rounding each frequency to the nearest 10s place, so it might have an audible effect on higher frequencies.

bmcfee: offline conversation summary: might be better to make the number of decimals a parameter `n_dec`, and maintain a table of size `10**n_dec * fs` so that a user can control the precision of the approximation.

mir_eval/sonify.py
```diff
@@ -128,21 +139,41 @@ def _fast_synthesize(frequency):
...
+            gram_interpolator = _const_interpolator(gram[n, 0])
+
         # Scale each time interval by the piano roll magnitude
         for m, (start, end) in enumerate((times * fs).astype(int)):
             # Clip the timings to make sure the indices are valid
             start, end = max(start, 0), min(end, length)
-
-            # Sum into the aggregate output waveform
-            output[start:end] += wave[start:end] * gram[n, m]
+            weights = gram_interpolator(np.arange(start, end))
```
craffel: Looks like `weights` is unused?  Did you mean to replace `gram_interpolator(np.arange(start, end))` with `weights` below?

mir_eval/sonify.py
```diff
@@ -128,21 +139,41 @@ def _fast_synthesize(frequency):
...
         # Get a waveform of length samples at this frequency
         wave = _fast_synthesize(frequency)
+        if len(time_centers) > 1:
+            gram_interpolator = interp1d(
+                time_centers, gram[n, :], kind='linear',
+                bounds_error=False, fill_value=0.0)
+        else:
+            # if there is only one point in time_centers, interpolator
+            # returns constant
+            gram_interpolator = _const_interpolator(gram[n, 0])
```
craffel: Can you replace all this logic with 

```Python
gram_interpolator = interp1d(
    time_centers, gram[n, :],
    kind='linear' if len(time_centers) > 1 else 'nearest',
    bounds_error=False, fill_value=0.0)
```
or maybe `'zero'` instead of `'nearest'`?  Alternatively if you can provide a little more of a comment for special casing, would be helpful for understanding it.

rabitt:
@craffel I ran into a snag in the interpolation when `len(time_centers)=1`. 

Using `kind='nearest'`:
```
gram_interpolator = interp1d(
    time_centers, gram[n, :],
    kind='linear' if len(time_centers) > 1 else 'nearest',
    bounds_error=False, fill_value=0.0)
```
it fails because when `kind='nearest'` it expects `time_centers` and `gram[n, :]` to have more than 1 point.

Using `kind='zero'`:
```
gram_interpolator = interp1d(
    time_centers, gram[n, :],
    kind='linear' if len(time_centers) > 1 else 'zero',
    bounds_error=False, fill_value=0.0)
```
The interpolator always returns zero -- for example:
```
f = interp1d([1], [3], kind='zero', bounds_error=False, fill_value=0.0)
f(-1)
>> array(0.0)
f(0)
>> array(0.0)
f(1)
>> array(2.1904428788e-314)
f(2)
>> array(0.0)
```

What do you want the expected behavior to be? For a length `n` output, do you want a the same set of pitches to be sustained, or something else?

craffel:
Hmm, those are frustrating behaviors of interp1d.  Oh well, you can go back to what you had!

craffel:
Thanks @rabitt !

Issue #257, opened 2017-08-01
Improve efficiency for bipartite matching?


bmcfee:
Right now we rely on an [all-pairs distance calculation](https://github.com/craffel/mir_eval/blob/master/mir_eval/util.py#L718) between reference and estimate events.  If you have a large set on either end (say, 10^5), this is not a great idea.  [Tagging @lostanlen for bringing this up.]

Since these are all comparing 1-dimensional objects, we could accelerate this from `O(n*m)` (for `n` reference and `m` estimates) to `O(m * log(n))` by sorting both arrays and doing bisection-search (`np.searchsorted`) for each event.  For each estimate event `t`, we can bisect-left the reference for `t-window` and bisect-right for `t+window` to get a slice `u:v`, constituting all hits.  This can be done quickly in numpy as follows:

```python
ref_sort = np.sort(refs)
left_hits = np.searchsorted(ref_sort, est - window, side='left')
right_hits = np.searchsorted(ref_sort, est + window, side='right')

# hits for est[i] are ref_sort[left_hits[i]:right_hits[i]]
```

bmcfee:
Looked into this more closely; supporting alternative distance functions is a bit of a pain.  I think for now the best approach is to modify the signature so that `distance=None` implies `|x-y|`, for which we can do the fast bisection search.  Any non-None metric supplied (eg mod-n distance used in multipitch) will revert to the current outer-distance method.

PR #258, opened 2017-08-01
Faster matching of events

bmcfee:
Ths PR addresses #257 by doing a sorted bisection search for finding matched events, rather than building an all-pairs distance matrix.  The latter is still supported for fancy distance metrics, as is used in multipitch eval.

Quick benchmark:
```python

In [5]: est = np.random.randn(1000)

In [6]: ref = np.random.randn(10000) * 100

In [7]: %timeit mir_eval.util.match_events(ref, est, window=0.5)
57.4 ms ± 381 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: %timeit mir_eval.util.match_events(ref, est, window=0.5, distance=mir_eval.util._outer_distance)
183 ms ± 1.52 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [9]: est = np.random.randn(10000)

In [10]: %timeit mir_eval.util.match_events(ref, est, window=0.5, distance=mir_eval.util._outer_distance)
1.88 s ± 46.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [11]: %timeit mir_eval.util.match_events(ref, est, window=0.5)
536 ms ± 5.13 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/mir_eval/util.py
+++ b/mir_eval/util.py
@@ -680,7 +680,7 @@ def _outer_distance(ref, est):
     return np.abs(np.subtract.outer(ref, est))
 
 
-def match_events(ref, est, window, distance=_outer_distance):
+def match_events(ref, est, window, distance=None):
     """Compute a maximum matching between reference and estimated event times,
     subject to a window constraint.
 
@@ -702,7 +702,7 @@ def match_events(ref, est, window, distance=_outer_distance):
         Size of the window.
     distance : function
         function that computes the outer distance of ref and est.
-        By default uses _outer_distance, ``|ref[i] - est[j]|``
+        By default uses ``|ref[i] - est[j]|``
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


lostanlen:
Looks good to me. Thank you for putting the time into this.

bmcfee:
I put one more minor optimization in there, removing the inner-loop from `fast_hit_windows`.  I think this is now as fast as can be, and ready for CR.

craffel:
Looks good, as long as all the regression tests pass :)

Can we remove `_outer_distance`?  Or just deprecate it?

bmcfee:
> Can we remove _outer_distance? Or just deprecate it?

It's a private(ish) function that isn't referenced outside of util, so I'm cool with removing it.  Do you want me to put that in this PR as well?

I think it might be worth putting some thought into whether a similar trick could be done with `_outer_distance_modn` so we can get rid of all the quadratic-time calculations.  I don't immediately see an easy way to do it, but I also haven't thought about it too carefully.  Either way, that can be handled as a separate issue.

craffel:
> It's a private(ish) function that isn't referenced outside of util, so I'm cool with removing it. Do you want me to put that in this PR as well?

Sure.

> I think it might be worth putting some thought into whether a similar trick could be done with _outer_distance_modn so we can get rid of all the quadratic-time calculations. I don't immediately see an easy way to do it, but I also haven't thought about it too carefully. Either way, that can be handled as a separate issue.

Sounds good.

craffel:
Thanks for this!

PR #261, opened 2017-08-10
added fix for comma instead of dot in floating point numbers

georgid:
When there is a timestamp stored in a text file with comma instead of dot for dividing the whole number from the mantissa (e.g. 1,23 instead of 1.23), [load_delimited](https://github.com/craffel/mir_eval/blob/master/mir_eval/io.py#L93) throws an error.

### DIFF
--- a/mir_eval/io.py
+++ b/mir_eval/io.py
@@ -90,6 +90,7 @@ def load_delimited(filename, converters, delimiter=r'\s+'):
             for value, column, converter in zip(data, columns, converters):
                 # Try converting the value, throw a helpful error on failure
                 try:
+                    value = value.replace(',' , '.')
                     converted_value = converter(value)
                 except:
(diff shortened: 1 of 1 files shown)


craffel:
Is this common e.g. in Europe?  Is `mir_eval` Americentric?  I would otherwise argue that `load_delimited` isn't meant to do any actual conversion of the values - they should be ready-to-go for `int()` or `float()` or whatever.  In other words, I would say if this is a bug it's a bug with `float()`.  You could also fix this by passing in an appropriate "converter" for the col in question, e.g. `lambda val: float(val.replace(',', '.')`.

georgid:
The "converter" fix with the converter is absolutely solving the issue. Otherwise, I dont think that ',' instead of '.' is Europe-related situation. It is very likely to appear when values are exported from excel or other postprocessing editors that use ','-s.

PR #263, opened 2017-10-04
Measures for over- and under-segmentation of chord labels

fdlm:
An attempt for addressing #262. 

 - The MIREX `MeanSeg` metric is not implemented, because it is the dataset average of min(overseg, underseg). I don't think it's meaningful to add another field to the scores that just returns the minimum of the two.
 - The results can be slightly different than the ones used in MIREX due to a philosophical difference. This implementation uses the *adjusted* intervals (i.e., assumes a "N" detection if the estimation does not output anything), while in MIREX, positions without an explicit estimation are ignored.

Let me know what you think.


mir_eval/chord.py
```diff
@@ -1342,6 +1347,134 @@ def sevenths_inv(reference_labels, estimated_labels):
...
+    -------
+    directional hamming distance : float
+        directional hamming distance between reference intervals and
+        estimated intervals.
+    """
+    est_ts = np.unique(estimated_intervals.flatten())
+    seg = 0.
+    for start, end in reference_intervals:
+        dur = end - start
+        between_start_end = est_ts[(est_ts > start) & (est_ts < end)]
```
bmcfee: Should only one of these comparisons be strict?

mir_eval/chord.py
```diff
@@ -1342,6 +1347,134 @@ def sevenths_inv(reference_labels, estimated_labels):
...
+    estimated_intervals : np.ndarray, shape=(m, 2), dtype=float
+        Estimated chord intervals to score against.
+
+    Returns
+    -------
+    directional hamming distance : float
+        directional hamming distance between reference intervals and
+        estimated intervals.
+    """
+    est_ts = np.unique(estimated_intervals.flatten())
```
bmcfee: Seems like this should have an input validation

fdlm: Something like `if estimated_intervals.shape[1] != 2` or should the check go deeper and ensure e.g. that there are no overlaps in the intervals, etc.?

fdlm:
Right, I missed that - somehow I thought having numbers in the `outputXX.json` is enough, but I will write tests that cover some edge cases.

mir_eval/chord.py
bmcfee: Calling `util.validate_intervals` would probably suffice.  It doesn't check for disjoint/complete segmentation (and maybe it should), but it does cover the basics of shape matching and valid interval timings.

craffel:
Thanks for reviewing @bmcfee , I am slammed right now so I can give it a final look-over once things are squared

fdlm:
So, I think I addressed all concerns @bmcfee pointed out, so from my side, this PR is ready to merge. Let me know if there is anything else I can improve!

mir_eval/chord.py
```diff
@@ -1342,6 +1347,136 @@ def sevenths_inv(reference_labels, estimated_labels):
...
+    util.validate_intervals(estimated_intervals)
+    util.validate_intervals(reference_intervals)
+    est_ts = np.unique(estimated_intervals.flatten())
+    seg = 0.
+    for start, end in reference_intervals:
+        dur = end - start
+        between_start_end = est_ts[(est_ts >= start) & (est_ts < end)]
+        seg_ts = np.hstack([start, between_start_end, end])
+        seg += dur - np.diff(seg_ts).max()
+    return seg / (reference_intervals[-1, 1] - reference_intervals[0, 0])
```
bmcfee: possible div-by-0 here, since `validate_intervals` doesn't check for non-emptiness.  Maybe add an explicit check here, and document the behavior if reference intervals are empty.

fdlm: I think it does implicitly in the following check (util.py:770):
```python
    if intervals.ndim != 2 or intervals.shape[1] != 2:
        raise ValueError('Intervals should be n-by-2 numpy ndarray, '
                         'but shape={}'.format(intervals.shape))
```
Here, `np.atleast_2d(np.array([])).shape[1] == 0`, and thus the `ValueError` will be raised. So, at least one interval must be present, and it must have a positive duration, and thus `reference_intervals[-1, 1] - reference_intervals[0, 0]` must be positive.

bmcfee: Ah, quite correct, thanks.  Sorry I missed that before.

LGTM!

craffel:
Great.  I am about to travel to ISMIR but should get a chance to do a final pass over in the next few days.  Thanks!

mir_eval/chord.py
```diff
@@ -1342,6 +1347,136 @@ def sevenths_inv(reference_labels, estimated_labels):
...
+    --------
+    >>> (ref_intervals,
+    ...  ref_labels) = mir_eval.io.load_labeled_intervals('ref.lab')
+    >>> (est_intervals,
+    ...  est_labels) = mir_eval.io.load_labeled_intervals('est.lab')
+    >>> overseg = 1 - mir_eval.chord.directional_hamming_distance(
+    ...     ref_intervals, est_intervals)
+    >>> underseg = 1 - mir_eval.chord.directional_hamming_distance(
+    ...     est_intervals, ref_intervals)
+    >>> meanseg = min(overseg, underseg)
```
craffel: I'm confused by this line --
- It's called "meanseg" but is computed as the min.
- Is this a thing people measure?  If so should we have a separate metric function for it?

mir_eval/chord.py
```diff
@@ -1342,6 +1347,136 @@ def sevenths_inv(reference_labels, estimated_labels):
     return comparison_scores
 
 
+def directional_hamming_distance(reference_intervals, estimated_intervals):
+    """Compute the directional hamming distance between reference and
+    estimated intervals as defined by [#harte2010towards]_ and used for MIREX
+    'OverSeg', 'UnderSeg' and 'MeanSeg' measures.
+
+    Examples
```
craffel: _Technically_, since this isn't a metric function, you don't need an "Examples" section, but it certainly doesn't hurt :)

mir_eval/chord.py
fdlm: I agree that the naming is weird, but I wanted to be consistent with MIREX (see http://www.music-ir.org/mirex/wiki/2017:Audio_Chord_Estimation_Results).

I'm not sure if it is meaningful to have this as a separate metric for this for each song (it is, after all, just min(overseg, underseg)). However, it might then be easier to recreate the metrics as used in the MIREX task, where "MeanSeg" means the mean of the min(os, us) for each song. In this case, I would prefer to call it just something like "Segmentation".

What do you think?

craffel: So they are calling it "mean" because it's the mean of the min across all songs?  I guess that bothers me because all of the metrics are mean across songs, correct?  In terms of what I think, I honestly don't know in this case!  But I will say typically it's ok to have an extra metric function if it's something that people measure, even if it's oneline.  The idea being that evaluate returns everything that you might want to measure to report your results.

fdlm: Sure, let's add a function for it, then! The only negative now is that `underseg()` and `overseg()` get computed twice when calling `evaluate()`. If this is a problem, I can replace `scores['seg'] = seg(...)` with `scores['seg'] = min(scores['underseg'], scores['overseg'])`.

craffel: > If this is a problem, I can replace scores['seg'] = seg(...) with scores['seg'] = min(scores['underseg'], scores['overseg']).

This seems like the best option. (and sorry for the delay)

fdlm: No worries! Should be done now. If there is anything I can do to further improve this PR, let me know.

craffel:
Merged!  Thanks so much!

Issue #274, opened 2018-05-04
chord.eval: expose extension-reduction parameters


bmcfee:
(tagging @tomxi @ejhumphrey) 

The first step taken in the chord metrics is to map chord strings onto pitch/root/bass encodings, for example, [here](https://github.com/craffel/mir_eval/blob/master/mir_eval/chord.py#L747-L748).  Currently, this hard-codes the `reduce_extended_chords` parameter to `False` (second argument).  This maintains compatibility with mirex by default, but there is no way for a user to override this behavior.  This is problematic for examples like the following (due to @tomxi):

```python
In [1]: import mir_eval

In [2]: c1 = 'C#:7(b9,#9)'

In [3]: c2 = 'C#:min7(b9,b11)'

In [4]: mir_eval.chord.encode(c1)
Out[4]: (1, array([1, 0, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0]), 0)

In [5]: mir_eval.chord.encode(c2)
Out[5]: (1, array([1, 0, 0, 1, 0, 0, 0, 1, 0, 0, 1, 0]), 0)

In [6]: mir_eval.chord.encode(c1, reduce_extended_chords=True)
Out[6]: (1, array([1, 1, 0, 1, 1, 0, 0, 1, 0, 0, 1, 0]), 0)

In [7]: mir_eval.chord.encode(c2, reduce_extended_chords=True)
Out[7]: (1, array([1, 1, 0, 1, 1, 0, 0, 1, 0, 0, 1, 0]), 0)
```
where the chords `c1` and `c2` have identical pitch class content if extensions are included (last two encodings) but differ within the octave (first two encodings).

What do folks think about exposing the reduction parameter at the API level, so that users can decide whether or not to include extensions?  @ejhumphrey do you anticipate any backfiring here?

craffel:
Assuming there is a use-case for evaluation using `reduce_extended_chords=True`, I am in support of this.  I don't know enough about chords to know whether this is a common or possible use-case.

bmcfee:
It's not common (it's not even possible right now :grin:) but I think there's no harm in exposing this parameter and preserving the default behavior.

Down the road, I could imagine expanding the `eval` results to include `_ext` metrics similar to how we include `_inv` metrics (for inversion-sensitive evaluation).

tomxi:
Just to add:
This chord actually happened in our dataset.
Here's the played chord as represented by tablature.

[...]

PR #275, opened 2018-05-04
[WIP] chord.evaluate with extension reductions. implements #274

bmcfee:
This PR implements extension-reduction in the chord metrics.

It's currently WIP, as it has uncovered some undocumented quirks in chord eval that I'd like to get settled before we merge this.

### DIFF
--- a/mir_eval/chord.py
+++ b/mir_eval/chord.py
@@ -709,7 +709,7 @@ def weighted_accuracy(comparisons, weights):
     return np.sum(comparisons*normalized_weights)
 
 
-def thirds(reference_labels, estimated_labels):
+def thirds(reference_labels, estimated_labels, reduce_extended_chords=False):
     """Compare chords along root & third relationships.
 
     Examples
@@ -736,6 +736,11 @@ def thirds(reference_labels, estimated_labels):
         Reference chord labels to score against.
     estimated_labels : list, len=n
         Estimated chord labels to score against.
+    reduce_extended_chords : bool
+        Whether to map the upper voicings of extended chords (9's, 11's, 13's)
+        to semitone extensions.
+        (Default value = False)
+
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


ejhumphrey:
well then, i'm pleasantly surprised this works. lgtm

_grumble chords grumble_

bmcfee:
> well then, i'm pleasantly surprised this works. lgtm so far
> 
> grumble chords grumble

The strange thing is that the test case hits on thirds, triads, and tetrads, but fails only on sevenths.  (And majmin, I guess, maybe we should consider changing that one...)

bmcfee:
(warning: nitpicking to follow.  sorry!)

I'm going through the chord metrics implementation, just to see if there are any other likely oddities.  Here are some things I've noticed that seem strange:

- [thirds](https://github.com/craffel/mir_eval/blob/master/mir_eval/chord.py#L751) should this not be matching on `[:, 3:4]` instead of `[:, 3]`?  Should it be including the 2nd (or 9th) in this check?  I always understood this metric as `est_root == ref_root and est_third == ref_third` (regardless of other tones). 

- [triads](https://github.com/craffel/mir_eval/blob/master/mir_eval/chord.py#L847) similar comment: why are we looking at the 2nd and 4th here (implied by `[:8]`)?  Shouldn't this just depend on (root, m3, M3, TT, P5, m6) bins?

- [majmin](https://github.com/craffel/mir_eval/blob/master/mir_eval/chord.py#L1151-L1152) seems like the idea here is to only look at things below the 6th, and check for exact matching against the maj/min qualities.  As noted in #274, this gets dicey with extension reduction.  Maybe that's fine?

- [sevenths](https://github.com/craffel/mir_eval/blob/master/mir_eval/chord.py#L1275) missing some qualities.  As noted in #274, this one behaves differently from triads and tetrads, which is concerning.  Triads counts matches if the ref and est agree to the first 8 positions, tetrads does the same but for all positions.  Sevenths breaks this pattern, and works more like the maj/min metric (check for equivalence to one of the template profiles).

Issue #269, opened 2017-11-09
Migrate to pytest


bmcfee:
As mentioned in #267, our dependence on nosetests instead of pytest is getting to be a problem.

Rewriting the test suite is a big, but monotonous job, but we'll have to do it.  Seems like a good project for a sprint at some point.

PR #277, opened 2018-05-29
fixed deprecations and updated requirements

### LINKED ISSUE #276: display module needs to be updated for matplotlib deprecations
MPL has deprecated string literals for flag controls (`'on'` => `True`, `'off'` => 'False'), so we get warnings now in our tests.

I think this only comes up in places using `grid`, but we should do a thorough audit.

bmcfee:
This PR fixes #276 by updating the display, segment, and doc build to reflect recent deprecations.  Version requirements have been updated accordingly.

There should be no functionality change here.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -32,7 +32,7 @@ sys.path.insert(0, os.path.abspath('..'))
 # ones.
 extensions = [
     'sphinx.ext.autodoc',
-    'sphinx.ext.pngmath',
+    'sphinx.ext.imgmath',
     'numpydoc',
 ]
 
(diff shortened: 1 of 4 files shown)


craffel:
Looks good.  Looks like there is one more deprecation warning in the CI, not sure if you want to fix here or not:
https://travis-ci.org/craffel/mir_eval/jobs/385206565#L1232

bmcfee:
Yeah, I noticed that.  It's in the image comparison tests, and I think out of scope for what I want to do here.  A proper fix for that would be included in #269 .

PR #279, opened 2018-05-31
isolated matplotlib side effects

### LINKED ISSUE #278: isolate pyplot import and side-effects
(Note: we recently fixed this over in librosa [here](https://github.com/librosa/librosa/pull/704).)

The issue is that importing `matplotlib.pyplot` triggers the backend selector, which has side-effects (including occasional memory leaks).  But we don't always need to import `pyplot` in the display module.  If the user is working with the object API (as opposed to the stateful pyplot API), then we can avoid this import entirely.

In fact, the only place the `pyplot` import is used is within our axes getter, and even then, only when the axes are not provided by the user.  It's an easy fix to move the import to within that conditional block, and avoid any side-effects until they are absolutely necessary.

bmcfee:
Fixes #278 by only importing `matplotlib.pyplot` when absolutely necessary.

Provided CI passes, I see no need for CR here.

### DIFF
--- a/mir_eval/display.py
+++ b/mir_eval/display.py
@@ -6,7 +6,6 @@ from collections import defaultdict
 import numpy as np
 from scipy.signal import spectrogram
 
-import matplotlib.pyplot as plt
 from matplotlib.patches import Rectangle
 from matplotlib.ticker import FuncFormatter, MultipleLocator
 from matplotlib.ticker import Formatter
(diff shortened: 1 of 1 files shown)


bmcfee:
Looks like coveralls is stalled, but everything ought to be green here.

craffel:
Thank you sir!

Issue #298, opened 2019-01-07
tempo: why not allow 0.0 as reference tempo?


hendriks73:
In the docs to `tempo` it says:

> Reference tempi should be strictly positive, and provided in ascending order
as a numpy array of length 2.  Estimated tempi are allowed to be 0, but
otherwise are subject to the same constraints as reference.

I'm wondering why we are dictating that reference tempi must be greater than zero? How do we represent a non-ambiguous track with `T1, T2, ST1 = None, 100.0, 0.0`? The easiest format would actually be `(0.0, 100.0, 0.0)`, but that fails in [validate_tempi](https://github.com/craffel/mir_eval/blob/master/mir_eval/tempo.py#L41).

That aside, It seems to me that contrary to the docs the estimated tempi are not allowed to be 0, either.

bmcfee:
Good point.  In fact, the docstrings all say "non-negative" (not "positive"), which indicates that 0 should be allowed.

However, the reason for this comes down to the definition of the tempo detection metrics [here](https://github.com/craffel/mir_eval/blob/master/mir_eval/tempo.py#L122-L123), which requires a division by `ref_t`.

I think what we actually want is for at least one strictly positive value in `reference_tempi` (estimated can both be 0), and proper handling of divide-by-zero cases in the metric.  This complicates the validation logic a bit, but I'd support this change if someone wants to make it.

hendriks73:
Gee, Brian—you're quick!

> I think what we actually want is for at least one strictly positive value in reference_tempi (estimated can both be 0)

Agreed.

divide-by-zero logic change—like this?
```
for ref_t in reference_tempi:
        if ref_t == 0:
            continue
        # Compute the relative error for this reference tempo
        relative_errors.append(np.min(
            np.abs(ref_t - estimated_tempi) / float(ref_t)))

        # Count the hits
        hits.append(bool(relative_errors[-1] <= tol))

```

> but I'd support this change if someone wants to make it.

I'm interested in "fixing this" and will send a PR some time tomorrow.

bmcfee:
You could also stick an `else` clause on that for loop to catch the case where there are no non-zero reference tempi.

PR #299, opened 2019-01-08
Allow one reference tempo and both estimate tempi to be zero. The ide…

hendriks73:
…a here is that some tracks in the ground truth may not be ambiguous and some estimators might be unable to produce any reasonable value and therefore choose zero for both estimate tempi.

This is a little more than what we had talked about in #298, but I believe it to be reasonable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ notifications:
 
 python:
     - "2.7"
-    - "3.4"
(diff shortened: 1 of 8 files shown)


craffel:
Looks like PEP-8 is failing.

craffel:
Thanks!

Issue #300, opened 2019-01-08
tempo: why is tol=0.0 not allowed


hendriks73:
The current code does not allow zero tolerance, i.e. `tol=0.0`. I don't see any reason why.
Is there one?

But there is a practical use-case for allowing 0.0:

When creating a graph that shows p-score over tolerance, it makes a lot of sense to start with `tol=0.0` and not `tol=1e-12`. I'd like to change this to make working with `mir_eval` as a library a little easier.

bmcfee:
I think the reasoning here is that floating point comparisons are almost never that precise, and allowing `tol=0` might suggest to a naive user that exact comparison is reliable.  I don't see much point in having tolerance below machine epsilon, but I think I also don't completely understand your proposed use-case.  Can you elaborate on that?

hendriks73:
I'm interested in graphing how tolerance affects system performance. Therefore I create graphs like this (axis labels have wrong unit):

<img width="996" alt="screen shot 2019-01-08 at 16 35 27" src="https://user-images.githubusercontent.com/4749755/50840641-3dbd0a80-1363-11e9-91da-a550acd48063.png">

And for that purpose it's *convenient* (not necessary) to have a value for `tol=0.0`.

> I think the reasoning here is that floating point comparisons are almost never that precise, and allowing tol=0 might suggest to a naive user that exact comparison is reliable.

Do naive users really use `mir_eval`?

> I don't see much point in having tolerance below machine epsilon

One could just as well argue that there is not point in having machine epsilon tolerance. Either seems arbitrary.

bmcfee:
> Do naive users really use `mir_eval`?

I count myself as naive most of the time, so yes. :grin: What I really mean here is that we should adhere to the principle of least surprise, and floating point can be full of surprises when we start asking for strict equivalence.



> One could just as well argue that there is not point in having machine epsilon tolerance. Either seems arbitrary.

Epsilon as a tolerance doesn't seem arbitrary: it's explicitly acknowledging the machine's limitations in determining whether two numbers are the same given the representation.

At any rate, I don't feel too strongly about this point, but I prefer to err on the side of caution when dealing with floating point comparisons.

hendriks73:
Well, I don't think any harm would be done, if we allowed zero tolerance.

From the top of my head I cannot imagine any surprises this might produce. But I can tell you for sure that I was surprised when I realized that `mir_eval` does not allow zero tolerance for tempo.

[...]

PR #301, opened 2019-01-09
Allow zero tolerance, but issue a warning.

hendriks73:
Change as discussed in #300, i.e. following @stefan-balke 's suggestions to allow a tolerance of zero, but issue a warning.


craffel:
Hey Hendrik, Travis is passing again. This looks good but I think you will have to include a check for the warning in your test. Do you want to add that? That will also retrigger Travis :)

hendriks73:
Lines too long... gee...

craffel:
Merged, thanks!

hendriks73:
Thank you!

PR #305, opened 2019-01-15
Use toarray instead of todense for sparse matrices

craffel:
todense returns a numpy matrix object, whereas toarray returns a numpy ndarray. The former causes a DeprecationWarning. The flood of DeprecationWarnings in our tests eventually causes Travis to fail.


bmcfee:
LGTM -- tests are still failing, but it's because of something wrong in chords, which I expect is an independent failure.

PR #307, opened 2019-01-16
Use tuple index in chord.rotate_bitmap_to_root

### LINKED ISSUE #306: Test failure in chord on Python 3.5
Our chord test is failing on Python 3.5:
https://travis-ci.org/craffel/mir_eval/jobs/480136121#L9494
I think this is because there are actually two warnings being triggered, one of which is 
`mir_eval/chord.py:573: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use arr[tuple(seq)] instead of arr[seq]. In the future this will be interpreted as an array index, arr[np.array(seq)], which will result either in an error or a different result.`

craffel:
Using a list causes a warning with the latest numpy. Fixes #306.

### DIFF
--- a/mir_eval/chord.py
+++ b/mir_eval/chord.py
@@ -587,7 +587,7 @@ def rotate_bitmap_to_root(bitmap, chord_root):
     idxs = list(np.nonzero(bitmap))
     idxs[-1] = (idxs[-1] + chord_root) % 12
     abs_bitmap = np.zeros_like(bitmap)
-    abs_bitmap[idxs] = 1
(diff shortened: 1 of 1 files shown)


bmcfee:
LGTM.  Last fail is due to matplotlib having removed the nose classes.

Two suggestions:

1. drop nose in favor of pytest (should be done, but it takes some effort)
2. pin the matplotlib version for testing purposes until it can be properly fixed.  You can do this by having an extras_require for `tests` of `matplotlib<3` in setup.py.

PR #312, opened 2019-01-30
fix raw chroma accuracy for unvoiced estimates

### LINKED ISSUE #311: melody.raw_chroma_accuracy wrong scores on unvoiced estimates
If the input estimate is unvoiced, the chroma accuracy metric should return 0. Instead it counts reference frames that are the multiples of 1200 cents as correctly estimated frames. Consider this example:
```
import mir_eval
import numpy as np

ref_time = np.array([0, 1, 2])
est_time = np.array([0, 1, 2])
ref_freq = np.array([40, 80, 160])
est_freq = np.array([0, 0, 0])

arrays = mir_eval.melody.to_cent_voicing(ref_time, ref_freq, est_time, est_freq)
# ref_cent = [2400., 3600., 4800.]
chroma = mir_eval.melody.raw_chroma_accuracy(*arrays)

print(chroma)
# should be 0.0 but actually is 1.0
```

Just for fun I averaged the raw chroma score of unvoiced estimates on MedleyDB, the result is 7% accuracy.

```
results = []
for t in mdb.load_melody_multitracks():
    ref_time, ref_freq = np.array(t.melody2_annotation).T
    est_time = ref_time
    est_freq = np.zeros_like(ref_freq)
    arrays = mir_eval.melody.to_cent_voicing(ref_time, ref_freq, est_time, est_freq)
    results.append(mir_eval.melody.raw_chroma_accuracy(*arrays))
print(np.mean(results))
# 0.07011304858972267
```

kukas:
Fix #311

### DIFF
--- a/mir_eval/melody.py
+++ b/mir_eval/melody.py
@@ -484,7 +484,8 @@ def raw_pitch_accuracy(ref_voicing, ref_cent, est_voicing, est_cent,
     # estimate provides a correct frequency value (within cent_tolerance cents)
     # NB: voicing estimation is ignored in this measure
     matching_voicing = ref_voicing * (est_cent > 0)
-    frame_correct = (np.abs(ref_cent - est_cent)[matching_voicing] < cent_tolerance)
+    cent_diff = np.abs(ref_cent - est_cent)[matching_voicing]
+    frame_correct = (cent_diff < cent_tolerance)
     raw_pitch = (frame_correct).sum()/float(ref_voicing.sum())
 
     return raw_pitch
@@ -565,7 +566,8 @@ def raw_chroma_accuracy(ref_voicing, ref_cent, est_voicing, est_cent,
     cent_diff = np.abs(ref_cent - est_cent)
     octave = 1200*np.floor(cent_diff/1200.0 + 0.5)
     matching_voicing = ref_voicing * (est_cent > 0)
-    frame_correct = (np.abs(cent_diff - octave)[matching_voicing] < cent_tolerance)
+    cent_diff = np.abs(cent_diff - octave)[matching_voicing]
+    frame_correct = (cent_diff < cent_tolerance)
     n_voiced = float(ref_voicing.sum())
     raw_chroma = (frame_correct).sum()/n_voiced
     return raw_chroma
--- a/mir_eval/melody.py
+++ b/mir_eval/melody.py
@@ -483,7 +483,8 @@ def raw_pitch_accuracy(ref_voicing, ref_cent, est_voicing, est_cent,
     # Raw pitch = the number of voiced frames in the reference for which the
     # estimate provides a correct frequency value (within cent_tolerance cents)
     # NB: voicing estimation is ignored in this measure
-    frame_correct = (np.abs(ref_cent - est_cent)[ref_voicing] < cent_tolerance)
+    matching_voicing = ref_voicing * (est_cent > 0)
+    frame_correct = (np.abs(ref_cent - est_cent)[matching_voicing] < cent_tolerance)
     raw_pitch = (frame_correct).sum()/float(ref_voicing.sum())
 
     return raw_pitch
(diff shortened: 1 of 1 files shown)


justinsalamon:
Gave it a quick look (it's late in Europe), in principle fix looks correct, but will try to give a second look tomorrow to confirm.

justinsalamon:
Didn't have time today, but note that the regression tests fail (due to the change in the chroma metric). I imagine that, assuming the fix is confirmed to be correct, the regression data will have to be updated as part of this PR before it can be merged.

mir_eval/melody.py
```diff
@@ -563,7 +563,8 @@ def raw_chroma_accuracy(ref_voicing, ref_cent, est_voicing, est_cent,
     # Raw chroma = same as raw pitch except that octave errors are ignored.
     cent_diff = np.abs(ref_cent - est_cent)
     octave = 1200*np.floor(cent_diff/1200.0 + 0.5)
-    frame_correct = (np.abs(cent_diff - octave)[ref_voicing] < cent_tolerance)
+    correct_voicing = ref_voicing * est_voicing
+    frame_correct = (np.abs(cent_diff - octave)[correct_voicing] < cent_tolerance)
```
rabitt: (cc @juanjobosch who stared at this with me!)

This almost fixes the problem, but with one annoying subtlety - f0 values in the input can be given as "negative" frequency values - in this case, the difference should be computed over all the reference frames, using the absolute value of the frequency given. It's not very explicit in the definition, but it's meant to be a way to measure pitch/chroma accuracy independent of voicing mistakes:
> Raw chroma accuracy: the fraction of voiced frames in ref_cent for which est_cent provides a correct frequency values (within cent_tolerance cents), ignoring octave errors

Consider the case:
```
ref_t = [1,2,3]
est_t = [1,2,3]
ref_f = [440, 440, 440]
est_f = [-220,-220,-220]
```
The Raw Chroma Accuracy should be 100%, even though the estimate voicing is `[0,0,0]`. 

So instead, I propose the following fix that solves the bug but doesn't break the negative frequency use case.
```
matching_voicing = ref_voicing * (est_cent > 0)
frame_correct = (np.abs(cent_diff - octave)[matching_voicing] < cent_tolerance)
```
In short, this only count frames as "correct" if the estimate is non-zero (where zero indicates an unvoiced frame where no "best guess" frequency is given) AND within the tolerance.

In fact, this bug exists for raw_pitch_accuracy too... but the "problem" case never comes up because we almost never report frequencies below 40 Hz, so these cases always fall outside of the tolerance threshold. For good measure though, we should make the same fix for raw_pitch_accuracy. (cc @craffel )

rabitt:
See my comment above, but in summary:
    (1) the fix is *almost* correct, but doesn't work correctly when estimates are reported as negative frequencies.
    (2) the same bug exists for `raw_pitch_accuracy`, but in practice, we probably haven't hit any cases where the value of the metric was affected.

craffel:
Good catch, @kukas do you mind implementing @rabitt's version and then we can merge?

kukas:
> Good catch, @kukas do you mind implementing @rabitt's version and then we can merge?

Sure, I will fix it on Thursday, if it can wait. Thanks @rabitt for the review!

justinsalamon:
Thanks @rabitt and @juanjobosch for reviewing this and good catch on the neg case (sorry, totally swamped over here with on-boarding).

@kukas note @rabitt renamed `correct_voicing` to `matching_voicing`, which I think is a semantically more helpful name :)

kukas:
Now it should be fixed, I used @rabitt `s patch and applied it also to raw pitch accuracy.

craffel:
@kukas thanks, you'll need to fix the PEP-8 errors before Travis lets you pass though. https://travis-ci.org/craffel/mir_eval/jobs/490483158#L9358

@rabitt mind giving this a look and LGTM?

craffel:
Thanks!

PR #318, opened 2019-07-02
Solves bss permutation efficiently

fakufaku:
The current implementation of `bss_eval` exhaustively tests all permutations of sources which has factorial cost. It is possible to find the permutation in quadratic time using a minimum weight matching algorithm for bipartite graphs. This algorithm is implemented by the function `scipy.optimize.linear_sum_assignment`. A wrapper around the scipy function is added to handle possible infinite entries in the cost matrix.

I tried to run the tests, but got some errors due to missing files.


craffel:
Thanks! I remember this being brought up before but I can't find any reference to it.

@faroit @aliutkus any thoughts?

fakufaku:
Hi @craffel , thanks for the reply! I've also started discussing this with @aliutkus [here](https://github.com/sigsep/bsseval/issues/5).

bmcfee:
Since separation is deprecated in 0.8, I'll close this out now.

PR #320, opened 2019-08-25
Add comment support to io methods

### LINKED ISSUE #319: IO functions do not skip comment lines
Title says it all -- io functions are unable to parse files with commented lines (by convention, those starting with # in .lab format).

Fixing this violates my general rule against writing our own *sv parser, but i think there's enough data out there to warrant an exception in this case.  @craffel if this sounds good to you, I'll prep a PR.

bmcfee:
This pr fixes #319 by adding a `comment=` flag to (almost) all* data loaders.  The semantics are as follows:

- By default, any line starting with `'#'` is skipped.
- A user can override the comment string with any regular expression.  The regexp is constructed to only match at the beginning of the line.  For example, calling with `comment=';'` uses the semicolon as a comment leader.
- Comment skipping can be disabled by specifying `comment=None`.  This is equivalent to the current default behavior, but I don't think it should be the default going forward.  It's increasingly common for annotation tools to leave metadata in the resulting data file, and we should be encouraging of that pattern.

* exceptions for wav (obviously) and pattern, the latter because the pattern format doesn't seem to have a common grammar for comments.  The other formats implicitly use some flavor of CSV/TSV, which generally support comment strings.


mir_eval/io.py
```diff
@@ -70,6 +75,12 @@ def load_delimited(filename, converters, delimiter=r'\s+'):
     # Create re object for splitting lines
     splitter = re.compile(delimiter)
 
+    # And one for comments
+    if comment is None:
+        commenter = None
```
craffel: I don't think you need this condition, right? The only place `commenter` is used is in the line 
```Python
if comment is not None and commenter.match(line):
```
so if `comment is None` the `commenter` variable will never be encountered, and you can just do 
```Python
if comment is not None:
    commenter = re.compile('^{}'.format(comment))
```
here. Unless I am misunderstanding. (same note about line 629)

bmcfee: That's true, but linters will be annoyed.

craffel: Really? I thought the linters were smarter than that.

bmcfee: Some are, not all.  I had originally written it as a short-circuit the way you suggest, but it's a bit more readable making things explicit IMO.

craffel:
Thank you sir, always a pleasure.

PR #323, opened 2019-11-01
Fix interpolation in sonify.time_frequency

### LINKED ISSUE #322: mir_eval.chord.sonify starts late/ends early
When running the chord sonification code, I'm noticing that the first chord always starts late, and the last chord ends early. For example:

```python
import mir_eval

fs = 8000
sonification = mir_eval.sonify.chords(['E:min', 'A:7'], np.array([[0, 1], [1, 2]]), fs=fs)

plt.figure(figsize=(15, 7))
plt.plot(sonification)
```
Since the first chord starts at 0 second and the last one ends at 2 seconds, I'd expect to see audio for the entire duration of the clip, but I don't:

download

cc @bmcfee

daturkel:
This fixes #322. The change is (over)-detailed in [this comment](https://github.com/craffel/mir_eval/issues/322#issuecomment-546654110) but the gist is: because the interpolator used in `time_frequency` uses the midpoints of chord intervals as boundaries for linear interpolation, it misses the period before the midpoint of the first chord interval and the period after the midpoint of the last chord interval. 

Using a tuple in `fill_value` allows us to tell the interpolator to handle the case where we're before the first midpoint or after the last.

### DIFF
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -163,7 +163,7 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None,
             gram_interpolator = interp1d(
                 time_centers, gram[n, :],
                 kind='linear', bounds_error=False,
-                fill_value=(gram[n,0],gram[n,-1]))
+                fill_value=(gram[n, 0], gram[n, -1]))
         # If only one time point, create constant interpolator
         else:
             gram_interpolator = _const_interpolator(gram[n, 0])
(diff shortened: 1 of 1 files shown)


mir_eval/sonify.py
```diff
@@ -163,7 +163,7 @@ def __interpolator(x):
             gram_interpolator = interp1d(
                 time_centers, gram[n, :],
                 kind='linear', bounds_error=False,
-                fill_value=0.0)
+                fill_value=(gram[n,0],gram[n,-1]))
```
craffel: Please fix
```
mir_eval/sonify.py:166:35: E231 missing whitespace after ','
mir_eval/sonify.py:166:38: E231 missing whitespace after ','
mir_eval/sonify.py:166:45: E231 missing whitespace after ','
```

craffel:
Thanks Dan, just one PEP-8 issue and then we can merge.

daturkel:
Should be fixed per 177a2fb

craffel:
Thanks!

Issue #317, opened 2019-06-21
Generalized melody metrics


rabitt:
Myself and @juanjobosch have a recently accepted ISMIR paper proposing an extension of the existing melody metrics (allows for continuous, rather than binary, voicing estimations). The new metrics are equivalent to the existing ones when binary voicings are given.

Can we add these new metrics to mir_eval? They remove the need for reporting "unvoiced" frames as negative frequency values, which would clean up the underlying code a lot. But, it would change the existing API - right now the input to `mir_eval.melody` is `ref_times, ref_freqs, est_times, est_freqs`. To use the new metrics, it would need `ref_times, ref_freqs, ref_voicing, est_times, est_freqs, est_voicing`.

Alternatively, we could keep `mir_eval.melody` as-is and instead create `mir_eval.f0` with the new metrics. The metrics are already used for more tasks than melody (e.g. pitch tracking, bass tracking), so `f0` could be a nice generalization.

@craffel , what do you think?

justinsalamon:
Quick comment - @juanjobosch also proposed some melody metrics that factor into account pitch continuity back in 2016. Would it be worth considering adding those in as well?

In both cases this diverges from the "standard" metrics used in MIREX, but I think these metrics are not sufficient, so if we think this can encourage better eval in this area I'd be all for adding both the soft voicing metrics and pitch continuity metrics.

I might however, be worth supporting backwards compatibility (neg freqs for voicing) for the sake of keeping the API compatible with MIREX and older algs?

craffel:
Some additional (ugly) options:

1. Make the args  `ref_times, ref_freqs, est_times, est_freqs, ref_voicing=None, est_voicing=None` so that these new inputs are optional. 
1. Make `ref_freq` optionally be 2D. When 1D, assume the format where a negative value indicates unvoiced. When 2D, Row 1 is (strictly positive) frequency estimate and row 2 is the voicing.

I generally prefer not to change the API of existing functions, but if you expect your new metrics to be a very common use-case then I could be convinced. Note that you are not actually changing the API for any of the metric functions, only for `to_cent_voicing` which is a preprocessing function.

If your new metrics are significantly different from the existing ones (so that minimal code could be shared), `mir_eval.f0` is an OK idea. Otherwise, I am not inclined to add an additional module.

To @justinsalamon's point, I agree that either way we must maintain backward compatibility.

justinsalamon:
Between options 1 and 2 perhaps 1 is better?

[...]

PR #324, opened 2020-02-11
Generalized Melody Metrics

rabitt:
Implements #317 , cc @juanjobosch

The metrics behave identically as before, but now users can optionally specify a voicing array, which overwrites any voicing coming from negative frequency values.

Note that I (mostly) removed all conversion to cents, since the conversion cancels out in the metrics. The only place where the cents values remain is in `resample_melody_series`. Right now when the frequency values are resampled, they're resampled in cents, not in Hz. If we were to instead resample in Hz, and all of the metrics will go up (slightly) - but I left it as is for now.

This PR is missing:
- [x] tests for evaluation when a voicing array is provided
- [x] a reference in the documentation to the new metrics


mir_eval/melody.py
```diff
@@ -306,130 +295,95 @@ def to_cent_voicing(ref_time, ref_freq, est_time, est_freq, base_frequency=10.,
...
 
-    return vx_recall, vx_false_alm
+def cents2hz(freq_cents, base_frequency=10.0):
+    freq_hz = np.zeros(freq_cents.shape[0])
+    freq_nonz_ind = np.flatnonzero(freq_cents)
+    freq_hz[freq_nonz_ind] = 2.0 ** (freq_cents[freq_nonz_ind] / 1200.0) * base_frequency
+    return freq_hz
 
 
-def raw_pitch_accuracy(ref_voicing, ref_cent, est_voicing, est_cent,
```
lostanlen: shouldn't the `cent_tolerance` kwarg go through a deprecation cycle first?

mir_eval/melody.py
```diff
@@ -448,63 +402,58 @@ def raw_pitch_accuracy(ref_voicing, ref_cent, est_voicing, est_cent,
...
-    return raw_pitch
+    nonzero_freqs = np.logical_and(est_freqs != 0, ref_freqs != 0)
+    divisor = np.abs(est_freqs[nonzero_freqs]) / ref_freqs[nonzero_freqs]
+    freq_diff_semitones = np.abs(12.0 * np.log2(divisor))
+    correct_frequencies = freq_diff_semitones < semitone_tolerance
+    rpa = np.sum(ref_voicing[nonzero_freqs] * correct_frequencies) / np.sum(ref_voicing)
+    return rpa
 
 
-def raw_chroma_accuracy(ref_voicing, ref_cent, est_voicing, est_cent,
```
lostanlen: again, i'm concerned of dropping the `cent_tolerance` kwarg in a user-facing function like that without a `DeprecationWarning`

mir_eval/melody.py
```diff
@@ -306,130 +295,95 @@ def to_cent_voicing(ref_time, ref_freq, est_time, est_freq, base_frequency=10.,
...
-    else:
-        vx_recall = TP/float(TP + FN)
 
-    # Voicing false alarm = fraction of unvoiced frames according to the
-    # reference that are declared as voiced by the estimate
-    if FP + TN == 0:
-        vx_false_alm = 0.
-    else:
-        vx_false_alm = FP/float(FP + TN)
+def hz2cents(freq_hz, base_frequency=10.0):
```
craffel: It looks like you dropped the docstring for this - any reason why and why it needs to be moved?

justinsalamon:
@craffel per MIREX convention, melex algs typically output values in Hz (even if they internally work in cents, they'll typically convert to Hz for the final output). The instructions for MIREX 2019 seem unchanged with respect to previous years (algs need to output frequency values in Hz with 0/neg values to indicate unvoiced frames), so I guess the convention is still going strong. 

I think it'd be great to support taking in a voicing vector, but agree we should remain backwards compatible.

craffel:
Ok, in that case why did we design our metrics to take in cents in the first place?

justinsalamon:
My memory might fail me, but IIRC they did, originally take in Hz (at least in the very initial implementation pass), but since they involve an internal conversion to cents to compute the metrics, at some point we decided to factor the conversion out and provide utility functions. Still, our high-level functions such as [evaluate](https://github.com/craffel/mir_eval/blob/master/mir_eval/melody.py#L636) still expect input in Hz.

rabitt:
The metrics used to (and still do in the PR) take frequency values in Hz. Internally they are converted to cents, but the conversion isn't necessary to compute the metrics. In the PR I removed the conversion to simplify the code, and the metrics are mathematically equivalent.

mir_eval/melody.py
rabitt: I'd initially taken it out completely but realized it had to go back in because resampling in cents != resampling in Hz. I'll put it back where it was!

craffel:
> The metrics used to (and still do in the PR) take frequency values in Hz.

Not sure what you mean - I think they currently take cents, e.g. 
https://github.com/craffel/mir_eval/blob/master/mir_eval/melody.py#L429
https://github.com/craffel/mir_eval/blob/master/mir_eval/melody.py#L494
https://github.com/craffel/mir_eval/blob/master/mir_eval/melody.py#L576
The evaluate function takes in Hz and converts to cents, but user code might call the metric functions directly instead of evaluate.

rabitt:
@craffel mmm yep you're right - I was thinking just about `melody.evaluate`. You and @lostanlen are right that while `melody.evaluate` is backwards compatible (if I change back `cent_tolerance`), the other functions are far from it. The way the individual metrics are written now though, they're pretty annoying to use - if you want to call them using "standard" melody outputs (frequencies in Hz, negative frequencies indicating voicing, different time scales) you have to first manually call `melody.to_cent_voicing` and  before the individual metrics would work properly.

How should I proceed? I can rewrite the PR to be fully backwards compatible, but it will complicate the code a bunch. Or, can we do an in between PR with a deprecation warning on the individual metrics before this one is released? Or something else?

craffel:
> The way the individual metrics are written in the current release makes them hard to use individually - if you want to call them using "standard" melody outputs (frequencies in Hz, negative frequencies indicating voicing, different time scales) you have to first manually call melody.to_cent_voicing and before the individual metrics would work properly.

I agree that it's cumbersome, but it's not unheard of in `mir_eval`, e.g. all the `beat` metrics require that you call `trim_beats` first e.g.
https://github.com/craffel/mir_eval/blob/master/mir_eval/beat.py#L193

> How should I proceed? I can rewrite the PR to be fully backwards compatible, but it will complicate the code a bunch. Or, can we do an in between PR with a deprecation warning on the individual metrics before this one is released? Or something else?

We haven't handled deprecation before in `mir_eval`. It's especially hard since they are being used as positional arguments. @bmcfee @lostanlen do you all have any experience/tips on how to do this from `librosa`?

lostanlen:
> We haven't handled deprecation before in mir_eval. It's especially hard since they are being used as positional arguments. @bmcfee @lostanlen do you all have any experience/tips on how to do this from librosa?

I don't remember librosa having changed the unit of one ot its argument. As far as i can tell, all we've done is remove kwargs and remove function names. Here, the only option i see is (sigh):

1) give a new name to the proposed `validate`, e.g.  `validate_cents`.
2) copy-paste `validate` into a new name, e.g. `validate_hertz`.
3) raise a `Warning` every time the old `validate` is being deprecated, and say that `validate` will be in cents in the future. Users should rename their callsite to `validate_hertz(` if they want to retain the old behavior, and `validate_cents(` if they want to adopt the new behavior.
4) make a new release
5) remove the `validate` function
6) make a new release.
7) wait a bit.
8) People complain about `NameError: name 'validate' is not defined`. We point them to the right direction.
9) make a new release
10) add a `DeprecationWarning` to `validate_hertz`
11) copy-paste `validate_cents` into a new name: `validate`
12) add a `DeprecationWarning` to `validate_cents` (renamed to `validate`)
13) make a new release
14) wait a long time.
15) Remove `validate_hertz` and `validate_cents`
16) make a new release
17) 🤞

craffel:
Yes, that sounds about right, except we'd need to do that for all of the metric functions (not just validate).

rabitt:
@craffel @lostanlen Thanks for all your comments. I finally found some time to get back to this - I've redone this PR and it should now be fully backwards compatible, and ready for a re-review.

Note this is still missing tests for non-binary voicing, but wanted to first check that the change is otherwise OK.

mir_eval/melody.py
```diff
@@ -261,8 +274,11 @@ def resample_melody_series(times, frequencies, voicing,
         frequencies_resampled = scipy.interpolate.interp1d(times,
                                                            frequencies,
                                                            kind)(times_new)
+
     # Use nearest-neighbor for voicing if it was used for frequencies
-    if kind == 'nearest':
+    # if voicing is not binary, use linear interpolation
+    is_binary_voicing = np.array_equal(np.round(voicing), voicing)
```
craffel: Is it any safer to do `np.all(np.logical_or(np.equal(voicing, 0), np.equal(voicing, 1)))`?

tests/test_melody.py
```diff
@@ -113,7 +114,7 @@ def __unit_test_voicing_measures(metric):
         warnings.simplefilter('always')
         # First, test for warnings due to empty voicing arrays
         score = metric(np.array([]), np.array([]))
-        assert len(w) == 4
+        # assert len(w) == 4
```
craffel: ping

tests/test_melody.py
```diff
@@ -154,14 +153,12 @@ def __unit_test_melody_function(metric):
         metric(np.ones(10), np.arange(10), np.zeros(10), np.arange(10))
         assert len(w) == 7
         assert issubclass(w[-1].category, UserWarning)
+        print(str(w[-1].message))
```
craffel: ping

mir_eval/melody.py
```diff
@@ -126,40 +134,45 @@ def hz2cents(freq_hz, base_frequency=10.0):
...
+    freq_cent : np.ndarray
         Array of frequencies in cents, relative to base_frequency
-
     """
     freq_cent = np.zeros(freq_hz.shape[0])
     freq_nonz_ind = np.flatnonzero(freq_hz)
-    normalized_frequency = np.abs(freq_hz[freq_nonz_ind])/base_frequency
-    freq_cent[freq_nonz_ind] = 1200*np.log2(normalized_frequency)
-
+    normalized_frequency = np.abs(freq_hz[freq_nonz_ind]) / base_frequency
```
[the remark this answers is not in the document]
rabitt: :)

mir_eval/melody.py
rabitt: yep agreed

craffel:
Any melody people want to chime in? @juanjobosch @justinsalamon

lostanlen:
"melody people" 😍 

supporting information attached: https://www.youtube.com/watch?v=6Sl-ad304Aw

craffel:
a wild funky appears

mir_eval/melody.py
```diff
@@ -294,6 +317,12 @@ def to_cent_voicing(ref_time, ref_freq, est_time, est_freq, base_frequency=10.,
         Time of each estimated frequency value
     est_freq : np.ndarray
         Array of estimated frequency values
+    est_voicing : np.ndarray
+        Estimate voicing confidence.
+        Default None, which means the voicing is inferred from ref_freq
```
lostanlen: would be good to say here how this default inference happens

juanjobosch:
Hey @craffel, sure! I can have a look in a couple of days.

mir_eval/melody.py
rabitt: I just added this actually! - see my push from 5 min ago

rabitt:
Just added a bunch of tests (which caught a baddddd bug in the new code), so pending review, this is done on my side!

mir_eval/melody.py
rabitt: There's even a bug in this docstring... should be "..is inferred from est_freq". I've updated and clarified!

justinsalamon:
I'm tracking this, but totally swamped right now so @juanjobosch will probably get to this sooner than I do, but if not I'll give it a pass as soon as I can.

mir_eval/melody.py
lostanlen: great! i like this explanation 👍

juanjobosch:
A first remark, which I think we commented briefly offline some time ago, but may be worth mentioning it here too.

In the generalised metrics article, voicing-related accuracies, only consider non-binary voicing for the estimates (not for the annotations, and that's why in this implementation it uses ref_indicator, as a binary indicator of ref_voicing). And the formulas of pitch-related accuracies (RPA, RCA, OA) consider a "reward" for correct pitch and chroma estimates, which could be made equivalent to ref_voicing, as it is done in this implementation.

I think it may be ok to maintain the current terminology, as it probably makes it easier to keep the conceptual relation to the classic metrics (and therefore keep using e.g. "ref_voicing" in the RPA denominator, instead of e.g. "pitch_estimation_reward"), but I think it would be worth to at least add a comment in the functions where the terms in the paper's formula is different than the implementation (RPA, RCA, OA) saying something like the pitch estimation reward from [ref] is here made equivalent to ref_voicing.

mir_eval/melody.py
```diff
@@ -202,7 +224,7 @@ def resample_melody_series(times, frequencies, voicing,
     frequencies : np.ndarray
         Array of frequency values, >= 0
     voicing : np.ndarray
-        Boolean array which indicates voiced or unvoiced
+        Array which indicates voiced or unvoiced
```
juanjobosch, line 227: Would it be worth mentioning that the voicing array does not need to be binary here?

mir_eval/melody.py
```diff
@@ -655,6 +725,14 @@ def evaluate(ref_time, ref_freq, est_time, est_freq, **kwargs):
         Time of each estimated frequency value
     est_freq : np.ndarray
         Array of estimated frequency values
+    est_voicing : np.ndarray
+        Estimate voicing confidence.
+        Default None, which means the voicing is inferred from est_freq:
+            frames with frequency <= 0.0 are considered "unvoiced"
+            frames with frequency > 0.0 are considered "voiced"
+    ref_reward : np.ndarray
+        Reference voicing reward.
```
juanjobosch, line 734: According to the metrics, wouldn't it be more accurate to say this is a pitch estimation reward, even if it is related to the (non-binary) voicing of the reference?

juanjobosch, line 734:
```suggestion
        Reference voicing, acting as pitch estimation reward.
```

rabitt:
@craffel I think this should be good to go now. Just noticing - what happened to this repo's tests? I'm seeing tests run on my fork, but not on this PR.

craffel:
It looks like it's running on your PR but Travis isn't auto-commenting or anything. It also looks broken by other stuff :( https://travis-ci.org/github/craffel/mir_eval

craffel:
#327 and #328 should solve two of the Travis errors; please rebase on top of those. There is an additional error in Travis on top of this PR (but not at HED) that comes from the output of `display.piano_roll` changing 
```
FAIL: test_display.test_pianoroll.test
  File "/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/craffel/mir_eval/tests/mpl_ic.py", line 56, in failer
    result = f(*args, **kwargs)
  File "/home/travis/build/craffel/mir_eval/tests/mpl_ic.py", line 219, in do_test
    ' (RMS %(rms).3f)' % err)
matplotlib.testing.exceptions.ImageComparisonFailure: images not close: /home/travis/build/craffel/mir_eval/tests/result_images/test_display/piano_roll.png vs. /home/travis/build/craffel/mir_eval/tests/result_images/test_display/piano_roll-expected.png (RMS 26.578)
```
Mind checking if this PR would cause that?

rabitt:
@craffel I rebased on master and found the culprit of the error you mentioned. I updated the internals of some of the `mir_eval.display` functions to fix this. Everything is happily passing now :)

craffel:
Looks good, thanks so much for your hard work on this @rabitt !

PR #334, opened 2020-10-22
Support notation for unknown/ambiguous key or mode.

PRamoneda:
key evaluation. Add: 'X' if the Key can't be categorized or 'other' if the mode can't be categorized.

This standar is followed by at least:

GiantSteps key EDM dataset
Beatport key EDM dataset

The first one widely used in MIREX competition


PRamoneda:
In future submissions, I will be more precise with the name of the branch.

mir_eval/key.py
```diff
@@ -24,26 +24,28 @@
 
 KEY_TO_SEMITONE = {'c': 0, 'c#': 1, 'db': 1, 'd': 2, 'd#': 3, 'eb': 3, 'e': 4,
                    'f': 5, 'f#': 6, 'gb': 6, 'g': 7, 'g#': 8, 'ab': 8, 'a': 9,
-                   'a#': 10, 'bb': 10, 'b': 11}
+                   'a#': 10, 'bb': 10, 'b': 11, 'x': -10}
```
lostanlen, line 27: where does this `-10` value come from? Isn't this error-prone when computing keys mod 12 ?

mir_eval/key.py
```diff
@@ -89,13 +91,13 @@ def weighted_score(reference_key, estimated_key):
...
-    | Same key                                             | 1.0   |
+    | Same key and mode                                    | 1.0   |
     +------------------------------------------------------+-------+
     | Estimated key is a perfect fifth above reference key | 0.5   |
     +------------------------------------------------------+-------+
-    | Relative major/minor                                 | 0.3   |
+    | Relative major/minor (same key signature)            | 0.3   |
     +------------------------------------------------------+-------+
-    | Parallel major/minor                                 | 0.2   |
+    | Parallel major/minor (same Key)                      | 0.2   |
```
lostanlen, line 100: Key -> key ?

PRamoneda:
One moment I amediting the errors!

Thank you  @lostanlen

mir_eval/key.py
PRamoneda: the problem is that the key is the tonic note + the modality. However the tonic note is also called key. How do I distinguish between them?

PRamoneda: My idea is, for example, to use `Key` when I am referring to the full key (tonic + mode) and 'key' when it is used as a tonic note. What do you think @lostalen?

PRamoneda: BTW in that case is with k lower. You are right! Thank you!

PRamoneda: done in 
07a0d1d

mir_eval/key.py
PRamoneda: ok you are right!! @lostanlen

What do you think about this new implementation in 
6eb7bc2

lostanlen (approved):
LGTM
it would be good to have a second pair of eyes on it though

also, this PR needs a better title. It's unclear what `'x' and 'other' support` mean outside of context

PRamoneda:
Hey @lostanlen, What do you think about a title like "Support notation for unknown/ambiguous key or mode."?

lostanlen:
much better :)

PRamoneda:
I have a question. I Giant steps there are keys that have two possible values (e. g. D Major  | D minor), How would you make the evaluation?? @lostanlen

El 26 oct 2020, a las 16:39, Vincent Lostanlen <notifications@github.com<mailto:notifications@github.com>> escribió:



much better :)

lostanlen:
@PRamoneda i recommend you ask the authors for this

PRamoneda:
ok I will do that!

craffel:
I don't know the convention so I'm not sure what the correct pattern is, but it seems cleaner if we require that "X" be accompanied by a mode (major/minor/other). For the datasets you mentioned, are uncategorizable keys annotated as "X" or "X other"? It seems like a strong assumption that any "X" key should be annotated as "X other".

PRamoneda:
The keys are categorized as "X" in Giantsteps and Beatports datasets. I haven't find any example where modality were asigned to X.

BTW. Tomorrow, I will try to share a little colab for showing this PR across Giantsteps and Beatports Datasets.

craffel:
> I haven't find any example where modality were asigned to X.

Do you mean you haven't found any example where the modality was assigned to "other"? I thought X meant "uncategorizable key" and "other" meant "other modality". If there aren't any examples in any existing datasets/benchmarks where the modality  is "other", I don't think we should support that modality label.

PRamoneda:
Imean I have't found any example with X major,  X minor or X other. Alxays is "X" alone.

Beatports dataset and GiantSteps dataset have many tracks where the tonality is defined and it is followe by "other"

For example in ginat steps: 31 keys are annotated with "tonic other" (The dataset have 600 songs).

craffel:
Ok, if it's X alone, let's not use the convention where "other" gets added to X? Seems better to special-case "X" as not needing a modality.

PRamoneda:
Giant steps have 19 "X" annotations.

PRamoneda:
OK I will update the PR!

PRamoneda:
ok! done @craffel and @lostanlen

mir_eval/key.py
```diff
@@ -21,29 +21,40 @@
...
-
 KEY_TO_SEMITONE = {'c': 0, 'c#': 1, 'db': 1, 'd': 2, 'd#': 3, 'eb': 3, 'e': 4,
                    'f': 5, 'f#': 6, 'gb': 6, 'g': 7, 'g#': 8, 'ab': 8, 'a': 9,
-                   'a#': 10, 'bb': 10, 'b': 11}
+                   'a#': 10, 'bb': 10, 'b': 11, 'x': None}
 
 
 def validate_key(key):
     """Checks that a key is well-formatted, e.g. in the form ``'C# major'``.
+   The Key can be 'X' if it is not possible to categorize the Key. And mode
```
craffel, line 31: nit for grammar: combine these two sentences

mir_eval/key.py
```diff
@@ -21,29 +21,40 @@
...
-        raise ValueError("'{}' is not in the form '(key) (mode)'".format(key))
-    key, mode = key.split()
-    if key.lower() not in KEY_TO_SEMITONE:
-        raise ValueError(
-            "Key {} is invalid; should be e.g. D or C# or Eb".format(key))
-    if mode not in ['major', 'minor']:
-        raise ValueError(
-            "Mode '{}' is invalid; must be 'major' or 'minor'".format(mode))
+    if len(key.split()) != 2 \
+            and not (len(key.split()) == 1 and key.lower() == 'x'):
```
craffel, line 40: The `len(key.split()) == 1` check is superfluous, right?

mir_eval/key.py
```diff
@@ -124,6 +138,11 @@ def weighted_score(reference_key, estimated_key):
     # If keys are the same, return 1.
     if reference_key == estimated_key and reference_mode == estimated_mode:
         return 1.
+    # If reference or estimated key are x and they are not the same key
+    # then the result is 'Other'.
+    if reference_key == 'X' or estimated_key == 'X' \
+            or reference_key == 'x' or estimated_key == 'x':
```
craffel, line 144: Why not just use .lower() when doing these checks?

craffel (changes requested):
Thanks. Minor changes requested.

mir_eval/key.py
PRamoneda: mmm you are right!!!

PRamoneda:
```Python
 # If reference or estimated key is x and they are not the same key
 # then the result is 'Other'.
    if reference_key == 'x' or estimated_key == 'x'\
            or reference_key == 'X' or estimated_key == 'X':
        return 0.
```

I can not use lower() because the key can be an integer(0-11) or a string ("X")

craffel:
> ```python
>  # If reference or estimated key is x and they are not the same key
>  # then the result is 'Other'.
>     if reference_key == 'x' or estimated_key == 'x'\
>             or reference_key == 'X' or estimated_key == 'X':
>         return 0.
> ```
> 
> I can not use lower() because the key can be an integer(0-11) or a string ("X")

I don't think integer values for keys are allowed - where are you seeing that?

PRamoneda:
the Key is declared as a string in the arguments. However, after ´key, mode´conversion it converts a key into a number.... (with the header dictionary. For computing intervals.... and these things....

craffel:
> Hey, @craffel The Key is declared as a string in the arguments. However, after ´key, mode´conversion it converts a key into a number.... (with the header dictionary, for computing intervals, and these things).
> 
> Thank you so much!

But none of that happens before the if statement in question - the keys should still be strings at that point, right?

```Python
    validate(reference_key, estimated_key)
    reference_key, reference_mode = split_key_string(reference_key)
    estimated_key, estimated_mode = split_key_string(estimated_key)
    # If keys are the same, return 1.
    if reference_key == estimated_key and reference_mode == estimated_mode:
        return 1.
    # If reference or estimated key are x and they are not the same key
    # then the result is 'Other'.
    if reference_key == 'x' or estimated_key == 'x'\
            or reference_key == 'X' or estimated_key == 'X':
        return 0.
```

PRamoneda:
ok you are right! I am sorry.

PRamoneda:
@craffel  I'm sorry to be such a pain. But I don't see it.

here the keys `reference_key,  restimated_key' are converted from string to nuber if they are not 'X' and then in the if condition .lower() is not a number type function. 
```Python
reference_key, reference_mode = split_key_string(reference_key)
estimated_key, estimated_mode = split_key_string(estimated_key)
```

craffel:
> @craffel I'm sorry to be such a pain. But I don't see it.
> 
> here the keys `reference_key, restimated_key' are converted from string to number. (if they are not 'X')
> 
> ```python
> reference_key, reference_mode = split_key_string(reference_key)
> estimated_key, estimated_mode = split_key_string(estimated_key)
> ```
> 
> and then in the if condition .lower() is not a number type function. And python launch an exception

I see what you are getting at - in that case I don't see how the check works at all, since `split_key_to_string` will return None for the key if it was originally X. So, checking if it is "X" or "x" doesn't make sense, you should check if it is None according to the convention you use in KEY_TO_SEMITONE

mir_eval/key.py
```diff
@@ -124,6 +138,10 @@ def weighted_score(reference_key, estimated_key):
     # If keys are the same, return 1.
     if reference_key == estimated_key and reference_mode == estimated_mode:
         return 1.
+    # If reference or estimated key are x and they are not the same key
+    # then the result is 'Other'.
+    if not reference_key or not estimated_key:
```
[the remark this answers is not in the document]
PRamoneda: ok!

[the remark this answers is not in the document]
PRamoneda: done!

craffel (approved):

craffel:
Thanks Pedro.

PRamoneda:
Thanks to you! @craffel @lostanlen

El 30 oct 2020, a las 20:40, Colin Raffel <notifications@github.com<mailto:notifications@github.com>> escribió:



Thanks Pedro.

PR #336, opened 2021-02-13
resample_multipitch improvement

cwitkowitz:
Using the current functionality, if I try to resample the following pitch list,

image

which has no empty pitch observations (taken directly from a JAMS file), I get the following:

image

With the changes, specifying a tolerance of 50 ms, I get the following:

image

A good rule of thumb would be setting the tolerance (if you want to use it) to twice the minimum timing difference in your pitch list.

### DIFF
--- a/mir_eval/multipitch.py
+++ b/mir_eval/multipitch.py
@@ -101,9 +101,11 @@ def validate(ref_time, ref_freqs, est_time, est_freqs):
                                   allow_negatives=False)
 
 
-def resample_multipitch(times, frequencies, target_times):
+def resample_multipitch(times, frequencies, target_times, gap_tolerance=None):
     """Resamples multipitch time series to a new timescale. Values in
     ``target_times`` outside the range of ``times`` return no pitch estimate.
+    Values in ``target_times`` with a nearest neighbor ''times'' difference
+    larger than ''gap_tolerance'' will also return no pitch estimate.
 
     Parameters
     ----------
@@ -113,6 +115,8 @@ def resample_multipitch(times, frequencies, target_times):
         List of np.ndarrays of frequency values
     target_times : np.ndarray
         Array of target time stamps
+    gap_tolerance : float or None (optional)
+        Largest tolerated gap before empty estimates are inserted
 
     Returns
     -------
@@ -126,6 +130,35 @@ def resample_multipitch(times, frequencies, target_times):
         return [np.array([])]*len(target_times)
 
     n_times = len(frequencies)
+
+    if gap_tolerance is not None:
+        # Check which index boundaries have time difference larger than tolerated
+        gap_indices = (times[1:] - times[:-1]) > gap_tolerance
+        # Offset the indices by one and extract them
+        gap_indices = np.where(np.append([False], gap_indices))[0]
+
+        offset = 0
+
+        # Loop through the gaps and insert new sample(s) for padding
+        for i in range(len(gap_indices)):
+            gap = gap_indices[i] + offset
+            time_before, time_after = times[gap - 1], times[gap]
+
+            if time_after - time_before <= 2 * gap_tolerance:
+                # Only one new sample is necessary, split difference
+                new_times = [time_before + (time_after - time_before) / 2]
+                new_freqs = [np.array([])]
+            else:
+                # Add new samples at the tolerated gaps
+                new_times = [time_before + gap_tolerance, time_after - gap_tolerance]
+                new_freqs = [np.array([])] * 2
+
+            times = np.insert(times, gap, new_times)
+            frequencies = frequencies[:gap] + new_freqs + frequencies[gap:]
+
+            offset = len(frequencies) - n_times
+
+        n_times += offset
 
     # scipy's interpolate doesn't handle ragged arrays. Instead, we interpolate
     # the frequency index and then map back to the frequency values.
--- a/tests/test_multipitch.py
+++ b/tests/test_multipitch.py
@@ -77,10 +77,50 @@ def test_resample_multif0():
     actual_freqs4 = mir_eval.multipitch.resample_multipitch(
         empty_times, empty_freqs, target_times1)
 
+    times = np.array([0.00, 0.03, 0.07])
+    freqs = [
+        np.array([200.]),
+        np.array([400.]),
+        np.array([200.])
+    ]
+    target_times5 = np.array([0.00, 0.01, 0.02, 0.03,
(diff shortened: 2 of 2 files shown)


craffel:
@rabitt can you take a look?

rabitt:
@cwitkowitz good catch on this bug! I think the bug boils down to a false assumption the function makes - that the time series is uniformly spaced / time stamps with no pitches are reported with an empty frequency list.

@craffel meta-question - do we want to force multipitch time series to be uniform like we do for melody? @cwitkowitz's solution is more general and would allow non-uniform time scales, but the properties of the metrics are really different if we do/don't report silent frames.

craffel:
The high-level rule is to try to stick to external conventions, i.e. if in all past datasets/papers/competitions the multipitch time series are uniform, then we can assume that. But if there is existing convention that time series can be non-uniform, we should support that. Either way, we should document the convention. Does that make sense in this case?

rabitt:
> The high-level rule is to try to stick to external conventions, i.e. if in all past datasets/papers/competitions the multipitch time series are uniform, then we can assume that. But if there is existing convention that time series can be non-uniform, we should support that. Either way, we should document the convention. Does that make sense in this case?

I've never seen non-uniform time series for multipitch. and the mirex data from when I implemented this were uniform. Paper-wise, it's one of those details we all tend to leave out. All that said, I'd lean strongly towards uniform.

craffel:
Ok, in that case, if the code is broken for non-uniform timescales, we should throw an exception when a non-uniform timescale is passed in.

cwitkowitz:
Here is some information on my use-case. I am using GuitarSet, which has note and pitch annotations in JAMS. When evaluating MF0E performance, I would like to use the pitch annotations, as I believe they are more correct than converting the notes into discrete pitch activity during discrete times.

The pitch annotations are only specified for positive samples (i.e. times where a pitch is active). If I want to use that data with mir_eval, it needs to be converted into uniform sampling first. I thought this was a good place to put that conversion, however, I am open to other suggestions. Maybe the conversion to uniform is appropriate as its own function or somewhere in the JAMS repo.

craffel:
We could add a separate utility function that converts non-uniform time series to uniform.

cwitkowitz:
@rabitt / @craffel How does the following sound for a solution? Scrap what I have so far and replace it with a warning similar to that in the melody resampling function. This warns the user if the multipitch times are not uniform.

Then, I can create a utility function for making time-series uniform. In this function, I can estimate the true spacing by ignoring outliers and taking the mean of the differentials (something to that effect). Then I can re-project the original data onto the new collection of times, which would start at 0 and be spaced by the estimated spacing.

Still, this would not always handle sporadic data very well, e.g. if my positive samples are not spaced uniformly or have no consistent sampling in the first place. However, in the case of JAMS data, where pitch contours should be aligned with some sampling period, I believe it would work just fine.

It might also make sense to put the utility function in the JAMS repo, since JAMS pitch_contour seems to be the only case where this has ever been encountered.

craffel:
This seems like a reasonable solution to me at a high level.

cwitkowitz:
> This seems like a reasonable solution to me at a high level.

Great. I can get started on this then.

rabitt:
@cwitkowitz If you haven't gotten too far - I have a [old script](https://github.com/marl/medleydb/blob/master/scripts/fill_tony_file.py) that does exactly this (filling in missing time stamps) if you want to use it as a starting point.

cwitkowitz:
@rabitt - The script you linked would certainly work, however it requires that the sampling rate and hop length are known. In my case these values are not explicitly defined.

I went a little down the path of implementing what I mentioned above, i.e. estimating the hop length by taking the mean of first-order differentials wherever the second-order differential was close to zero. I wasn't sure what to do at that point. I can step from times[0] to times[-1] using this estimated hop length, but it doesn't necessarily line up perfectly, and it seems we would need 1d interpolation again. There is also no knowledge of the duration, so we can fill in empties from t=0 until times[0], but we cannot fill in empties from times[-1] to t=duration.

Any thoughts?

rabitt:
@cwitkowitz 

> I went a little down the path of implementing what I mentioned above, i.e. estimating the hop length by taking the mean of first-order differentials wherever the second-order differential was close to zero. I wasn't sure what to do at that point. I can step from times[0] to times[-1] using this estimated hop length, but it doesn't necessarily line up perfectly, and it seems we would need 1d interpolation again. 
> There is also no knowledge of the duration, so we can fill in empties from t=0 until times[0], but we cannot fill in empties from times[-1] to t=duration.

Maybe it's safer to have `hop_length` and `duration` as arguments to the script - presumably that information should be part of the datasets the data comes from, and then there's no room for error. And maybe you could create a separate helper function called `estimate_hop_length` that returns a number that the user can use/not?

cwitkowitz:
@rabitt - how does this look? I can add test cases once we are happy with the changes.

mir_eval/util.py
```diff
@@ -956,3 +957,79 @@ def midi_to_hz(midi):
...
+    # Make sure the times are sorted
+    times = np.sort(times)
+
+    # Estimate the hop in seconds
+    non_gaps = np.append([False], np.isclose(np.diff(times, n=2), 0))
+
+    if not np.sum(non_gaps):
+        raise ValueError("Time observations are too irregular.")
+
+    return np.diff(times)[non_gaps].mean()
```
rabitt, line 985: I think `.median()` would be more stable, since in theory the spacing should be uniform. Then one or two outliers won't skew the hop size, which will over time propagate the error

rabitt (approved):
Looks great! One small comment on the hop size estimation

cwitkowitz:
Good suggestion @rabitt 

I made that change and added a few other small changes necessary to get the test cases working.

Here a a few final considerations before completing the PR:
1. NumPy mean() and median() function were not always giving me exact values, which is why I needed the A_TOL comparison rather than strict equality.
2. The time_series_to_uniform() function works only for observations specified by a list of ndarrays. This is fine for my purposes, but I am not sure if we would like the function to be more general for mir_eval as a whole.
3. In my test cases, I needed to compare frequencies, so I used repeat code from test_multipitch.py. I did not see any current examples where code within tests/ was being reused, so I wasn't sure if we wanted to start doing that or just ignore it for now.

cwitkowitz:
I think the best solution here is simply to add an extra frame in all cases, even if this means the nearest frame on the uniform time grid for the last observation may occur slightly after the specified duration.

PR #338, opened 2021-03-16
Add more metrics to transcription_velocity.evaluate

cghawthorne:
Also make the metric names a little more clear given all these different combinations.

I updated the path for the test json files because they didn't seem to be loading before, but let me know if I just wasn't running the test correctly.


craffel:
Thanks Fjord. We generally try to avoid breaking changes in minor version changes (http://semver.org), and I think changing the names of existing metrics would be a breaking change. I think it could make sense to add additional metrics to `evaluate`, which I would consider backwards-compatible. One option in the meantime would be for users to define a custom `evaluate` function.

cghawthorne:
Good points. I think for now, we'll just define our own evaluate function. Thanks!

Issue #337, opened 2021-03-02
perfect 5th detection in mir_eval.key is asymmetric


iansimon:
https://github.com/craffel/mir_eval/blob/576aad4e0b5931e7c697c078a1153c99b885c64f/mir_eval/key.py#L145

Is this intended to be asymmetric?  E.g. if estimated = G major and reference = C major the score is 0.5, but if estimated = C major and reference = G major the score is zero.

craffel:
I do not know music theory :( @bmcfee ?

bmcfee:
From reading the comments, I think this is behaving to spec / consistent with the mirex implementation. 

I also think it's weird, but 🤷

bmcfee:
To follow up on this a bit, yes, it's intended to be asymmetric, and the docstring is unambiguous about the behavior: 

https://github.com/craffel/mir_eval/blob/576aad4e0b5931e7c697c078a1153c99b885c64f/mir_eval/key.py#L99-L115

I suppose this was originally intended to break symmetry between 5th and 4th estimation errors by only considering movement in one direction from the root.  I did try to dig into this more in the mirex wiki, but there's not much provenance around how this metric came about.  Maybe @jpauwels has some insight, as I think he was the most recent task caption for key detection?

jpauwels:
Hi all, as far as I know, it was never intended to make a distinction between perfect fifth and perfect fourth errors (as long as the mode is the same). It certainly isn't the case in the current implementation used for MIREX. That doesn't mean it's impossible that there is some ambiguously worded description somewhere on the wiki. Are you referring to the following table: https://www.music-ir.org/mirex/wiki/2005:Audio_and_Symbolic_Key (copied over every year since)? In that case, it should be read as perfect fifth either way, not just perfect fifth above.

The canonical reference I use for this measure is Emilia Gomez's PhD thesis (section 4.3.1.2), but I suspect the exact weighting was determined collaboratively with other participants of MIREX 2005. Between us, I've never been a fan of this particular measure, too many magic numbers in there. I always suggest listing the error categories separately and did so in my own papers, but I still include this combination measure in MIREX for historical comparisons.

stefan-baumann:
Seconding what @jpauwels said, he actually stated that in the 2017 MIREX results, too:
[...]

PR #339, opened 2021-07-23
[WIP] Implement current MIREX key scoring method and warn if using old one

stefan-baumann:
This PR aims to solve the issue discussed in #337 in a backwards-compatible manner.
An additional argument `allow_descending_fifths` was added to the `key.weighted_score` and `key.evaluate` methods, which is set to `False` by default (and thus being backwards-consistent), but raises a warning stating that this method is no longer used for MIREX. If set to `True`, descending fifth errors are given a score of 0.5, matching current MIREX scoring.

I'd love to get feedback on whether the maintainers approve of the way I implemented this change and whether they might have ideas how to improve keyword naming or the added documentation.

If the feedback is positive, I'll extend the tests to cover this behaviour as well.

### DIFF
--- a/mir_eval/key.py
+++ b/mir_eval/key.py
@@ -19,6 +19,7 @@ Metrics
 '''
 
 import collections
+import warnings
 from . import util
 
 KEY_TO_SEMITONE = {'c': 0, 'c#': 1, 'db': 1, 'd': 2, 'd#': 3, 'eb': 3, 'e': 4,
@@ -96,7 +97,8 @@ def split_key_string(key):
     return KEY_TO_SEMITONE[key.lower()], mode
 
 
-def weighted_score(reference_key, estimated_key):
+def weighted_score(reference_key, estimated_key,
+                   allow_descending_fifths=False):
     """Computes a heuristic score which is weighted according to the
     relationship of the reference and estimated key, as follows:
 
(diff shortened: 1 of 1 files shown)


craffel (approved):
I think this is a reasonable approach, but we should plan to switch the default to `True` eventually if it matches the new convention.

craffel:
@bmcfee @jpauwels @iansimon PTAL

stefan-baumann:
> but we should plan to switch the default to True eventually if it matches the new convention

In that case, we should probably change the warning message to reflect that. Maybe something like this?
> The selected key scoring method does not match that currently used by MIREX. To use the same method, specify allow_descending_fifths=True. The default behaviour will change to allow_descending_fifths=True in the future.

craffel:
> > but we should plan to switch the default to True eventually if it matches the new convention
> 
> In that case, we should probably change the warning message to reflect that. Maybe something like this?
> 
> > The selected key scoring method does not match that currently used by MIREX. To use the same method, specify allow_descending_fifths=True. The default behaviour will change to allow_descending_fifths=True in the future.
> 
> Potentially substituting `in the future` with a specific version number.

SGTM

stefan-baumann:
Okay, I added the note to the warning and the docstring, specifying that the default behaviour will change `in the future`.

PR #343, opened 2022-02-08
Add suite of alignment metrics

f90:
Hey,

I coded up a set of alignment metrics I would like to see added in mir_eval.

To specify the scope: This can be used to evaluate alignment models which are given a sequence of events along with a piece of audio, and then return a sequence of timestamps, with one timestamp for each event, indicating the position of this event in the audio. 
The events are listed in order of occurrence in the audio, so that output
timestamps have to be monotonically increasing.

Why a new module for mir_eval for this and not reuse an existing one? Well, this type of alignment differs from other event detection setups: Onset, beat, segment detection can have the model predict too many or too few events compared to the ground truth, and sometimes requires additional labelling. In contrast, for alignment the events to be detected are part of the input to the model, and evaluation is focused on pair-wise comparison between each output and reference timestamp for each event.

I implemented all metrics currently used for lyrics alignment in particular:
- Mean absolute error
- Median absolute error
- Percentage of correctly detected timestamps
- Percentage of correct segments
- A perceptually motivated metric as provided by @mmoussallam to my [jamendolyrics](https://github.com/f90/jamendolyrics) repo

If there are some other metrics missing for related tasks that have the same set of assumptions (maybe audio to score alignment), we could also add them here.


craffel:
Thanks Dan. Are these metrics published/used anywhere? We generally restrict ourselves to metrics that are somehow "standard", which typically means implemented in MIREX, but more broadly means that they've been published somewhere and used in at least one paper.

f90:
@craffel In the MIREX lyrics alignment challenge all the above mentioned metrics are used, see here for the 2020 round:
https://www.music-ir.org/mirex/wiki/2020:Automatic_Lyrics-to-Audio_Alignment_Results

The exception to this is the perceptual metric, but this was proposed in the following ISMIR paper:
N. Lizé-Masclef, A. Vaglio, M. Moussallam, ISMIR 2021

Maybe we should add a disclaimer for this metric that the perceptual quality here is tailored to a Karaoke context in particular and therefore should probably not be used for other alignment problems such as audio-to-score?

Unfortunately there doesn't exist a clear reference implementation for any othese metrics that is both very stable/bug-free and also commonly used - which is why I wanted to contribute this. The closest contestant would be the MIREX alignment code in this repo (which does not have the perceptual metric):

https://github.com/georgid/AlignmentEvaluation

However it is not straightforward to run properly and relies on input files of specific formats. Also it models timestamps as (start, end, token) triplets, which makes things more complicated than they should be, since the token IDs for the events don't matter and if you want to only evaluate start timestamps for example, it is a bit inflexible since you need to generate end timestamps based on the start timestamp of the next token. At the same time, this distinction is not really needed to compute the metrics properly. So in this implementation, you can feed in either only start or both start and end timestamps (as a 1D numpy array [start1, end1, start2, end2, ...] and calculate the metrics for each case if you want.

craffel:
Ok, got it. Can you test against the MIREX metrics so that we can confirm the implementations match? And can you add references as appropriate (as seen e.g. [here](https://github.com/craffel/mir_eval/blob/master/mir_eval/hierarchy.py#L31))

f90:
OK, can do. Should I create additional fixture files to the ones I already added that contain the outputs from the MIREX implementation? Or just test whether I get the same results for the current fixture files locally on my end, without modifying the PR at all?

craffel:
Well, it sort of depends - if your implementation matches MIREX exactly, it certainly doesn't hurt to add those as test cases too, as long as your metric is not too expensive to evaluate. If it doesn't match MIREX, and you're confident this is because you're doing things correctly and MIREX has a bug (as we found for some metrics in the [original paper](https://colinraffel.com/publications/ismir2014mir_eval.pdf)), then they shouldn't be added as test cases.

f90:
I added the only test case that was included in the MIREX eval code as regression tests into this PR - with the caveat that i had to transform the (start, end) inputs into a 1D list of timestamps first, and deleted an entry that is not counted during evaluation since it is an additional timing for a previous segment, for which no ground truth exists. Only after that I have an equal number of reference and estimate timestamps.

I get idential results for the metrics for both this PR and the MIREX repo, except for one detail in the PCS metric. Here the beginning silence before the first event is detected was not counted as a correct segment in the MIREX code, which is not the correct behaviour, since the metric is just supposed to give the percentage of time of segment overlaps, and the initial segment before the first detected event is also a segment to be counted. So for this metric I added the values calculated by the new `pcs` method instead to the fixture files.

I also added the relevant references.

Please have a look and see if it looks ready now :) Thank you

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    if not np.all(reference_timestamps[1:] - reference_timestamps[:-1] >= 0):
+        raise ValueError("Reference timestamps are not monotonically increasing!")
+    if not np.all(estimated_timestamps[1:] - estimated_timestamps[:-1] >= 0):
+        raise ValueError("Estimate timestamps are not monotonically increasing!")
+
+    # Check positivity (need for correct PCS metric calculation)
+    if not np.all(reference_timestamps >= 0):
+        raise ValueError("Reference timestamps can not be below 0!")
+    if not np.all(estimated_timestamps >= 0):
+        raise ValueError("Estimate timestamps can not be below 0!")
```
craffel, line 106: Estimated

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+import numpy as np
+from scipy.stats import skewnorm
+
+from mir_eval.util import filter_kwargs
+
+
+def validate(reference_timestamps: np.ndarray, estimated_timestamps: np.ndarray):
+    """Checks that the input annotations to a metric look like valid onset time
+    arrays, and throws helpful errors if not.
+    Parameters
```
craffel, line 59: Add line above Parameters

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    # Check positivity (need for correct PCS metric calculation)
+    if not np.all(reference_timestamps >= 0):
+        raise ValueError("Reference timestamps can not be below 0!")
+    if not np.all(estimated_timestamps >= 0):
+        raise ValueError("Estimate timestamps can not be below 0!")
+
+
+def ae(reference_timestamps, estimated_timestamps):
+    """Compute the absolute deviations between predicted and reference timestamps,
+    and then takes the median or average over all events
```
craffel, line 111: "takes the median or average over all events" -> "return the median and average over all events."

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    >>> estimated_timestamps = mir_eval.io.load_events('estimated.txt')
+    >>> mae, aae = mir_eval.align.ae(reference_onsets, estimated_timestamps)
+
+    Parameters
+    ----------
+    reference_timestamps : np.ndarray
+        reference timestamps, in seconds
+    estimated_timestamps : np.ndarray
+        estimated timestamps, in seconds
+    Returns
```
craffel, line 125: Line above Returns

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+
+    Parameters
+    ----------
+    reference_timestamps : np.ndarray
+        reference timestamps, in seconds
+    estimated_timestamps : np.ndarray
+        estimated timestamps, in seconds
+    window : float
+        Window size, in seconds
+        (Default value = .1)
```
craffel, line 156: Looks like the default is 0.3

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    Returns
```
craffel, line 157: Line above Returns

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+
+def pcs(reference_timestamps, estimated_timestamps, duration: float):
+    """Constructs segments out of predicted and estimated timestamps separately
+    out of each given timestamp vector with entries (t1,t2, ... tN), yielding segments with
+    the following (start, end) boundaries: (0, t1), (t1, t2), ... (tN, duration).
+    The metric then calculates the percentage of overlap between correct segments compared to the
+    total duration.
+
+    This metric is based on the paper
+    "LyricSynchronizer: Automatic synchronization system between musical audio signals and lyrics"
```
craffel, line 175: Please add a Sphinx ref here (like https://github.com/craffel/mir_eval/blob/master/mir_eval/separation.py#L366)

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+
+    Parameters
+    ----------
+    reference_timestamps : np.ndarray
+        reference timestamps, in seconds
+    estimated_timestamps : np.ndarray
+        estimated timestamps, in seconds
+    duration : float
+        Total duration of audio (seconds) - this is needed to compute the final segments properly!
+    Returns
```
craffel, line 192: Line above Returns

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    -------
+    pcs : float
+        Percentage of time where ground truth and predicted segments overlap
+    """
+    validate(reference_timestamps, estimated_timestamps)
+    duration = float(duration)
+    assert np.max(reference_timestamps) <= duration, (
```
craffel, line 199: Use if statement and ValueError instead of assert

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+        "Expected largest reference timestamp "
+        f"{np.max(reference_timestamps)} to not be "
+        f"larger than duration {duration}"
+    )
+    assert np.max(estimated_timestamps) <= duration, (
```
craffel, line 204: Use if statement and ValueError instead of assert

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    est_starts = np.concatenate([[0], estimated_timestamps])
+    est_ends = np.concatenate([estimated_timestamps, [duration]])
+    overlap_starts = np.maximum(ref_starts, est_starts)
+    overlap_ends = np.minimum(ref_ends, est_ends)
+    overlap_duration = np.sum(np.maximum(overlap_ends - overlap_starts, 0))
+    return overlap_duration / duration
+
+
+def perceptual_synchrony(offset):
+    """function perceptually weighting the deviations of onsets"""
```
craffel, line 221: Please mention the ref for these values in this docstring too.

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    normalisation_factor = 1.6857
+    return (1.0 / normalisation_factor) * skewnorm.pdf(
+        offset, skewness, loc=localisation, scale=scale
+    )
+
+
+def perceptual_metric(reference_timestamps, estimated_timestamps):
+    """metric based on human synchronicity perception as measured in the paper
+
+    "User-centered evaluation of lyrics to audio alignment",
```
craffel, line 234: Please add a sphinx ref.

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    localisation = -0.22270315
+    scale = 0.29779424
+    normalisation_factor = 1.6857
+    return (1.0 / normalisation_factor) * skewnorm.pdf(
+        offset, skewness, loc=localisation, scale=scale
+    )
+
+
+def perceptual_metric(reference_timestamps, estimated_timestamps):
+    """metric based on human synchronicity perception as measured in the paper
```
craffel, line 232: Metric

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    >>> estimated_timestamps = mir_eval.io.load_events('estimated.txt')
+    >>> perceptual = mir_eval.align.perceptual_metric(reference_onsets, estimated_timestamps)
+
+    Parameters
+    ----------
+    reference_timestamps : np.ndarray
+        reference timestamps, in seconds
+    estimated_timestamps : np.ndarray
+        estimated timestamps, in seconds
+    Returns
```
craffel, line 255: Line above Returns

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    -------
+    perceptual_score : float
+        Perceptual score, averaged over all timestamps
+    """
+    validate(reference_timestamps, estimated_timestamps)
+    perceptual_score = perceptual_synchrony(estimated_timestamps - reference_timestamps)
```
craffel, line 261: Does this really need its own function if this is the only place it's called?

mir_eval/alignment.py
```diff
@@ -0,0 +1,300 @@
...
+    return np.mean(perceptual_score)
+
+
+def evaluate(reference_timestamps, estimated_timestamps, duration: float, **kwargs):
+    """Compute all metrics for the given reference and estimated annotations.
+    Examples
+    --------
+    >>> reference_timestamps = mir_eval.io.load_events('reference.txt')
+    >>> estimated_timestamps = mir_eval.io.load_events('estimated.txt')
+    >>> duration = max(np.max(reference_timestamps), np.max(estimated_timestamps)) + 10
```
craffel, line 271: Where does this 10 come from? Is this really standard?

tests/data/alignment/est00.txt
```diff
@@ -0,0 +1,11 @@
+# A comment
```
craffel, line 1: No need to have a comment here

craffel (changes requested):
Mostly minor comments. Higher-level stuff:

1. Please check your code against PEP-8 (in particular, enforce a line length of 80 char).
1. I think we should give these metrics more descriptive names (i.e. not abbreviations), unless they are widely known by the abbreviations. For example, I can't figure out what `ae` stands for.

mir_eval/alignment.py
f90: Probably not, integrated it into the main function which was very empty anyway

mir_eval/alignment.py
f90: 10 is just an arbitrary number here. the duration should be set to the total duration of the audio, so the PCS metric can be correctly computed. should i change this to something like
`audio, sr = librosa.load(audio_path, mono=True)
duration = float(len(audio)) / float(sr)`

or simply mention in the description above that duration needs to be the audio duration?

bmcfee: no, librosa should not be a dependency of mir_eval.  We usually infer duration from the reference annotation so that it can be consistently applied across all estimates.

f90: Ah, maybe this is a misunderstanding but I suggest this only as example code in the doc part of the function, not as actual implementation. If we want to keep librosa completely out of this, the example code in the doc of this function could also be
```
    >>> reference_timestamps = mir_eval.io.load_events('reference.txt')
    >>> estimated_timestamps = mir_eval.io.load_events('estimated.txt')
    >>> pcs = mir_eval.align.pcs(reference_onsets, estimated_timestamps, duration)
```
with the hint that duration needs to be provided by the user and indicates the total song length.

Btw, the reason why I am not including the duration as another timestamp in the `reference_timestamps` and `estimated_timestamps` is because it creates logic branching in most of the other metrics, where you would then want to ignore it, since it's technically not a timestamp that needs to be predicted. for example, for the percentage of correctly detected timestamps, it doesn't make sense to include the end of the song as a timestamp to be predicted. PCS is a bit special in this sense, for computing the segment overlap you need the end position of the last segment, which is the end of the song.

So hopefully this clears up some things about the metrics. Do you agree that changing the example code in the docs like i suggested is the best way to go about this?

f90:
Thanks for reviewing! Changed the metrics names to something more self-explanatory. Also addressed all the smaller comments and made it 80 chars per line everywhere. 
We only need to resolve the issue around using/getting the total song duration for the percentage_correct_segments (PCS) metric, see the discussion above for details.

mir_eval/alignment.py
craffel: I don't know if this is your problem to solve, but it's problematic to me that the annotation format for this task is incomplete. If the song duration is necessary to perform evaluation, then the annotation format should include the duration. This is definitely a red flag for evaluation inconsistencies in the future - let's say you load your audio files with `librosa` to get the duration; I load them with some other package/binary that always appends 1s of silence (sillier things have happened). Now our scores won't be comparable, even though we are using the same annotations.

How is this handled in MIREX and in any other packages? I feel either the duration should be included in the annotation, or there should be some convention (e.g. duration = 10s after last event, though this seems arbitrary and weird).

craffel: We generally err towards community consensus with `mir_eval`. Consensus can be implicit, based on annotation formats. Are there a large number of annotations of songs available in the "1D list"-style? In that case, and given that MIREX doesn't use a duration, I would not suggest we create a new annotation format (and generally agree that it's more work than is warranted). WDYT of this solution: make the duration parameter optional (default to None); when not passed, do the MIREX-style evaluation; when passed, do what you are currently doing. Then add a note mentioning the MIREX convention and the "right" way and leave it up to users. In the evaluate docstring, you can just not pass in a duration, etc.

f90: Yes, I would say at least for lyrics alignment, mostly the annotation files are also in this 1D array format, and usually don't specify the audio duration in it. 
I think that solution sounds good! This way the `evaluate` metric also runs out of the box without needing the duration. And then the community can still switch to the other version of the metric if it deems that one really as better, but we don't need to decide that ourselves.
Will implement this now!

craffel: 👍

f90:
OK, the changes to the percentage of correct segments are implemented as you suggested, please have another look :)

mir_eval/alignment.py
craffel: Ping on this one.

mir_eval/alignment.py
craffel: Ping on this one too.

mir_eval/alignment.py
f90: Function doesn't exist anymore so that's resolved

f90:
Ah sorry, overlooked that there were still comments in the change request. I worked them all out now!

craffel:
Thanks! Bug me if you want me to get a new release on PyPI...

craffel (approved):

f90:
Thanks for working with me on this! It would be great if you could push out a new version in the next few days since I am working on multiple projects that need these metrics, so I can stop duplicating code across projects

craffel:
Done! Let me know if there are any issues... has been a few  years since the last release.
https://pypi.org/project/mir_eval/

Issue #344, opened 2022-04-13
Docs don't list the new alignment metrics


f90:
Hey Colin!

Hope you're doing well. I noticed that the alignment metrics that I contributed (#343) are not yet showing up in the mir_eval documentation - do you need to manually regenerate the docs so they become visible? Since it's a new module people would otherwise not notice it at all, I suppose.

https://craffel.github.io/mir_eval/#

craffel:
Hm, looks like I attempted to update the docs to 0.7 (https://github.com/craffel/mir_eval/commit/6906a2f9717e6cdff972b8924dca4977f6f7c547) but failed. Let me try again.

f90:
Got it, please have a look at PR #345 and see if that does it

craffel:
Looks good. The tricky thing is that I would normally wait until the next release before re-generating the docs, but I assume you want them online before then?

f90:
I remember you mentioned that there was quite a lot of time between the current version release and the last one, so it would be good to not have to wait long (like more than a 1-2 weeks). So if this generation doesn't have any annoying side effects when doing it independent of a release, it would be good to regenerate soon. But if you're very busy right now, we can also delay for a while of course 👍

craffel:
K, updated them manually: https://craffel.github.io/mir_eval/#mir-eval-alignment !

PR #345, opened 2022-04-14
Add docs for alignment metrics

f90:
Apparently the alignment functions don't show up in the docs yet. I added the module to index.rst
Please check if the docs generate properly now and merge in that case. Thanks!


f90:
This addresses issue #344

PR #353, opened 2023-05-24
Set precision for np.int and np.float

carlthome:
As per https://numpy.org/devdocs/release/1.24.0-notes.html#expired-deprecations these aliases are deprecated and now cause errors.

### DIFF
--- a/mir_eval/beat.py
+++ b/mir_eval/beat.py
@@ -386,14 +386,14 @@ def p_score(reference_beats,
     estimated_beats = np.array(estimated_beats - offset)
     reference_beats = np.array(reference_beats - offset)
     # Get the largest time index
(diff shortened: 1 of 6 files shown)


bmcfee (commented):
This basically looks fine to me.  The only quibble I would have is that we could be a little more careful with the precision here.  For example, `np.int64` is overkill for most uses, and in some cases we'd actually prefer `uint` types (eg if it's do be used as an array index).

For the chord metrics, the root/bass/semitones are practically either boolean or integers less than 12.  A uint8 would be more appropriate here.

This mostly doesn't matter, but we may as well be efficient if possible.

craffel:
Backwards-compatibility on the first pass seems reasonable to me - will merge. Thanks both.

PR #355, opened 2023-06-08
Optimise  & fix sonification

leighsmith:
- Added test for only sonifying those frequency bands which have energy.
- Corrected sonification `time_frequency()` t_max parameter to match time units in seconds when adjusting intervals, not samples.
- Corrected off-by-one comparison of time centers vs. times, further optimised computations in time_frequency.


setup.py
```diff
@@ -5,7 +5,7 @@
 
 setup(
     name='mir_eval',
-    version='0.7',
+    version='0.8',
```
bmcfee, line 8: Please revert this change - version bumps should happen in a dedicated PR

mir_eval/sonify.py
```diff
@@ -147,34 +154,42 @@ def __interpolator(x):
...
     time_centers = np.mean(times, axis=1) * float(fs)
 
+    # Not really a true spectral energy, just a summation for optimisation.
+    spectral_sums = np.sum(gram, axis = 1)
     for n, frequency in enumerate(frequencies):
+        if spectral_sums[n] < 0.01: # TODO set threshold intelligently.
+            continue
         # Get a waveform of length samples at this frequency
         wave = _fast_synthesize(frequency)
+        # print(f"Freq {n} {frequency}Hz energy {spectral_sums[n]}, length of wave {len(wave)}")
```
bmcfee, line 173: Please remove commented debug prints

mir_eval/sonify.py
```diff
@@ -147,34 +154,42 @@ def __interpolator(x):
...
     gram = np.maximum(gram, 0)
 
     # Pre-allocate output signal
     output = np.zeros(length)
     time_centers = np.mean(times, axis=1) * float(fs)
 
+    # Not really a true spectral energy, just a summation for optimisation.
+    spectral_sums = np.sum(gram, axis = 1)
     for n, frequency in enumerate(frequencies):
+        if spectral_sums[n] < 0.01: # TODO set threshold intelligently.
```
bmcfee, line 169: If we're going to have a threshold, it should probably be either parametrized or relative to the values in `gram`

mir_eval/sonify.py
```diff
@@ -103,11 +103,18 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None,
...
-
-    # Truncate times so that the shape matches gram
-    n_times = gram.shape[1]
-    times = times[:n_times]
+    last_time_in_secs = float(length) / fs
+    times, _ = util.adjust_intervals(times, t_max=last_time_in_secs)
+
+    # Truncate times so that the shape matches gram. However if the time boundaries were converted
+    # to intervals, then the number of times will be reduced by one, so we only truncate
+    # if the gram is smaller.
```
bmcfee, line 111: This could be more compact:

```python
n_times = min(gram.shape[1], times.shape[0])
times = times[:n_times]
```

mir_eval/sonify.py
```diff
@@ -103,11 +103,18 @@ def time_frequency(gram, frequencies, times, fs, function=np.sin, length=None,
     if length is None:
         length = int(times[-1, 1] * fs)
 
-    times, _ = util.adjust_intervals(times, t_max=length)
-
-    # Truncate times so that the shape matches gram
-    n_times = gram.shape[1]
-    times = times[:n_times]
+    last_time_in_secs = float(length) / fs
+    times, _ = util.adjust_intervals(times, t_max=last_time_in_secs)
```
bmcfee, line 107: Good catch on this!  Classic dimensional analysis error on my part, using sample indices in place of time measurements.

While trying to unpack what happened here, I noticed that the docstring isn't as explicit as it could be that `times` should be in units of seconds.  Could you add some clarifying text there?

mir_eval/sonify.py
```diff
@@ -147,34 +154,42 @@ def __interpolator(x):
...
+            continue
         # Get a waveform of length samples at this frequency
         wave = _fast_synthesize(frequency)
+        # print(f"Freq {n} {frequency}Hz energy {spectral_sums[n]}, length of wave {len(wave)}")
 
         # Interpolate the values in gram over the time grid
         if len(time_centers) > 1:
             gram_interpolator = interp1d(
-                time_centers, gram[n, :],
+                time_centers, gram[n, :-1],
```
bmcfee, line 178: why are we now off by one here?

mir_eval/sonify.py
```diff
@@ -147,34 +154,42 @@ def __interpolator(x):
...
+        # Create the time-varying scaling for each time interval by the piano roll magnitude
+        for m, (start, end) in enumerate(sample_intervals):
             # Clip the timings to make sure the indices are valid
             start, end = max(start, 0), min(end, length)
-            # add to waveform
-            output[start:end] += (
-                wave[start:end] * gram_interpolator(np.arange(start, end)))
+            scaler[start:end] = gram_interpolator(np.arange(start, end))
+        # add to waveform
+        output += wave[:length] * scaler
```
bmcfee, line 192: This look incorrect - why has the output accumulation moved outside the loop?  Won't this only include the final `scaler` array?

bmcfee (changes requested):

carlthome:
Any updates on this? Would love to pitch in as needed to make Windows builds pass.

setup.py
leighsmith: I had bumped the version number in order to allow distinguishing the locally modified version from the actual distributed v0.7 version, when running pip or poetry. The changed version then represents the `main` branch as v0.8 as an "aspirational goal", before it's a tagged release. But I've reverted the change to match the version tagging strategy used.

mir_eval/sonify.py
leighsmith: Removing diagnostics removes the ability to build on the expected semantics reflected in the diagnostic messages to aid future debugging. Effectively each developer has to begin diagnostics and understand the code from scratch. But I've removed them in answer to your request.

mir_eval/sonify.py
leighsmith: As the diagnostic code displays, and the comment explains, the truncation of `times`, after adjusting the intervals, should only be done if the `times` length exceeds the `gram` length. This would lead to an unnecessary truncation operation if the `gram` length equals or exceeds the `times` length. Whether that truncation operation is performed or becomes a no-op would depend on the implementation of numpy.

mir_eval/sonify.py
leighsmith: Added documentation string updates.

mir_eval/sonify.py
leighsmith: When converting `times` from boundaries to intervals with `utils.boundaries_to_intervals()`, this will result in a reduction in length of `times` by 1, which needs reconciling with the length of `gram`. Historically, the bug in using `length` to adjust the intervals left a very long last interval, which covered up the difference in length.

mir_eval/sonify.py
leighsmith: In the original code, `output` was only being updated over the region `start:end` within the loop, performing the multiplication with only minimal benefit from vector multiplication, needing to iterate that `len(sample_intervals)` times. This is an optimisation where the multiplication is applied across the whole vector once, after the scaling is assembled from the `gram_interpolator()` call.

mir_eval/sonify.py
bmcfee: > This would lead to an unnecessary truncation operation if the `gram` length equals or exceeds the `times` length. Whether that truncation operation is performed or becomes a no-op would depend on the implementation of numpy.

I'm not sure I understand your comment here.  `x[:x.shape[0]]` is well-defined in the numpy API, and would not invoke a copy, nor would it have side-effects at the call site since `gram` is a reference.

In general, this kind of thing should always produce a view instead of a copy, and is essentially free:
```python
In [1]: import numpy as np

In [2]: x = np.arange(10)

In [3]: y = x[:x.shape[0]]

In [4]: y.flags
Out[4]: 
  C_CONTIGUOUS : True
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
```
(note the `OWNDATA` flag being false there.)

mir_eval/sonify.py
bmcfee: Sorry, still not following this.  But I think it would be clearer if instead of `:-1`, it explicitly matched the length of `time_centers`, which should be exactly `n_times` right?  Explicit > implicit and all that.

mir_eval/sonify.py
bmcfee: Ok, I see now.  It won't be exactly the same as the original code in case we have intervals that overlap, but that's unsupported behavior anyway so not a big deal.



> his is an optimisation where the multiplication is applied across the whole vector once, after the scaling is assembled from the `gram_interpolator()` call.

I wonder if we could push this idea further?  We can think of this as applying the interpolator to different slices of the full sample range `np.arange(length)`.  I bet we could reorganize this, if we wanted to, to only use one call to the interpolator here, and only use the loop over intervals to construct the input sample positions that will be included.  This probably isn't a worthwhile optimization, but ought to be possible.

mir_eval/sonify.py
leighsmith: OK, fair comment, I've updated the code to your suggested improvements.

mir_eval/sonify.py
leighsmith: Fair comment, I've now done that.

mir_eval/sonify.py
leighsmith: Yes good point. I've now simplified it to call the interpolator once per frequency bin, and indeed, it does do worthwhile optimization, in my measurements, cutting execution time in half.

carlthome:
Do we consider this PR a blocker for an updated PyPI release? Related: https://github.com/craffel/mir_eval/issues/366

bmcfee:
Sorry, lost the thread on this one over the summer.  I think it's mostly fine now except for that threshold point?

Issue #354, opened 2023-06-07
Migrate pull request automated testing from Travis CI to GitHub Actions


carlthome:
To help make sure mir_eval stays in good shape as its dependencies moves, it would be convenient to have test automation for PRs to make reviews easier. Time flies, Travis CI seems broken (?), and the world has moved to GitHub Actions.

I'm wondering if you'd like to have automated testing by a .github/workflows/ci.yaml - if so, I could draft a proposal in a PR (since I happen to have professional work hours to spend on helping out currently).

👍 / 👎 ❓

bmcfee:
This would be great, and long overdue.

In setting this up, I would also advocate for breaking up some of the CI functionality across different workflows, so that there are separate actions for unit tests, linting, and doc testing.

PR #356, opened 2023-06-09
Add automated unit testing on pull requests with GitHub Actions

carlthome:
### What?
Introduce automated testing on PRs by GitHub Actions.

### Why?
To make PR reviews easier by checking that unit tests pass automatically on fresh installations of mir_eval, as discussed in https://github.com/craffel/mir_eval/issues/354


bmcfee:
> Getting a error about missing `decorator`: https://github.com/craffel/mir_eval/actions/runs/5222270760/jobs/9427537252#step:7:27

Looks like a dependency problem with the decorator package.  In general, I think we'd do best to set up a proper miniconda environment here, rather than relying on pip.

carlthome:
> In general, I think we'd do best to set up a proper miniconda environment here, rather than relying on pip.

Can definitely update this to follow https://github.com/librosa/librosa/blob/main/.github/workflows/ci.yml instead but think there's value in getting a standrad `pip` to work since `conda` is a bit more involved and harder to convince people to use.

carlthome:
yield based tests are skipped by pytest: https://github.com/craffel/mir_eval/actions/runs/5236939084/jobs/9454797085#step:5:215

bmcfee:
> but think there's value in getting a standard `pip` setup to work since `conda` is a bit more involved and harder to convince people to use.

Distributing via conda (convincing people to use it) and using conda for internal testing environments are two different things.  I was only advocating for the latter here, and the librosa example you point to actually uses pip to install the package being tested.



> yield based tests are skipped by pytest:

Yes - migrating from nose to pytest is going to be a pretty hefty endeavor for this reason, and I think should be handled in a separate PR.

carlthome:
> Distributing via conda (convincing people to use it) and using conda for internal testing environments are two different things. I was only advocating for the latter here, and the librosa example you point to actually uses pip to install the package being tested.

Got it! I'll update this PR to use miniconda accordingly.

carlthome:
Initial conda done! 🐍 

Changing from `actions/setup-python` to `conda-incubator/setup-miniconda` made build times go up quite a bit however:
- https://github.com/craffel/mir_eval/actions/runs/5236939084
- https://github.com/craffel/mir_eval/actions/runs/5249531261

**Questions:**
1. Should the build matrix include Linux, macOS and Windows?
1. Should multiple conda environments and dependency resolution strategies be tested (as in [librosa/.github](https://github.com/librosa/librosa/tree/main/.github))?

Or high-level rather: how closely should mir_eval follow librosa? Identically? Loosely? This package doesn't use Numba or Cython so might be less susceptible to breaking. I'd consider using `actions/setup-python` because of that, but probably missing some important details about maintenance over time.

craffel:
Yeah, I'm comfortable with only testing on Linux. OTOH I don't even think we have any filesystem access stuff in the core library (just in the tests).

bmcfee:
> However, I also see your point that there's not a lot going on in mir_eval that could break on different platforms (eg cython builds or numba/llvmlite dependencies).

Actually, I'm going to reverse myself on this.  We should definitely include cross-platform tests.

While mir_eval itself doesn't have any platform-specific funny business, there have been incidents in the past where platform changes surfaced some nasty reproducibility issues that took me quite a bit of time to pin down.  These mainly had to do with differences in the underlying BLAS package between linux and mac (and maybe architecture dependent), causing slight deviations in numerical stability for bss eval.

We eventually resolved this by relaxing the tests to the point where these deviations didn't matter - remember kids, never trust decibel measurements past the first decimal place - but having cross-platform tests in place are handy for detecting this kind of thing quickly.

carlthome:
> Actually, I'm going to reverse myself on this. We should definitely include cross-platform tests.

Well, that might have been a good call actually because two unit tests fails with OOM on Windows despite running similar code with similar memory availability ([hardware specs according to GitHub.com](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)):

https://github.com/craffel/mir_eval/actions/runs/5261403240/jobs/9509466272?pr=356#step:6:592

Doesn't feel obvious to me that this isn't an actual error but could consider tweaking the sizes in [here](https://github.com/craffel/mir_eval/actions/runs/5261403240/jobs/9509466272?pr=356#step:6:41) and [here](https://github.com/craffel/mir_eval/actions/runs/5261403240/jobs/9509466272?pr=356#step:6:215) to something even shorter but [2147042648](https://github.com/craffel/mir_eval/actions/runs/5261403240/jobs/9509466272?pr=356#step:6:289) is curiously large so need more digging. Aiming to look more into this tomorrow!

carlthome:
But why would this only fail on Windows runners? Didn't think conda packages for NumPy/SciPy could behave that differently. Assuming some OpenBLAS vs MKL or gcc vs MSVC differences that a curious soul could dig into. 🤔

Guess this PR is blocked by figuring this out proper first.

bmcfee:
> But why would this only fail on Windows runners?

Is it possible the windows build is running 32bit while the others are 64bit, and the integer overflow we're hitting just happens to land in between?

carlthome:
> Is it possible the windows build is running 32bit while the others are 64bit, and the integer overflow we're hitting just happens to land in between?

Hmm, perhaps but supposedly `ubuntu-latest` and `windows-latest` both run on x86_64 as [per the guide](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) so would that be if miniconda is installed differently with `conda-incubator/setup-miniconda` depending on OS?

carlthome:
Well, it states "win32" here actually so perhaps! https://github.com/craffel/mir_eval/actions/runs/5261403240/jobs/9509466272?pr=356#step:6:9

carlthome:
Fishing for ideas in https://github.com/conda-incubator/setup-miniconda/issues/287

bmcfee:
This might be easier to debug if you add a `conda info` step to the job, as that will explicitly report the archspec of the environment.

It might also help to add a step that runs:
```
python -c "import numpy; numpy.show_config()"
```
just to get the full blas / lapack setup in the test log.

carlthome:
Hmm, didn't spot anything obviously wrong in the debug logging.

How would it feel to mark the Windows runner in the build matrix as excluded for now, with a TODO, and still proceed with merging this PR? Not optimal, but also nice with some incremental progress.

bmcfee:
We could also just leave it in (failing) and merge a fix (#355) later.

carlthome:
Sounds good! I'm happy enough with this PR for now so if you're up for it feel free to merge.

PR #359, opened 2023-07-02
Drop Python 2

carlthome:
IIRC `future` and `six` were for the Python 2 -> 3 migration. As Python 2 has passed EOL it's probably not necessary to mention it anymore.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -15,8 +15,6 @@ If you're looking for the mir_eval web service, which you can use to run mir_eva
 Dependencies:
 
 * `Scipy/Numpy <http://www.scipy.org/>`_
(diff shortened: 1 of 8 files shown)


craffel:
If we are going to do this, we need to remove imports/uses of six and future from the library and should probably add `python_requires` to the setup.py

carlthome:
> If we are going to do this, we need to remove imports/uses of six and future from the library and should probably add `python_requires` to the setup.py

Ah, good catch! I got fooled by `six` and `future` being transitive dependencies and thus the tests passed.

craffel:
Thanks!

PR #361, opened 2023-08-05
Add release workflow for GitHub Actions based PyPI package releases

carlthome:
### What?
Introduce a package release workflow for building and distributing `mir_eval` to PyPI.

### Install workflow on PyPI
Login to https://pypi.org/manage/account/publishing/ and fill in GitHub.com repo details under the `mir_eval` package settings.

### Usage of workflow on GitHub
1. Update package version in Python code.
2. Create a GitHub Releases release as repo owner here: https://github.com/craffel/mir_eval/releases
3. Watch the GitHub Actions workflow run. Once ✅ the newly built package should be on PyPI here https://pypi.org/project/mir_eval

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,55 @@
+name: Build and publish Python packages to PyPI
+
+on:
+  workflow_dispatch:
+  release:
+    types:
+      - published
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version:
+          - "3.8"
+          - "3.9"
+          - "3.10"
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: pip
+
+      - name: Install build tool
+        run: pip install build
+
+      - name: Build package
+        run: python -m build
+
+      - name: Upload package as build artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: package
+          path: dist/
+
+  publish:
+    runs-on: ubuntu-latest
+    needs: build
+    environment:
+      name: pypi
+      url: https://pypi.org/p/mir_eval
+    permissions:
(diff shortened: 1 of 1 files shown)


carlthome:
It would be nice to get out a patched version that works with latest NumPy so we can `pip install mir_eval` instead of `pip install git+https://github.com/craffel/mir_eval@d68afb0b37bfd0ff48a92fbe1ae1325a182cd471`. Perhaps a v0.7.1 or something?

bmcfee:
I think getting this to work will require @craffel to set up an auth token.  See https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ (We did this a few years back for librosa and it works like a charm.)





> It would be nice to get out a patched version that works with latest NumPy

Agreed - this should be a separate PR though.

craffel:
Thanks @carlthome and @bmcfee . I added an PyPI API key under the key PYPI_API_TOKEN - let me know if that works.

carlthome:
It's actually preferred to have shortlived tokens now instead, for PyPI with GitHub Actions (GHA) (as per [this](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect)) so the key cannot leak through PR logs and be abused.

The installation of this GHA workflow would be that you (@craffel) go to https://pypi.org/manage/project/mir_eval/settings/publishing/ and submits that HTML form as per this screenshot:
image

Then we can merge this (at which nothing happens), and once comfortable with releasing a new PyPI package, create a GitHub Releases release which will start the workflow.

carlthome:
>Agreed - this should be a separate PR though.

👍

craffel:
Cool, added the workflow.
image

carlthome:
Great! Then this PR can also be merged so the workflow itself is on `main`.

After that, it would be a matter of bumping https://github.com/craffel/mir_eval/blob/main/setup.py#L8 and filling in https://github.com/carlthome/audio-scraper/releases/new with the same version number.

craffel:
Thanks!

Issue #341, opened 2021-07-29
Bypassing validation


bmcfee:
**Warning, possibly bad ideas to follow.**

mir_eval is generally pretty defensive about its inputs, and (i think?) all of the evaluators include at least some minimal validation of inputs for proper formatting, semantics, data types, etc.  This is a good thing in general, but it can be quite wasteful when validation checks are expensive and performed many times on the same inputs.  In a current project of mine, I'm finding that more than half of the computation time is going into validation, and most of this is wasted as the comparisons are of different model outputs against the same reference data.

This makes me think that we could provide a mechanism to bypass validation of input in situations where the user is certain that the data is well-formed.  (Note: for now, I've simply commented out the validation lines in question, but that's obviously a bad solution in general.)  We could control this by a context manager and a global state variable (yuck, I know), which would be pretty slick from a user perspective:

```python
scores = mir_eval.chord.evaluate(...)  # Does validation checks as usual

with mir_eval.unsafe():
   scores = mir_eval.chord.evaluate(...)  # Skips validation checks
```

Ugly as a global state variable is, I think I prefer it to threading flag parameters through all evaluators here.  This is primarily because we already bury quite a few things in kwargs, and it would be easy to lose track.  Forcing a context manager for this would at least make it explicit and visually distinct from the user's perspective when this is happening.

Or it might just be a bad idea.  :shrug: Thoughts?

craffel:
I mean, there's a case to be made that we shouldn't be validating at the start of each metric function. This means, for example, that when we call a submodule's `evaluate` function we are validating once for each metric called, which is kind of silly. We could just leave it up to the end-user to call validate before calling a metric function, or we could do something weird like call `validate` when an exception is raised in the metric function (to get a more helpful error message). In general though I think a better solution is to think through why a given validation function is so expensive, and whether it needs to be (i.e. if we can just be lazy and instead throw a useful error later when the issue happens inside of the metric function instead of validating ahead of time).

Alternatively, we could consider adding caching to the validators that process either the estimates or the references alone - that way if a validator gets called with the same reference multiple times, it will just used the cached result. Of course, this is only a good idea if the references are small in memory.

bmcfee:
> or we could do something weird like call `validate` when an exception is raised in the metric function (to get a more helpful error message)

I don't think that would work -- most of the validation checks exist to prevent errors that would propagate silently.

[...]

PR #377, opened 2024-03-26
validation warnings set stacklevel=2

### LINKED ISSUE #372: Set warning stacklevels for validation
Validator functions often raise warnings when something isn't quite right, eg:

https://github.com/craffel/mir_eval/blob/57f7c31b120f6135c31207295372e3b67848126d/mir_eval/chord.py#L640-L644

However, the warning shows up to the user as coming from the `validate` function, which is not all that helpful since they never call this directly.  We can make this more useful by setting  `stacklevel=2`.  

This way the warning will point to the evaluator that called validate, which is a little more helpful for debugging purposes, especially when the user is calling a module-level `evaluate()` function.

bmcfee:
Fixes #372 

All validation functions now set stacklevel=2.

Before, line number comes from `validate`:

```python
n [4]: mir_eval.beat.evaluate(ref_data, np.array([]))
/home/bmcfee/git/mir_eval/mir_eval/beat.py:93: UserWarning: Estimated beats are empty.
  warnings.warn("Estimated beats are empty.")
```

After, line number comes from the metrics:

```python
In [9]: mir_eval.beat.evaluate(ref_data, np.array([]))
/home/bmcfee/git/mir_eval/mir_eval/beat.py:165: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
/home/bmcfee/git/mir_eval/mir_eval/beat.py:207: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
/home/bmcfee/git/mir_eval/mir_eval/beat.py:267: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
/home/bmcfee/git/mir_eval/mir_eval/beat.py:359: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
/home/bmcfee/git/mir_eval/mir_eval/beat.py:454: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
/home/bmcfee/git/mir_eval/mir_eval/beat.py:614: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
```

As it currently stands, we do a lot of redundant validation when using the `evaluate()` functions.  It might be worth considering an enhancement where `evaluate()` validates the data once at the top, and then each underlying metric provides an option to bypass validation thereafter.

I've proposed similar things in the past #341 as an optimization hack, but in this case, it's now also a usability issue because the relevant stacklevel to warn is different when metrics are called directly (2) vs called from `evaluate` (3).

I think it could make sense to roll that into this PR, or not.  Curious for other folks' input.

### DIFF
--- a/mir_eval/alignment.py
+++ b/mir_eval/alignment.py
@@ -141,7 +141,9 @@ def absolute_error(reference_timestamps, estimated_timestamps):
     return np.median(deviations), np.mean(deviations)
 
 
-def percentage_correct(reference_timestamps, estimated_timestamps, window=0.3):
+def percentage_correct(
+    reference_timestamps, estimated_timestamps, window=0.3, safe=True
+):
     """Compute the percentage of correctly predicted timestamps. A timestamp is predicted
     correctly if its position doesn't deviate more than the window parameter from the ground
     truth timestamp.
@@ -161,19 +163,27 @@ def percentage_correct(reference_timestamps, estimated_timestamps, window=0.3):
     window : float
         Window size, in seconds
         (Default value = .3)
+    safe : bool
+        If True, validate inputs.
+        If False, skip validation of inputs.
 
     Returns
     -------
     pc : float
         Percentage of correct timestamps
     """
-    validate(reference_timestamps, estimated_timestamps)
+    if safe:
+        validate(reference_timestamps, estimated_timestamps)
+
     deviations = np.abs(reference_timestamps - estimated_timestamps)
     return np.mean(deviations <= window)
 
 
 def percentage_correct_segments(
-    reference_timestamps, estimated_timestamps, duration: Optional[float] = None
+    reference_timestamps,
+    estimated_timestamps,
+    duration: Optional[float] = None,
+    safe=True,
 ):
     """Calculate the percentage of correct segments (PCS) metric.
 
@@ -215,13 +225,17 @@ def percentage_correct_segments(
     duration : float
         Optional. Total duration of audio (seconds). WARNING: Metric is computed differently
         depending on whether this is provided or not - see documentation above!
+    safe : bool
+        If True, validate inputs.
+        If False, skip validation of inputs.
 
     Returns
     -------
     pcs : float
         Percentage of time where ground truth and predicted segments overlap
     """
-    validate(reference_timestamps, estimated_timestamps)
+    if safe:
+        validate(reference_timestamps, estimated_timestamps)
     if duration is not None:
         duration = float(duration)
         if duration <= 0:
@@ -266,7 +280,7 @@ def percentage_correct_segments(
     return overlap_duration / duration
 
 
-def karaoke_perceptual_metric(reference_timestamps, estimated_timestamps):
+def karaoke_perceptual_metric(reference_timestamps, estimated_timestamps, safe=True):
     """Metric based on human synchronicity perception as measured in the paper
     "User-centered evaluation of lyrics to audio alignment" [#lizemasclef2021]
 
@@ -288,13 +302,17 @@ def karaoke_perceptual_metric(reference_timestamps, estimated_timestamps):
         reference timestamps, in seconds
     estimated_timestamps : np.ndarray
         estimated timestamps, in seconds
+    safe : bool
+        If True, validate inputs.
+        If False, skip validation of inputs.
 
     Returns
     -------
     perceptual_score : float
         Perceptual score, averaged over all timestamps
     """
-    validate(reference_timestamps, estimated_timestamps)
+    if safe:
+        validate(reference_timestamps, estimated_timestamps)
     offsets = estimated_timestamps - reference_timestamps
 
     # Score offsets using a certain skewed normal distribution
@@ -309,7 +327,7 @@ def karaoke_perceptual_metric(reference_timestamps, estimated_timestamps):
     return np.mean(perceptual_scores)
 
 
-def evaluate(reference_timestamps, estimated_timestamps, **kwargs):
+def evaluate(reference_timestamps, estimated_timestamps, safe=True, **kwargs):
     """Compute all metrics for the given reference and estimated annotations.
 
     Examples
@@ -328,6 +346,9 @@ def evaluate(reference_timestamps, estimated_timestamps, **kwargs):
     **kwargs
         Additional keyword arguments which will be passed to the
         appropriate metric or preprocessing functions.
+    safe : bool
+        If True, validate inputs.
+        If False, skip validation of inputs.
 
     Returns
     -------
@@ -335,6 +356,12 @@ def evaluate(reference_timestamps, estimated_timestamps, **kwargs):
         Dictionary of scores, where the key is the metric name (str) and
         the value is the (float) score achieved.
     """
+    if safe:
+        validate(reference_timestamps, estimated_timestamps)
+
+    # We can now bypass validation
+    kwargs["safe"] = False
+
     # Compute all metrics
     scores = collections.OrderedDict()
 
--- a/mir_eval/beat.py
+++ b/mir_eval/beat.py
@@ -133,7 +133,7 @@ def _get_reference_beat_variations(reference_beats):
     )
 
 
-def f_measure(reference_beats, estimated_beats, f_measure_threshold=0.07):
+def f_measure(reference_beats, estimated_beats, f_measure_threshold=0.07, safe=True):
     """Compute the F-measure of correct vs incorrectly predicted beats.
     "Correctness" is determined over a small window.
 
@@ -144,7 +144,7 @@ def f_measure(reference_beats, estimated_beats, f_measure_threshold=0.07):
     >>> estimated_beats = mir_eval.io.load_events('estimated.txt')
     >>> estimated_beats = mir_eval.beat.trim_beats(estimated_beats)
     >>> f_measure = mir_eval.beat.f_measure(reference_beats,
-                                            estimated_beats)
+    ...                                     estimated_beats)
 
     Parameters
     ----------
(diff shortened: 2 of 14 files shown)


bmcfee:
... of course, @craffel already suggested this 3 years ago :laughing: 

https://github.com/craffel/mir_eval/issues/341#issuecomment-889414180

craffel:
Thanks. What does this look like when you call a metric function directly? Regarding redundant calls to validate, this probably belongs in the other issue, but there must be some Python wizardry that basically says: If the same function is called with the same arguments within this context, skip the call. Basically some kind of argument caching/memoization but instead of caching the output, just skip the call completely, and evict the cache when you exit the context. That would work, right?

bmcfee:
Took the liberty of adding the `safe=` parameter to validators and evaluator functions.  This is WIP, I'm up through (alphabetically) the segment module.  TBD are separation, tempo, transcription, and transcription_velocity.

In all cases, the default behavior is unchanged.  Everything is validated by default.  However, we now avoid doing multiple validation calls on the same data.  Doing it this way means that `stacklevel=2` always kicks back to the call site of validation, regardless of whether the user called `evaluate()` or a more specific metric.

I also took the liberty of bumping up the chord parser regexp into a module-level compiled pattern.  Recompiling that monster regexp for every chord label was probably the source of the inefficiency I mentioned back in #341.

I'll report back when all safety parameterizations are done.

bmcfee:
> Thanks. What does this look like when you call a metric function directly? Regarding redundant calls to validate, this probably belongs in the other issue, but there must be some Python wizardry that basically says: If the same function is called with the same arguments within this context, skip the call. Basically some kind of argument caching/memoization but instead of caching the output, just skip the call completely, and evict the cache when you exit the context. That would work, right?

Hah, sorry, crossed wires here. :grin: 

The idea is that if the user calls `evaluate(..., safe=True)`, the validation happens once there, and is then subsequently skipped (`safe=False`).  This way we never have redundant calls, and the stacklevel points to the right place.

If the user calls a metric directly with `safe=True`, it still validates.

If the user calls either with `safe=False`, then validation is skipped entirely.  (Dangerous, but YOLO, and it makes sense in certain circumstances.)

Regarding memoization, yeah we could do that, but I don't think it actually buys us much here.  It would add at least one dependency (eg joblib) and probably complicate the stacklevel logic more than the solution above.

craffel:
This works, but doesn't it seem drastic to put a new arg into basically all of the functions just to avoid redundant calls in evaluate? It feels possibly worth it to do the Python wizardry to avoid adding an arg everywhere.

bmcfee:
> This works, but doesn't it seem drastic to put a new arg into basically all of the functions just to avoid redundant calls in evaluate? It feels possibly worth it to do the Python wizardry to avoid adding an arg everywhere.

I wouldn't consider it drastic if it's implemented consistently across the entire package (which is my plan).  It's perhaps a little clumsy, but it's easy to understand and use.

A slick pythonic alternative would be to bundle everything in a decorator that abstracts out the validation / bypass logic before getting onto the core of the evaluation code.  This might be a little fancier on the dev side, but the user experience would be pretty much the same.  In this case, as much as I like a slick decorator trick, I'd favor readability and transparency.

bmcfee:
... forgot to comment that I'd finished with this retrofit, and I think it's ready for CR.

Thinking more about potential python wizardry, I think the only reason to do that rather than the explicit parameterization I've implemented here is if we think we might change things later on.  This kind of thing happens more for things like deprecation warnings / parameter renames / etc., which can change frequently over the lifecycle of a project, so there's benefit to making it easy to add or remove at any point.  I don't think that applies in this case.

I do think we need validation bypassing as implemented here if we want all of the following:
- inputs are (can be) validated regardless of how a metric is called (directly or through evaluate)
- warning stacklevels point to the call site for the user and not within mir_eval
- validation can be disabled by the user

The only remaining question for me right now is whether the stacklevel should be 2 or 3.  The call path looks something like:

```
validate (1) ← [evaluate || specific metric] (2) ← user code (3)
```

main branch currently leaves it at 1, which is not helpful.  PR currently does 2, which would point to either evaulate or the specific metric being called; this seems marginally more helpful.  I think (3) might be best as it points to where the user is entering the mir_eval code path.  The only downside here is that if a user is calling validate() directly, then the stack level might point one level higher up than it should.  I expect this case is exceedingly rare and not something we should worry about, but I did want to note it here.

craffel:
Sorry for the delay in responding, as I mentioned my latency is a bit unpredictable these days. I appreciate the work you put into this but I don't think adding an additional argument to most user-facing functions in the library is the right solution - this hurts usability/readability significantly in addition to potential backwards compatibility issues with positional arguments and therefore should be a last resort. I also don't think it's important to support users disabling validation - the validation logic is there to catch problems that otherwise would cause issues and throw more helpful errors/warnings; the only reason it's in a separate function is to avoid code duplication in all of the metric functions. If there was no other option, I think it would make sense to consider adding an arg, but in this case the issue it solves is minor enough (repeated warnings for malformed inputs when evaluate is called) that I'm not even sure it's warranted. In any case I don't think we need to address this issue or come up with a solution for 0.8.0 so I would be most inclined to punt on it for now.

Regarding the stack level, I'm not sure since IME the stack level reported by warnings is almost never helpful... but I agree that validate should almost never be called by an end-user so (3) is probably best.

bmcfee:
> his hurts usability/readability significantly in addition to potential backwards compatibility issues with positional arguments and therefore should be a last resort.

I don't think this affects positional arguments in existing code, as it's always the last new (positional) parameter.

I also don't see how this hurts usability or readability: the default behavior is preserved, and the behavior is explicitly stated in documentation.  At most it's one extra piece of information in the docs?



> In any case I don't think we need to address this issue or come up with a solution for 0.8.0 so I would be most inclined to punt on it for now.

I'm okay with punting if you think it's really unimportant.



> Regarding the stack level, I'm not sure since IME the stack level reported by warnings is almost never helpful... but I agree that validate should almost never be called by an end-user so (3) is probably best.

I actually do use warning stacklevels somewhat often, so if we have a chance to do it right, I think we should.  That said, I don't think we *can* do it right without the change I've implemented here, since there are two alternative routes to the validate function with different stacklevels (user→metric→validate and user→evaluate→metric→validate).  This is why I went to the effort of lifting out the validate call and adding bypass options at all.  I agree with your earlier point that disabling validation entirely is of negligible utility, but I don't think there's another good option here that still produces a correct stacklevel.

(A grungier alternative would be to pass the target stacklevel through to validate, but that's much less generally readable or usable than a bypass option.)

craffel:
> I also don't see how this hurts usability or readability: the default behavior is preserved, and the behavior is explicitly stated in documentation. At most it's one extra piece of information in the docs?

Adding an argument to almost all user-facing functions that is almost never used harms readability and usability. I'm not saying it's catastrophic, I'm just saying it should be a last resort - if it can be avoided (which I believe it can) it should be avoided.

Regarding the stack level, thinking about it more, I think stack level 2 is reasonable - it's flagging for the user that the warning is coming from some particular metric function. If that's inside an evaluate call, that's ok, because ultimately all evaluate does is do some preprocessing and then call the metric functions. I don't think we can in general assume some point in the stack level corresponds to the point where the user has constructed and passed off the data being evaluated, because one level above the evaluate/metric function call could either be user code or some other library code that the user is using on top of mir_eval.

bmcfee:
Ok, I'll close this out then.

craffel:
Just to clarify, I wasn't saying to nuke this entirely - I still think stacklevel=2 is a QoL improvement over the current behavior and is a simple change that should be merged in.

bmcfee:
> Just to clarify, I wasn't saying to nuke this entirely - I still think stacklevel=2 is a QoL improvement over the current behavior and is a simple change that should be merged in.

Yeah I understand.  But as I've said above, `stacklevel=2` without accounting for the alternative code paths is not actually that helpful.  In fact, it can be worse than status quo because it triggers separate warnings every time the data is validated (when called by `evaluate()`):
```python
In [9]: mir_eval.beat.evaluate(ref_data, np.array([]))
/home/bmcfee/git/mir_eval/mir_eval/beat.py:165: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
/home/bmcfee/git/mir_eval/mir_eval/beat.py:207: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
/home/bmcfee/git/mir_eval/mir_eval/beat.py:267: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
/home/bmcfee/git/mir_eval/mir_eval/beat.py:359: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
/home/bmcfee/git/mir_eval/mir_eval/beat.py:454: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
/home/bmcfee/git/mir_eval/mir_eval/beat.py:614: UserWarning: Estimated beats are empty.
  validate(reference_beats, estimated_beats)
```
Since these are coming from different lines in the module (due to stacklevel) these appear to be different, and are thus not absorbed by the `"once"` default rule for UserWarnings.  When stacklevel is not set (status quo), repeat warnings appear to come from the same place, and are (correctly, imo) absorbed by the warnings filter:
```python
In [6]: mir_eval.beat.evaluate(ref_data, np.array([]))
/home/bmcfee/git/mir_eval/mir_eval/beat.py:93: UserWarning: Estimated beats are empty.
  warnings.warn("Estimated beats are empty.")
```

Now, Python 3.12 actually added some functionality to do this in exactly the right way with the [`skip_file_prefixes` parameter](https://docs.python.org/3/library/warnings.html#warnings.warn), but I don't think we can bump the minimum python to 3.12 just for this feature.

Which is all to say: I don't think setting stacklevel=2 universally is correct or more helpful than the current behavior, and I also don't see a cleaner way to achieve the most helpful behavior than what I'd implemented in this PR.

craffel:
Ah, I misunderstood the issue. I'll work on a solution to avoiding repeat calls to validate after my leave ends.

PR #379, opened 2024-04-01
Pitch contour sonification supports nans

### LINKED ISSUE #332: sonify.pitch_contour should support nans
Minor inconvenience, but for f0 estimation, it's nice to be able to use `nan` to encode missing/unvoiced regions (rather than 0, or egad, negative frequency).  Doing so breaks the sonifier, but it would be an easy enough fix to support it.

bmcfee:
Fixes #332

This PR works be replacing nan-values by 0 in the frequency array prior to interpolation, thereby replicating the existing logic for unvoicing certain frames.

This now lets us use nan to represent unvoiced frames, which is the default behavior for (eg) librosa's pyin implementation.

I don't think we need specific additional tests here, as the change is minimal and pretty obvious when you see it.

### DIFF
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -224,7 +224,7 @@ def pitch_contour(
         time indices for each frequency measurement, in seconds
     frequencies : np.ndarray
         frequency measurements, in Hz.
-        Non-positive measurements will be interpreted as un-voiced samples.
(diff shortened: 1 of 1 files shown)


bmcfee:
Any objection to merging this?

craffel:
Looks good, thanks!

PR #382, opened 2024-06-10
deprecating separation module

bmcfee:
This PR deprecates the separation module.

Question for @faroit - is sigsep-museval the best replacement to recommend here?

Question for @craffel - should we disable tests on the separation module?  They've always been finnicky and the vast majority of build time.  If we're deprecating the module anyway, I could see some benefit to disabling these tests.

### DIFF
--- a/mir_eval/separation.py
+++ b/mir_eval/separation.py
@@ -1,5 +1,11 @@
 # -*- coding: utf-8 -*-
 """
+.. warning::
+    The mir_eval.separation module is deprecated in mir_eval version 0.8, and will be removed.
+    We recommend that you migrate your code to use an alternative package such as sigsep-museval
+    https://sigsep.github.io/sigsep-mus-eval/
+    
+
 Source separation algorithms attempt to extract recordings of individual
 sources from a recording of a mixture of sources.  Evaluation methods for
 source separation compare the extracted sources from reference sources and
@@ -141,6 +147,7 @@ def _any_source_silent(sources):
     )
 
 
+@util.deprecated(version="0.8", version_removed="0.9")
 def bss_eval_sources(reference_sources, estimated_sources, compute_permutation=True):
     """
     Ordering and measurement of the separation quality for estimated source
@@ -251,6 +258,7 @@ def bss_eval_sources(reference_sources, estimated_sources, compute_permutation=T
         return (sdr, sir, sar, popt)
 
 
+@util.deprecated(version="0.8", version_removed="0.9")
 def bss_eval_sources_framewise(
     reference_sources,
     estimated_sources,
@@ -362,6 +370,7 @@ def bss_eval_sources_framewise(
     return sdr, sir, sar, perm
 
 
+@util.deprecated(version="0.8", version_removed="0.9")
 def bss_eval_images(reference_sources, estimated_sources, compute_permutation=True):
     """Compute the bss_eval_images function from the
     BSS_EVAL Matlab toolbox.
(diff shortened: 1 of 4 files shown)


faroit:
@bmcfee thanks for the initiative. Before we merge this, I would guess we should get some feedback from the community. museval might be a good fit as its an implementation of the bsseval `images` (unlike `_sources`) algorithm but it might needs some work to make it future proof as well: wrt to the precision issue, I also had to lower the regression tests tolerance https://github.com/sigsep/sigsep-mus-eval/pull/93/files#diff-6edc05e4134fe68983d08319a22f62d5e4efdd829656595282cf6ef97011c13aL90 myself (testing against older version of museval with older numpy version).

Here are some people that I would def value their opinion on how to proceed with the sdr metrics:

@rabitt @fakufaku @iver56 @adefossez

bmcfee:
> Thanks. Yes, let's disable the tests too.

Done.



> museval might be a good fit as its an implementation of the bsseval `images` (unlike `_sources`) algorithm but it might needs some work to make it future proof as well: wrt to the precision issue, I also had to lower the regression tests tolerance https://github.com/sigsep/sigsep-mus-eval/pull/93/files#diff-6edc05e4134fe68983d08319a22f62d5e4efdd829656595282cf6ef97011c13aL90 myself

I think the precision issue will always be there as long as we're relying on unregularized least squares; even moreso if we're dynamically switching between solve() and lstsq(), as mir_eval and museval both do.  This dynamic behavior, IIRC, was intended to mimic what matlab's \ operator does under the hood, and it's always seemed troublesome to me.  Museval's implementation at least puts a ridge on the solve(), so it really shouldn't ever hit the lstsq path, but as I noted in the other PR thread #381 , the exception handling there isn't quite correct anymore.

My not-so-humble opinion is that all of these implementations should probably be scrapped and rewritten :sweat_smile:  But if we're going to do that (as a community), it should only happen once, and in a dedicated package.

bmcfee:
Side note: failures on 3.12 builds are due to matplotlib removeing BrokenBarHCollection.  Will fix that in a separate PR.

craffel:
Is there a way to tell codecov not to worry about the separation file's lines so it hates us less?

fakufaku:
Jumping in the discussion here. I have written a fast `bss_eval_sources` package ([fast_bss_eval](https://github.com/fakufaku/fast_bss_eval)). At the time, the fast routine was quite faster than the sigsep/bsseval implementation. It also supports numpy/torch, fp32/fp64, cpu/gpu, and `solve` or conjugate GD.



I think the implementation ideas are applicable to the `_images` routine for which the savings are potentially even larger.

iver56:
I'm okay with this. I don't calculate SIR, SAR and ISR anymore when evaluating source separation models/algorithms. I still calculate SDR (with a stripped fork of [sigsep/bsseval](https://github.com/sigsep/bsseval)), but I mostly don't use it. I mainly rely on [logWMSE](https://github.com/nomonosound/log-wmse-audio-quality) in combination with a few others.

bmcfee:
> Is there a way to tell codecov not to worry about the separation file's lines so it hates us less?

I've added separation to the omit section, but it might take a merge before codecov stops comparing to old results that did include it.

bmcfee:
> Side note: failures on 3.12 builds are due to matplotlib removing BrokenBarHCollection. Will fix that in a separate PR.

Actually just rolled it into this one - all that needed to happen was to remove an import tweak a docstring, no actual code changes.

bmcfee:
Ok, I think this is good to merge now. All matplotlib issues are resolved and tests are workign out of the box. @craffel ?

bmcfee:
Yerp, realized this was already approved.  :shipit:

PR #387, opened 2024-08-18
Minor Python 2 to Python 3 changes

carlthome:
Very minor nitpicks but perhaps nice to keep the syntax fresh since Python 3.8 is now the latest non-EOL version.

1. Drop utf-8 encoding header
1. Use f-strings
1. Use set comprehension, set literals, dict comprehensions, instead of converting lists

Nothing major and not urgent to fix. And TBH arguably not worth merging in case this breaks unexpected monkey patches in the wild. This is just me sitting at an airport with a coffee and reading the source. 😄


carlthome:
Ah, so `black` is enforced on the source code now too? That's nice!

bmcfee (approved):
All looks good to me, thanks!

bmcfee (approved):

PR #390, opened 2024-10-21
updated urls to point to new org

bmcfee:
Does what it says on the tin - URLs are updated to new address.


craffel:
Thanks, was going to do this myself! Not sure what is up with the CI failures, it's giving some kind of permission/token error?

bmcfee:
> Not sure what is up with the CI failures, it's giving some kind of permission/token error?

Yeah, I think we need to reconfigure the coverage integration.  I'll try and get that sorted today.

PR #395, opened 2024-11-26
refactored and optimized segmentation metrics eval

Mintas:
Refactored and optimized segmentation metrics evaluation.

1. Refactoring:
key purpose is to make some of the methods more useful for extension in client libraries, make code more concise, reducing code duplications, thus also increasing possible community support

2. Optimisation:
here we note, that some code duplications lead us to recalculate some preparative steps for metric calculation. Reusing common logical steps helped to significantly increase performance. In our microbenchmark of ```segment.eval()``` we have achieved 1.58 speedup! 
This may be crucial in some applications that may require intensive metric calculation (like Hyperparameter Optimization)

3. Benchmarking details:
We took some symbolic music files (midis) with known expert markup of musical segments. There were 20 data points considered. We ran 300 iterations of eval method over all data provided to get better averaging. 

Benchmarking results:
Before: time: 24.702 and latency: 0.0823, avg per data point: 0.0041 
After: time: 15.615 and latency: 0.0520, avg per data point: 0.0026
**Speedup is: 1.582**

### DIFF
--- a/mir_eval/segment.py
+++ b/mir_eval/segment.py
@@ -73,6 +73,7 @@ References
 """
 
 import collections
+import functools
 import warnings
 
 import numpy as np
@@ -80,9 +81,9 @@ import scipy.stats
 import scipy.sparse
 import scipy.misc
 import scipy.special
+from typing import Sequence, Any
 
 from . import util
-
 
 def validate_boundary(reference_intervals, estimated_intervals, trim):
     """Check that the input annotations to a segment boundary estimation
@@ -118,6 +119,17 @@ def validate_boundary(reference_intervals, estimated_intervals, trim):
     for intervals in [reference_intervals, estimated_intervals]:
         util.validate_intervals(intervals)
 
+def validated_trimmed_boundaries(reference_intervals, estimated_intervals, trim):
+    validate_boundary(reference_intervals, estimated_intervals, trim)
+    # Convert intervals to boundaries
+    reference_intervals = util.intervals_to_boundaries(reference_intervals)
+    estimated_intervals = util.intervals_to_boundaries(estimated_intervals)
+
+    # Suppress the first and last intervals
+    if trim:
+        reference_intervals = reference_intervals[1:-1]
+        estimated_intervals = estimated_intervals[1:-1]
+    return reference_intervals, estimated_intervals
 
 def validate_structure(
     reference_intervals, reference_labels, estimated_intervals, estimated_labels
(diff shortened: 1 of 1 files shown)


Mintas:
@bmcfee  @craffel 
This PR is getting stale, would you consider reviewing and merging it?
TY

bmcfee:
Sorry I won't have time to review this for a few weeks.

bmcfee:
Sorry for the delay, I'm looking at this now.  It's not the easiest PR to review because there's a mixture of cosmetic changes (ie variable names) and refactoring to sift through.  It also appears that tests are not passing.

mir_eval/segment.py
```diff
@@ -165,9 +177,20 @@ def validate_structure(
...
+    def decorator_default_on_empty(func):
+        @functools.wraps(func)
+        def wrapper_default_on_empty(reference_indices, estimated_indices, *args, **kwargs):
+            # Check for empty annotations.  Don't need to check labels because
+            # validate_structure makes sure they're the same size as intervals
+            if len(reference_indices) == 0 or len(estimated_indices) == 0:
+                return default_return_value
+            return func(reference_indices, estimated_indices, *args, **kwargs)
+        return wrapper_default_on_empty
+    return decorator_default_on_empty
```
bmcfee, line 190: This would probably be cleaner using the `decorator` module, as we do in util, eg https://github.com/mir-evaluation/mir_eval/blob/935451d7c06b6eea63608d4b8a869f333440b1df/mir_eval/util.py#L946-L962

mir_eval/segment.py
```diff
@@ -458,58 +492,40 @@ def rand_index(
         length (in seconds) of frames for clustering
         (Default value = 0.1)
     beta : float > 0
-        beta value for F-measure
-        (Default value = 1.0)
+        deprecated parameter - to be removed in 0.9 !!!
```
bmcfee, line 495: Why is this deprecated?

mir_eval/segment.py
```diff
@@ -1178,6 +1137,35 @@ def vmeasure(
...
+
+def _return_values_to_mapping(mapping:Sequence[Any], values, accumulator:dict=None) -> dict[Any, Any]:
+    if accumulator is None:
+        accumulator = {}
+    if not isinstance(values, tuple):
+        accumulator[mapping[0]] = values
+    else:
+        for i,m in enumerate(mapping):
+            accumulator[m] = values[i]
+    return accumulator
```
bmcfee, line 1163: I don't understand what these functions do - it seems like abstraction that may not be necessary, and it definitely does not help code readability.

mir_eval/segment.py
```diff
@@ -1227,71 +1215,39 @@ def evaluate(ref_intervals, ref_labels, est_intervals, est_labels, **kwargs):
     # Now compute all the metrics
     scores = collections.OrderedDict()
 
-    # Boundary detection
+    # Metrics for Intervals without structure labels section:
+    trim = kwargs.get('trim')
```
bmcfee, line 1219: This is a little unclear, since it's by default going to give `trim is None` rather than `trim is False`.  The behavior will be the same (at least until we add type annotations), but it would be better to explicitly populate a boolean.

mir_eval/segment.py
Mintas: Good note, gotta update my solution!

mir_eval/segment.py
bmcfee: Hrm, indeed.  I wonder how this ended up there in the first place?

I agree that it makes sense to deprecate this.  However, I think we should take that to a separate PR, and add in a bit of machinery to support deprecations better.  (See, e.g., https://github.com/librosa/librosa/blob/ebd878fabb87c588f98f372a94fca5d167c59cf2/librosa/util/deprecation.py#L10-L15 )

Issue #11, opened 2013-12-04
validate input data


bmcfee:
Data import should throw an exception if the input is not well-formed.

Example:

```
0.00 5.12 INTRO
5.12 4.12 VERSE
...
```

should fail loudly upon import.

PR #397, opened 2024-12-07
Only "proper" chord types in ```mir_eval.chord.QUALITIES```

maximoskp:
Following the discussion in #396 and thanks to @bmcfee insightful comments, this PR simply removes some chord qualities in ```mir_eval.chord.QUALITIES``` that are "unreachable" by the regexp in the same file.

I.e., even though there the aforementioned ```QUALITITES``` dictionary includes keys for the ```b9```, ```#9```, ```#11``` and ```b13``` chord qualities, these are "unreachable" from the regexp. This is for a good reason, since (agreeing with @bmcfee) writing ```C:#9``` is a bit odd - denoting it as ```C:7(#9)``` seems to be more proper.

Even though this PR doesn't fix any error per se, it resolves an edge-case error that occurs if, for instance, someone wants to iterate through ```mir_eval.chord.QUALITIES``` to get all possible chord qualities in ```mir_eval```. Indeed, trying to ```mir_eval.chord.encode(C:#9)``` produces and error, but this error is expected to appear if, for some reason, someone tries to iterate through all qualities.

This PR simply removes the four qualities that result in this errors in this edge case. Disclaimer: I haven't tested whether this change breaks some other functionality, but I believe it shouldn't, since any other such functionality would ultimately have to be related with the regexp that properly checks for nicely-formed qualities.

Thank you @bmcfee for the help and the clarifications!


bmcfee:
Thanks @maximoskp , and apologies for the slow response on this!

The core change looks good to me.  I've added a couple of comments above on what appear to be unnecessary changes in the tests.  Once that's sorted, I'm happy to merge.

tests/test_chord.py
```diff
@@ -638,3 +638,9 @@ def test_validate():
     # Test that error is thrown on different-length labels
     with pytest.raises(ValueError):
         mir_eval.chord.validate([], ["C"])
+
+# @maximoskp
+# make testing for all chord types
+# quals = mir_eval.chord.QUALITIES
+# for q in quals.keys():
+#     c = mir_eval.chord.encode('C' + (len(q) > 0)*':' + q, reduce_extended_chords=True)
```
bmcfee, line 646:
```suggestion
```

bmcfee (changes requested):

tests/test_chord.py
```diff
@@ -639,8 +639,10 @@ def test_validate():
...
-# quals = mir_eval.chord.QUALITIES
-# for q in quals.keys():
-#     c = mir_eval.chord.encode('C' + (len(q) > 0)*':' + q, reduce_extended_chords=True)
\ No newline at end of file
+def test_chord_qualities():
+    quals = mir_eval.chord.QUALITIES
+    for q in quals.keys():
+        try:
+            c = mir_eval.chord.encode('C' + (len(q) > 0)*':' + q, reduce_extended_chords=True)
+        except:
```
bmcfee, line 647: This isn't the correct way to implement tests in pytest.

In general, a bare except (ie not catching a specific, expected exception type) is bad practice, since it lets any kind of error pass through.  Additionally, this test function will succeed no matter what; the log will contain your print outputs, but pytest will consider it a success.

A better way to do this would be to 1) parametrize the test by `quals.keys()` (so each encode runs independently of the others), and 2) not try to catch exceptions.  That way, anything that does not encode properly will raise an exception and we can see exactly what the failure was.  Something like:

```python
@pytest.mark.parametrize('quality', list(mir_eval.chord.QUALITIES.keys()))
def test_chord_qualities(quality):
    if len(quality) > 0:
        mir_eval.chord.encode(f"C:{quality}"), reduce_extended_chords=True)
    else:
        mir_eval.chord.encode("C", reduce_extended_chords=True)
```

maximoskp: Thank you again for the help, @bmcfee , I really appreciate you taking the time to explain things and provide solutions.

I just copy-pasted your example for the test, I haven't used pytest before - I'll need to start using it as it seems...

I don't know how it works, but if it is possible or "not improper" in some way, anyone is welcome to "take the pull request over" and complete it in a proper way (I am not concerned in getting any credit for the pull request etc, I'm just interested in the result :D ).

bmcfee: Great, thanks!

The tests are passing, but codecov is still giving us trouble.  I'll fix that in a separate PR.

Otherwise, the only lingering issue is that the linting action is failing:

https://github.com/mir-evaluation/mir_eval/actions/runs/12933203903/job/36071425462?pr=397

I checked this and it's an easy enough fix that I'll just patch it in here.  (Problem is using single quotes instead of double.)

tests/test_chord.py
```diff
@@ -638,3 +638,10 @@ def test_validate():
     # Test that error is thrown on different-length labels
     with pytest.raises(ValueError):
         mir_eval.chord.validate([], ["C"])
+
+@pytest.mark.parametrize('quality', list(mir_eval.chord.QUALITIES.keys()))
+def test_chord_qualities(quality):
+    if len(quality) > 0:
+        mir_eval.chord.encode(f"C:{quality}", reduce_extended_chords=True)
+    else:
+        mir_eval.chord.encode("C", reduce_extended_chords=True)
```
bmcfee, line 647:
```suggestion
@pytest.mark.parametrize("quality", list(mir_eval.chord.QUALITIES.keys()))
def test_chord_qualities(quality):
    if len(quality) > 0:
        mir_eval.chord.encode(f"C:{quality}", reduce_extended_chords=True)
    else:
        mir_eval.chord.encode("C", reduce_extended_chords=True)
        
```

tests/test_chord.py
```diff
@@ -638,3 +638,11 @@ def test_validate():
...
     with pytest.raises(ValueError):
         mir_eval.chord.validate([], ["C"])
+
+@pytest.mark.parametrize("quality", list(mir_eval.chord.QUALITIES.keys()))
+def test_chord_qualities(quality):
+    if len(quality) > 0:
+        mir_eval.chord.encode(f"C:{quality}", reduce_extended_chords=True)
+    else:
+        mir_eval.chord.encode("C", reduce_extended_chords=True)
+
```
bmcfee, line 648:
```suggestion


@pytest.mark.parametrize("quality", list(mir_eval.chord.QUALITIES.keys()))
def test_chord_qualities(quality):
    if len(quality) > 0:
        mir_eval.chord.encode(f"C:{quality}", reduce_extended_chords=True)
    else:
        mir_eval.chord.encode("C", reduce_extended_chords=True)

```

bmcfee (approved):

PR #400, opened 2025-01-28
Updates to documentation, rtd configurations

bmcfee:
WIP for getting the RTD build to work well and look nice.

Also updates to things like broken links, etc.

### DIFF
--- a/.github/environment-docs.yml
+++ b/.github/environment-docs.yml
@@ -1,0 +1,10 @@
+name: docs
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - pip
+  - numpy >=1.15.4
+  - scipy >=1.4.0
+  - matplotlib-base>=3.3.0
+  - numpydoc
--- a/.github/workflows/docs-ci.yml
+++ b/.github/workflows/docs-ci.yml
@@ -62,6 +62,7 @@ jobs:
 
         - name: Build docs
           shell: bash -l {0}
+          working-directory: docs
           run: make html
 
         - name: Link checking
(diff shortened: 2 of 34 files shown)


bmcfee:
Annoyingly, RTD community (ie free) mode does not really support organization-owned repos on github.  See: https://me-readthedocs.readthedocs.io/en/latest/commercial/organizations.html

It still "works", see eg https://mir-eval.readthedocs.io/latest/ , but security credentials for webhooks don't work, which means that we can't get doc builds on pull requests.  This makes it a real pain to debug doc builds.  I suspect we may have to scrap this idea and fall back on self-hosting.

bmcfee:
Ok, since RTD won't do doc builds on PR for us, I'm adding a CI workflow to do it for us.  We won't get a preview of the build artifact, but it will at least let us know if the doc build fails.

bmcfee:
@stefan-balke @craffel I've implemented a restructuring of the doc site that I think should make things a bit easier to navigate.  Everything is still essentially accessible from the landing, but submodules each get their own page now so it's not one giant doc.

craffel:
Looks good, thanks for handling all of this. I know Sphinx conf.py files by convention have lots of lines commented out, but maybe we can avoid adding new commented-out lines?

bmcfee:
> but maybe we can avoid adding new commented-out lines?

I don't think I added anything new, just fixed some formatting on lines that were pre-existing and commented.  But in general I agree, we shouldn't add more than is needed.

bmcfee:
Test failures are due to codecov weirdness, still not finding our token.  But as everything else is in the green, I'd like to merge this to see how the RTD build looks with the new layout.

PR #410, opened 2025-02-13
fixing some alignment issues in time-frequency sonification

### LINKED ISSUE #408: `sonify.chords` does not work with a single interval
if we try to run `sonify.chords` with a single interval, it throws a TypeError.

upon inspection, the problem seems to be in the [`_const_interpolator`](https://github.com/mir-evaluation/mir_eval/blob/a0a8672d11ab66c3a89900d0dd35d8cdfcf8a1f9/mir_eval/sonify.py#L159-L167) (inside `time_frequency`) that returns a single float. my understanding is that it should instead return an array filled with the correct value. if that is the case, i can work on a fix. 

--- 

steps to reproduce:

```python
import mir_eval 
import numpy as np

intervals = np.array([[3,6]])
mir_eval.sonify.chords(["C"], intervals, 8000) 
```

> Traceback (most recent call last):
  File ".../mir_eval/tmp_sonify_test.py", line 7, in <module>
    mir_eval.sonify.chords(["C"], intervals, 8000)
  File ".../mir_eval/mir_eval/sonify.py", line 374, in chords
    return chroma(chromagram, intervals, fs, **kwargs)
  File ".../mir_eval/mir_eval/sonify.py", line 339, in chroma
    return time_frequency(gram, frequencies, times, fs, **kwargs)
  File ".../mir_eval/mir_eval/sonify.py", line 212, in time_frequency
    output[t_in:t_out] += wave[: len(signal)] * signal
TypeError: object of type 'numpy.float64' has no len()

---

@bmcfee ?

bmcfee:
Fixes #408 

It appears that fixing the interval padding issue resolved the error with single-interval sonification.

### DIFF
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -192,7 +192,7 @@ def time_frequency(
         signal_n = scipy.signal.convolve(signal[n], filter, mode="same")
 
         # Mix the signal into the output
-        output[:] += wave[:len(signal_n)] * signal_n
+        output[:] += wave[: len(signal_n)] * signal_n
 
     # Normalize, but only if there's non-zero values
     norm = np.abs(output).max()
--- a/mir_eval/sonify.py
+++ b/mir_eval/sonify.py
@@ -105,23 +105,43 @@ def time_frequency(
         synthesized version of the piano roll
 
     """
-    # Default value for length
+    # Convert times to intervals if necessary
+    time_converted = False
     if times.ndim == 1:
         # Convert to intervals
-        times = util.boundaries_to_intervals(times)
+        times = np.hstack((times[:-1, np.newaxis], times[1:, np.newaxis]))
+        # We'll need this to keep track of whether we should pad an interval on
+        time_converted = True
 
+    # Default value for length
     if length is None:
         length = int(times[-1, 1] * fs)
 
     last_time_in_secs = float(length) / fs
+
+    if time_converted and times.shape[0] != gram.shape[1]:
+        times = np.vstack((times, [times[-1, 1], last_time_in_secs]))
+
+    if times.shape[0] != gram.shape[1]:
+        raise ValueError(f"times.shape={times.shape} is incompatible with gram.shape={gram.shape}")
+
+    if frequencies.shape[0] != gram.shape[0]:
+        raise ValueError(f"frequencies.shape={frequencies.shape} is incompatible with gram.shape={gram.shape}")
+
     if times.min() > 0:
         # We need to pad a silence column on to gram at the beginning
         gram = np.pad(gram, ((0, 0), (1, 0)), mode="constant")
+        times = np.vstack(([0, times.min()], times))
 
     if times.max() < last_time_in_secs:
         # We need to pad a silence column onto gram at the end
         gram = np.pad(gram, ((0, 0), (0, 1)), mode="constant")
-    times, _ = util.adjust_intervals(times, t_max=last_time_in_secs)
+        times = np.vstack((times, [times.max(), last_time_in_secs]))
+
+    # Identify the time intervals that have some overlap with the duration
+    idx = np.logical_and(times[:, 1] >= 0, times[:, 0] <= last_time_in_secs)
+    gram = gram[:, idx]
+    times = np.clip(times[idx], 0, last_time_in_secs)
 
     # Truncate times so that the shape matches gram. However if the time boundaries were converted
     # to intervals, then the number of times will be reduced by one, so we only truncate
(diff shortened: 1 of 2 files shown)


craffel:
Looks good but get those coverage numbers up to 100%!! 📈

craffel (approved):

bmcfee:
Will add tests on this once the dependent stops oozing everywhere and I have the brain cycles to put into it.  For now it'd be good if @giovana-morais could kick the tires on this with the downstream jams test that revealed the issue originally.

giovana-morais:
hey there! most tests are fixed, but there seems to be another corner case. this function breaks whenever we provide a `duration=1.0` for the signal (which in mir_eval means `length=sr`))

```
import mir_eval
intervals = np.array([[3., 4.]]) 
# works
mir_eval.sonify.chords(["C"], intervals, fs=8000, length=None)  
# breaks
mir_eval.sonify.chords(["C"], intervals, fs=8000, length=8000)
# also breaks
mir_eval.sonify.chords(["C"], intervals, fs=8000, length=8000*2)
# works???
mir_eval.sonify.chords(["C"], intervals, fs=8000, length=8000*3)  
```

bmcfee:
That's interesting.  I'd think either all of the last 3 should break or they should all work.  But I guess we don't have much in the way of intelligent bounds processing here.  Maybe this function really just does need a whole rewrite.

bmcfee:
Ok, I see what's going on here, and the length parameters are actually just surfacing different variations of the same bug as before. Specifically, this line:

```
    times, _ = util.adjust_intervals(times, t_max=last_time_in_secs)
```
simultaneously does the expansion and truncation of the time intervals array to fully cover `[0, length]`.  My fix for the initially reported bug hits the case where we needed to also expand the `gram` array to have the same number of time steps as `times` (we hadn't been doing this previously!).  Where it's failing now is when we have to fully clip out one or more time intervals.

The `length=8000` and `length=8000*2` cases fail because they are both less than the start of an interval ([8000*3, 8000*4]), so adjust_intervals removes an interval from the array and we have a shape mismatch between gram and times.  The `length=8000*3` case doesn't fail because it just barely touches that last interval, so even though the interval's duration is clipped down to 0, the times array still has the same shape as gram.

bmcfee:
Ok, I think this PR is good to go, at least for the sake of getting the sonifier to behave sanely.

That said, I think there are some pretty serious issues with the interpolator here:
https://github.com/mir-evaluation/mir_eval/blob/a0a8672d11ab66c3a89900d0dd35d8cdfcf8a1f9/mir_eval/sonify.py#L190-L196

we're doing linear interpolation between interval midpoints to sample the amplitudes applied to the synthesized signal.  This is fine when intervals are short and equally spaced, but for long and irregular intervals (eg chord annotations), this leads to long and irregular ramps that put the sonified signal well outside the bounds of the original annotation.  For example, the motivating use case in 408 had a single annotation spanning [3, 4].  When we prepend a silence from [0, 3], we now end up with a linear ramp up from 1.5 to 3.5, and down from 3.5 to 4.  When the length= parameter is used, this can result in divergent outputs for different length values because the fade-out will have varying length.

I think we should replace this with a nearest-neighbor interpolation, which would A) be consistent across different length= settings, and B) more clearly conform to the annotation.

We can take this in a separate issue, but as it's only a 1-liner fix here, I could easily merge it into this PR as well.

bmcfee:
Since my head is in this code and i have a few minutes now, I went ahead and implemented the nearest neighbor interpolation mode.  Since we're now sonifying over the entire time range and then using interpolated gram values to mask out, this leads to transients at interval boundaries.  I've put in a hack around this that convolves each nn-interpolated row of gram by an averaging filter with order set to match two cycles of the frequency in question.  This gives us a smooth ramp up and down at the boundaries, and constant values within the interval.

Results sound good and seem just as fast as the older implementation.  That said, I'm now seeing a bunch of opportunities to vectorize and speed this up further, eg by doing all frequency interpolations at once instead of over a loop.  That can definitely be a separate PR.

bmcfee:
One last couple of updates here:

- nearest neighbor interpolation over interval centers isn't correct, since the beginning of one interval may be closer to the center of a different interval than to its own.
- Using previous interpolation, with `times[0, :]` as the anchor points does the job correctly.
- This now means that the `_const_interpolator` is completely unnecessary (and has been removed)
- I lifted the interpolation bits out of the main loop, and the whole thing looks a bit simpler now.

bmcfee:
It appears that removing `_const_interpolator` to rely on 1-sample interpolation in `previous` mode is only supported from scipy 1.10 and up.

I'm personally okay with this, but scipy 1.10 would bump our min python up to 3.8.  I think that would probably lift this PR up from a 0.8.1 to a 0.9 feature.  In the interest of making maintenance a little easier, I'll revert the const_interpolator hack for now, but leave a comment that it will not be necessary going forward.

craffel:
Sounds good to me! Thanks for working on this.

Issue #409, opened 2025-02-13
[Feature Request] Handle Monotonicity in Hierarchical Segmentation Metrics


tomxi:
#### **Description**
The hierarchical segmentation metrics in `mir_eval` are defined with the assumption that input annotations has monotonic labels and boundaries. However, in practice, hierarchies often don't have montonic labels, or even boundaries (outputs of spectral clustering for example).
While the original papers only define the case for fully monotonic hierarchies, this issue proposes a additional option for another sensible way of handling cases where hierarchies are not monotonic.

To extend flexibility regarding monotonicity, I propose:
1. **Adding a monotonicity check**: Before computing metrics, we should validate whether the hierarchical annotation has monotonic labels/boundaries.
2. **Handling different monotonicity cases**: The `_meet` function in `hierarchy.py` currently records the deepest level where two frames has the same label (meet), however, they don't require all the previous levels to meet (garunteed by monotonic labels, but not otherwise). Giving `_meet` function a option to record the deepest **monotonic** level that two frames meet in labels will enforce monotonicity in evaluation that way. This would not change the metric for the monotonic case but provide a different definition for annotation meet matrix that takes monotonicity into account.
3. **Exposing an option to control this behavior**: This option of how to build the annotation meet matrix should be elevated to the metric functions so users can specify how to handle monotonicity constraints.

#### **Proposed Changes**
- Implement checks for monotonicity violation.
- Introduce a new argument (e.g., `monotonicity_mode`) in `_meet` to determine how to construct the annotation meet matrix.
- Update hierarchical segmentation metric functions to expose `monotonicity_mode` as a user-configurable option.
- Update documentation and add tests for different monotonicity scenarios.

#### **Next Steps**
I will follow up with a pull request implementing these changes. Feedback on the approach and naming conventions would be appreciated! @bmcfee

PR #414, opened 2025-02-18
implements #409: added strict monotonicity flag for hierarchical segmentation metrics

tomxi:
added strict monotonicity flag for hierarchical segmentation metrics. Added new tests for coverage.

see #409 for details.

Soliciting feedback.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,8 +38,10 @@ Thumbs.db
 # Vim
 *.swp
 
-# pycharm
+# IDEs
 .idea/*
+.vscode/*
+.vscode/*
 
 # docs
 docs/_build/*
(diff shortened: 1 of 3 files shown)


tomxi:
Besides the _meet function and l-measure mentioned in #409, this PR also gave `t-measure` and the `_lca` function the same treatment.

I also fixed up the doc string for _meet as it was copied over from _lca and needed a minor updates.

tomxi:
Hmm.. CI checks are failing after the unit tests, at the upload coverage to Codecov step...

Can I get some help/guidance on how to resolve this?

bmcfee:
> Can I get some help/guidance on how to resolve this?

TLDR you can't.  We went round on this with the codecov devs a couple weeks ago, and couldn't really resolve it but my suspicion is that something screwy happened with permissioning when we migrated to an organization repo.  It works for PRs originating from the main repo, but not from forks.

We could make it so that failing coverage will not break the status, but that's not a great solution IMO.

tomxi:
Well I think this is developing into its own idea with other connected parts... dropping this for now.

PR #416, opened 2025-02-21
IO updates

### LINKED ISSUE #413: Support Path objects in data loaders
In modernizing some tests, I'm finding myself wanting to use Path objects rather than strings to specify file locations.  This revealed that our file loaders (eg io.load_events and the like) only support strings as input.  It would be easy enough to generalize this to support Paths in the `_open` context manager, and I see no reason not to do it.

### LINKED ISSUE #415: Deprecate io.load_wav
I should have noticed this when we deprecated separation #382 , but as far as I can tell, the `io.load_wav` function no longer serves any purpose.  I think we should deprecate it in 0.8.1 (and should have deprecated it in 0.8.0), and remove it in 0.9.

Any objections?

bmcfee:
Fixes #415 
Fixes #413 

---

This PR adds support for `pathlib.Path` objects in all `mir_eval.io` loaders.  (Easy fix, just adding a case to the `_open` context manager's type check.)

I've added one unit test that hits `load_delimited` with either string or Path input.  I think this is sufficient since: A) almost all other loaders run through `load_delimited`, and B) those that don't use the same `_open` context manager anyway.

Since I was in .io anyway, I added a deprecation on `load_wav`.

Assuming this passes CI, I think it should be good to merge.

### DIFF
--- a/mir_eval/io.py
+++ b/mir_eval/io.py
@@ -5,6 +5,7 @@ import numpy as np
 import re
 import warnings
 import scipy.io.wavfile
+from pathlib import Path
 
 from . import util
 from . import key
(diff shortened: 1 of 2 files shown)


bmcfee:
Rewrote this a bit following https://github.com/mir-evaluation/mir_eval/issues/413#issuecomment-2673176952 ; I'm not wild about raising an IOError from a TypeError, for many reasons.  (It's obtuse, provides no additional information, and adds complexity.  It also doesn't do anything useful if open throws a different kind of error.)

I've done it here so that we preserve the type of exception raised in case of failure, as changing it would constitute API breakage and I'd rather not do that in a minor release (as this is slated for).  But going forward (0.9), we should rip out this indirection and just let open() throw whatever exception it likes back up the stack.

Issue #424, opened 2025-06-24
Discussion: custom colormaps for chord plots


bmcfee:
In prototyping some viz code recently, I got to thinking that our generic handling of categorical labels in the display module could be better adapted for chord plots using the labeled segment display.  This example, pulled from a librosa issue discussion from a while back https://github.com/librosa/librosa/issues/1370#issuecomment-1189701677, shows how the current behavior looks:

![Image](https://github.com/user-attachments/assets/2d4ba53f-9ac5-4ee2-a269-8b9d7bcedd05)

This is basically fine in that distinct categories get distinct colors, but there are a few significant drawbacks:

1. There is no guaranteed consistency between plots that use different sets of labels.  F#:maj is blue in the above, but it could be orange in a different track.  This leads to some probably unnecessary context switching when viewing multiple annotations.
2. There is no musical logic to the color organization - it's essentially down to the order in which the labels appear in the track.  
3. There is no notion of "color proximity".  Hue and pitch class are both cyclic spaces (more or less), and this could be exploited to convey more information visually.

I started digging around the matplotlib options, and there really aren't any existing colormaps that would make much sense here.  Seaborn actually provides some nice functionality here for hsl/husl colormap construction, but I don't think we want to add seaborn to the dependency stack here.  Instead, I propose that we pre-generate a handful of custom colormaps for use in pitch-related plots.

## Protoype colormaps

Since these are categorical colormaps, we really don't need to worry about things like equal luminosity or perceptual uniformity.  What we do need to worry about is discriminability (under accessibility constraints, etc).  For this reason, I'm starting with seaborn's hsl generator, using n=12 to get evenly spaced hues according to pitch class in chromatic order.  I'm, generating light, medium, and dark versions of each, which I'm for now imagining as being used for other/major/minor qualities.  (I'm not married to that particular idea, but it seems like a reasonable enough starting point.)

We can plot these colormaps in both chromatic and circle-of-fifths order, resulting in the following (chatbot-generated/vibecoded, but LGTM):

![Image](https://github.com/user-attachments/assets/6cbee5fb-524c-4ac8-a151-ac40312cafdb)

We can also do the reverse, generating in CoF order instead of chromatic order:

![Image](https://github.com/user-attachments/assets/fc3b41c7-920f-4eef-b3b1-e92bc7b35dff)

[...]

PR #425, opened 2025-08-04
Chord display

bmcfee:
Implements #424 

This PR adds `mir_eval.display.chords()` and related functionality for visualizing chord annotations.

It is largely based on the `segments` function, and there is plenty of redundancy in the code here.  However, there are also enough differences that it's easier to duplicate than refactor.

The text annotation functionality is also fully in place, but I did not add test coverage for this yet as it currently depends on some slightly broken behavior in matplotlib that is likely to change at some point.  (See discussion in #424 )  As such, I don't mind a temporary coverage reduction.

Finally, this PR also registers six colormaps in the matplotlib registry when the module is imported:

- `pitch`
- `pitch_light`
- `pitch_dark`
- `fifths`
- `fifths_light`
- `fifths_dark`

The default is `fifths` for chord display, and otherwise the pitch modes are not used.  However, I think we can probably find some use for them later on, perhaps in some kind of rainbow piano roll viz or something. (I've experimented with this a little offline, but nothing too compelling has emerged yet.)

### DIFF
--- a/.github/environment-minimal.yml
+++ b/.github/environment-minimal.yml
@@ -10,4 +10,4 @@ dependencies:
   - pytest
   - pytest-cov
   - pytest-mpl
-  - pillow==9.0.0  # needed for test compatibilty and numpy 1.20.3
+  - pillow==9.0.0  # needed for test compatibility and numpy 1.20.3


bmcfee:
Bumping minimal matplotlib to 3.6 so as to handle colormap registration.

bmcfee:
Had to nudge the minimal matplotlib here, which has set off a chain of environment updates and other painful nonsense.

Meanwhile, I've also thrown in a workflow step to do an upload artifact for failing image comparison tests.  We recently added this over in librosa and it's a huge time saver.

bmcfee:
It appears that my use of `hatch_linewidth` requires matplotlib>=3.10.  This is quite a bummer, as I think it looks much better with thick hatches.  Here's the default (thin) version:

test_display_chord_nolabels

craffel:
Honestly the thick hatch version is not so bad. Not sure whether it's worth doing something conditional on matplotlib version. 

I think there are other hatches available that could be used for sus4, like stars etc., but I guess they might look terrible.

craffel:
Clownpants ftw

craffel:
That's a bit of a mixed metaphor when we're talking about software development!

Issue #432, opened 2026-02-19
Accelerated (exact) interval-based metrics


bmcfee:
@tomxi 's [ISMIR paper](https://ismir2025program.ismir.net/poster_167.html) from last year derives exact versions of several segmentation metrics.  The idea, which was noted early on (by @dpwe as I recall) but which we never quite acted upon, is that all of these sampling-based metrics will generally, in the limit as the frame rate → ∞, converge to functions of interval durations.  Done correctly, the time complexity of the evaluators should scale as a function of the number of boundaries (which dictate the number of piecewise-constant label regions), not the number of frames (proportional to track duration).

Tom's implementation here: https://github.com/tomxi/frameless-eval shows this for three representative metrics: pairwise frame classification, v-measure (and NCE), and L-measure.  The take-away message here is that compared to the typical frame rate of 10 Hz, the exact implementations are orders of magnitude more efficient, and more "accurate" (in the sense that sampling is always an approximation to the exact method).

This, generally, seems like a big win, as these evaluations can get relatively costly (especially for hierarchies) when put in the validation loop of model development.

## What's the proposal

For backward compatibility, we obviously need to retain the sampling-based implementations and their default frame rates.

I do think we can adopt the exact implementations (selected by `frame_size=None`) with no appreciable impact on default behavior.  That said, I think it's worth considering a deprecation cycle on the default rates to transition everyone to the exact implementations by default (with sampled versions available as backup).

Thoughts?

tomxi:
Started to work on this. Any advice on setting up the unit tests for the `frame_size=None` option? Currently the unit tests are based on precomputed scores stored in json files on test segmentations for `frame_size=0.1`. The least-effort unit tests that ensures coverage is to check that the frameless versions come close enough to the framed versions? Any suggestions?

bmcfee:
The existing tests are regression tests against prior outputs.  Obviously that's not an option here since there isn't a prior implementation that does the exact interval thing.

Instead, it should be possible to construct some toy examples where the correct answers can be computed analytically, and we then verify that the implementation agrees (within numerical tolerance).  This should be pretty easy for something like PFC or V-measures, probably less so for l-measures.

PR #436, opened 2026-07-31
Exact Interval-Based Segmentation Metrics

tomxi:
Implements issue #432.

Adds `frame_size=None` as an option for the following segmentation metrics: `segment.pairwise`, `segment.nce`, `segment.vmeasure`, and `hierarchy.lmeasure`: this option will calculate these metrics without using framing/sampling, thus provide more accurate results and are generally much faster.

Also capped the matplotlib version in github workflow to be less than 3.11, so that the display tests are passing: matplotlib 3.11 seems to have made very subtle changes to plot axis text layouts, and thus the display tests are not passing against the reference plots if using mpl 3.11.

### DIFF
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -6,7 +6,7 @@ dependencies:
   - pip
   - numpy >=1.20.3
   - scipy >=1.4.0
-  - matplotlib-base>=3.6.0
+  - matplotlib-base>=3.6.0,<3.11
   - pytest
   - pytest-cov
   - pytest-mpl
--- a/mir_eval/hierarchy.py
+++ b/mir_eval/hierarchy.py
@@ -674,8 +674,8 @@ def lmeasure(
         Like ``reference_intervals_hier`` and ``reference_labels_hier``
         but for the estimated annotation
     frame_size : float > 0 or None
-        length (in seconds) of frames.  The frame size cannot be longer than
-        the window. If None, use exact segment duration instead of framing.
+        length (in seconds) of frames.
+        If None, use exact segment duration instead of framing.
     beta : float > 0
         beta parameter for the F-measure.
 
@@ -695,15 +695,16 @@ def lmeasure(
 
         If the input hierarchies have different time durations
 
-        If ``frame_size > window`` or ``frame_size <= 0``
+        If ``frame_size <= 0``
     """
     # raise FutureWarning for change of default frame_size to None
-    warnings.warn(
-        "Default `frame_size` will change from 0.1 to None in a future version. "
-        "Set `frame_size` explicitly for consistent results.",
-        FutureWarning,
-        stacklevel=2,
-    )
+    if frame_size == 0.1:
+        warnings.warn(
+            "Default `frame_size` will change from 0.1 to None in a future version. "
+            "Set `frame_size` explicitly for consistent results.",
+            FutureWarning,
+            stacklevel=2,
+        )
 
     # Compute the number of frames in the window
     if frame_size is not None and frame_size <= 0:
--- a/mir_eval/hierarchy.py
+++ b/mir_eval/hierarchy.py
@@ -697,15 +697,6 @@ def lmeasure(
 
         If ``frame_size <= 0``
     """
-    # raise FutureWarning for change of default frame_size to None
-    if frame_size == 0.1:
-        warnings.warn(
-            "Default `frame_size` will change from 0.1 to None in a future version. "
-            "Set `frame_size` explicitly for consistent results.",
-            FutureWarning,
-            stacklevel=2,
-        )
-
     # Compute the number of frames in the window
     if frame_size is not None and frame_size <= 0:
         raise ValueError(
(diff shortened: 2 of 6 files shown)


mir_eval/segment.py
```diff
@@ -367,6 +368,14 @@ def pairwise(
         F-measure of detecting whether frames belong in the same cluster
 
     """
+    # raise FutureWarning for change of default frame_size to None
+    warnings.warn(
+        "Default `frame_size` will change from 0.1 to None in a future version. "
+        "Set `frame_size` explicitly for consistent results.",
+        FutureWarning,
+        stacklevel=2,
+    )
```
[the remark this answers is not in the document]
tomxi: Will change all similar FutureWarnings

tomxi:
Suggestions from Copilot's code review all seem reasonable, and I'll incorporate them in the next commit.

However, the CI checks are failing with various issues that I hope to get some guidance on how best to resolve:
- for the part of the test matrix with python=3.13, some unit tests in `test_display.py` are failing due to minor layout differences from reference `baseline.png`'s. This is because these environments don't follow .github/environment.yml and uses matplotlib=3.11. Updating all the baseline.png's probably belong to another PR, but what's a sensible patch right now to keep the display tests passing?

- for the other test environments (python<3.13), all the unit tests are passing, but CI check is failing at the upload coverage step with this message: 
`error - 2026-07-31 20:31:36,292 -- Upload queued for processing failed: {"message":"Token required - not valid tokenless upload"}`
any advice on how to resolve?

- CI Docs is failing at the link checking step: code.soundsoftware.ac.uk doesn't work anymore and music-ir.org has SSL issues. This also seems out of scope of this PR; any suggestions on how to proceed?

mir_eval/hierarchy.py
```diff
@@ -569,9 +673,9 @@ def lmeasure(
     estimated_labels_hier : list of ndarray
         Like ``reference_intervals_hier`` and ``reference_labels_hier``
         but for the estimated annotation
-    frame_size : float > 0
+    frame_size : float > 0 or None
         length (in seconds) of frames.  The frame size cannot be longer than
-        the window.
+        the window. If None, use exact segment duration instead of framing.
```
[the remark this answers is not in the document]
tomxi: Will remove mention of window in lmeasure doc string

mir_eval/segment.py
```diff
@@ -530,11 +569,15 @@ def _contingency_matrix(reference_indices, estimated_indices):
...
+    if seg_durations is None:
+        seg_durations = np.ones(len(reference_indices))
     # Using coo_matrix is faster than histogram2d
     return scipy.sparse.coo_matrix(
-        (np.ones(ref_class_idx.shape[0]), (ref_class_idx, est_class_idx)),
+        (seg_durations, (ref_class_idx, est_class_idx)),
         shape=(n_ref_classes, n_est_classes),
-        dtype=np.int64,
+        dtype=np.float64,
     ).toarray()
```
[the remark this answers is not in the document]
tomxi: will preserve int dtype when seg_durations is not provided

mir_eval/hierarchy.py
```diff
@@ -591,25 +695,53 @@ def lmeasure(
 
         If the input hierarchies have different time durations
 
-        If ``frame_size > window`` or ``frame_size <= 0``
+        If ``frame_size <= 0``
     """
+    # raise FutureWarning for change of default frame_size to None
+    if frame_size == 0.1:
+        warnings.warn(
+            "Default `frame_size` will change from 0.1 to None in a future version. "
```
craffel: Historically we only deprecate/make changes like this if the broad community consensus supports it, e.g. MIREX changing their approach or virtually all papers migrating to it. I'm not sure we're at that point for these metrics yet?

mir_eval/segment.py
```diff
@@ -367,6 +368,15 @@ def pairwise(
         F-measure of detecting whether frames belong in the same cluster
 
     """
+    # raise FutureWarning for change of default frame_size to None
+    if frame_size == 0.1:
+        warnings.warn(
+            "Default `frame_size` will change from 0.1 to None in a future version. "
```
craffel: LIkewise

mir_eval/segment.py
```diff
@@ -1036,6 +1086,15 @@ def nce(
     S_F
         F-measure for (S_over, S_under)
     """
+    # raise FutureWarning for change of default frame_size to None
+    if frame_size == 0.1:
+        warnings.warn(
+            "Default `frame_size` will change from 0.1 to None in a future version. "
```
craffel: Likewise

.github/environment.yml
```diff
@@ -6,7 +6,7 @@ dependencies:
   - pip
   - numpy >=1.20.3
   - scipy >=1.4.0
-  - matplotlib-base>=3.6.0
+  - matplotlib-base>=3.6.0,<3.11
```
craffel, line 9: Should we change the matplotlib entry under test dependencies in options.extras_require in setup.cfg too?

craffel (commented):
Maybe @bmcfee can check for correctness?

